PR #46, opened 2019-12-26
platform: Add Heltec ESP32 board

kallisti5:
* This board is meaningful because it is cheap, comes with a
  built-in OLED display *AND* has a built-in LoRa radio enabling
  over 6 Miles (10km) of long-range communication at a slow rate.
* Firmware boots, produces serial output, and blinks the Wifi LED
  on my local hardware.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,7 @@ env:
   - PIO_ENV=openevse_featheresp32
   - PIO_ENV=openevse_esp32-gateway
   - PIO_ENV=openevse_esp32-gateway-e
+  - PIO_ENV=openevse_esp32-heltec-wifi-lora-v2
   # - SCRIPT=ci_arduino.sh
   #   BUILD_TARGET=esp8266:esp8266:huzzah:FlashSize=4M1M
 
--- a/platformio.ini
+++ b/platformio.ini
@@ -298,3 +298,30 @@ monitor_speed = 115200
 extra_scripts = ${common.extra_scripts}
 board_build.partitions = ${common.build_partitions_esp32}
 board_build.extra_flags = "-DARDUINO_ESP32_GATEWAY=\'E\'"
+
+[env:openevse_esp32-heltec-wifi-lora-v2]
+platform = ${common.platform_esp32}
+board = heltec_wifi_lora_32_V2
+framework = arduino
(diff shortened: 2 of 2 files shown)


kallisti5:
Here's the board:
https://heltec.org/project/wifi-lora-32

Here's the schematic:
https://github.com/Heltec-Aaron-Lee/WiFi_Kit_series/blob/master/SchematicDiagram/WiFi_LoRa_32(V2)/WIFI_LoRa_32_V2(868-915).PDF

This board is *really* interesting in OpenEVSE since the units could mesh together and report stats back over a single gateway node with an internet connection over a 6+ mile radius

jeremypoulter:
Changes look ok, add to the `.travis.yml` file then should be good.

kallisti5:
Done. Thanks!  Getting an OpenEVSE 48A kit, so will be able to play with this one more in the future.

jeremypoulter:
Thanks, can't do more than check it builds at this stage, although that does look like a nice little board. Looking forward to see additional PRs that make use of the additional features.

glynhudson:
That's an impressive little board, interested to hear how LoRa with OpenEVSE could work

kallisti5:
@glynhudson good question :-)   So "LoRaWAN" is a protocol over LoRa which is a standard around meshing and giving LoRa nodes (basic IoT) access to the internet.  The downside of LoRaWAN is it requires a "gateway" which essentially is an internet connected system with "8+" LoRa radios.

The gateways are kind of hard to find and expensive at the moment :-|

I'm leaning towards just basic point to point LoRa communication since it's simpler and cheaper. The OpenEVSE units then could "talk to each other"

Anyway, feedback definitely welcome.. the options are infinite, but the "good uses" for such a thing are limited :-)

glynhudson:
I'm aware of TTN LoRaWAN network, which covers significant areas of cities in the UK (nowhere close to the whole country though). As a test could it be possible to use the TTN network for OpenEVSE where coverage is available? A larger antenna may be required?

kallisti5:
A TTN gateway would need to be within range.. and OpenEVSE would need code to speak LoRaWAN.

I was hacking together a  Raspberry Pi gateway with 2 LoRa modems, but the code is pretty... new and support for the chipsets I choose is non-existent :-|

PR #47, opened 2019-12-27
lora: Introduce basic LoRa support over SPI (WIP)

kallisti5:


kallisti5:
This is a *BIG* work in progress. Merging is not recommended. I mostly opened this for feedback/commentary on these potential additions :-)

The use-case of this isn't fully realized yet. At the moment the Heltec ESP32 just yells the OpenEVSE unit's hostname into the ether *continuously*. There are some standard methods for a rough "packet based" communication method where each device has a simple address and only listens / responds when their address comes up. None of this is implemented (yet)

kallisti5:
oh, and everything is wrapped with "ENABLE_LORA" to prevent compiling and including the LoRa code on devices that don't want it.

lib/lora/API.md
```diff
@@ -0,0 +1,315 @@
+# Heltec LoRa basic library API
```
jeremypoulter: Can you add the library via PlatformIO's library manager? https://platformio.org/lib/show/6051/Heltec%20ESP32%20Dev-Boards

Is there a library that is not specific to the Heltec board? maybe https://platformio.org/lib/show/1167/LoRa

platformio.ini
```diff
@@ -190,8 +190,6 @@ src_build_flags =
 build_flags =
   ${common.build_flags}
   ${common.build_flags_esp32}
-  -DRX1=25
-  -DTX1=27
```
jeremypoulter: Think these are removed by mistake

src/emonesp.h
```diff
@@ -88,4 +88,12 @@
 #define HAL_SHORT_ID_LENGTH 4
 #endif
 
+#if defined(ENABLE_LORA)
+//#define LORA_BAND 433E6	// 433 Mhz (Asia)
+//#define LORA_BAND 868E6	// 868 Mhz (Europe,Russia,India,Africa)
+#define LORA_BAND 915E6		// 915 Mhz (US,Australia)
```
jeremypoulter: This is likely to need to be changed so best to allow override from `platformio.ini`

```suggestion
#ifndef LORA_BAND
￼#define LORA_BAND 915E6		// 915 Mhz (US,Australia)
#endif
```

src/emonesp.h
```diff
@@ -88,4 +88,12 @@
...
+
+#define LORA_PABOOST true
```
jeremypoulter: ditto

```suggestion
#ifndef LORA_PABOOST
#define LORA_PABOOST true
#endif
```

platformio.ini
kallisti5: Oops. I removed these because none of the code seems to use them :-)

src/emonesp.h
kallisti5: Can-do. Good point.

lib/lora/API.md
kallisti5: Can do. The Heltec library should be generic for most of the SX chipsets, but lacks LoRaWAN support.  It was a "nice low bar" to get started.

kallisti5:
Some comments above :-)

 To be honest this is all pretty early.  I'm piecing together a cheap LoRaWAN gateway with some Raspberry Pi's and some custom PCB's.    One neat potential feature for this stuff is letting OpenEVSE units register on a network like https://www.thethingsnetwork.org.

This would let OpenEVSE units report their usage status every few minutes without a nearby wifi connection as long as a LoRaWAN gateway is "within a few miles" of range.  Just like the commercial players, there would be a way to use make a public EVSE network with open source code stuff.

platformio.ini
jeremypoulter: They are used in the ESP32 BSP to set the Serial pins, so you won't find the references in the code, just overrides the defaults.

chris1howell:
I really like this concept. I wonder if we could make a LoRaWAN gateway out of a ESP32. Ethernet would probably be best but WiFi should work also. Then relay messages from LoRa to MQTT at the gateway.

kallisti5:
> I really like this concept. I wonder if we could make a LoRaWAN gateway out of a ESP32. Ethernet would probably be best but WiFi should work also. Then relay messages from LoRa to MQTT at the gateway.

MQTT Pub/Sub from OpenEVSE over LoRa would be neat :-)

So i've been having trouble getting a LoRa gateway going. There aren't many LoRa *gateway* products on the market (in US frequencies) for reasonable prices.  The code to turn a Rasbperry Pi into a LoRa gateway is a bit sketchy and has limited radio support. (LoRa gateway from what I understand is a device with 8+ simultaneous communication channels (8 radios, or one radio + concentrator?)

I'm pondering just ignoring the "LoRa Gateway" aspect, and just keeping it simple with point to point LoRa support.  It would limit the "generic scalable mesh" capabilities, but would make the bar to entry *much* lower (and keep the code a lot simpler smaller for the short-haul)

jeremypoulter:
I think that would be a good way forward. From what I understand the gateways are a bit of a minefield both in terms of complexity and patents, which is what you pay the money for. So a point-to-point solution is probably a first good step to prove the use case, then expand on that to use The Things Network or similar. 

FYI I know a few folks who have an interest in LoRa and TTN through Reading Hackspace and I also know folks at LoRa Alliance if you have any LoRa issues.

kallisti5:
I'm looking back at LoRaWAN again.   OpenEVSE would be 100% proprietary on LoRa without LoRaWAN.  Given the bandwidth limitation of LoRa connections, it would be a huge fight for resources on the limited bandwidth LoRa frequencies without using LoRaWAN to coordinate. OpenEVSE would also have to roll it's own LoRa protocol stack, packet formats, collision mitigation, etc.

I just purchased one of these which have started showing up for cheap on Ebay:

https://www.ebay.com/itm/LoRaWAN-Gateway-Collector-LoRa-Tektelic-IoT-Smart-City-Kona-Pico-Micro/293543448769?ssPageName=STRK%3AMEBIDX%3AIT&_trksid=p2060353.m2749.l2649

(no affiliated with this person in any way)

kallisti5:
Woot. I now have a LoRaWAN gateway working and an ESP32 sending data to it using generic libraries.

I'm pretty sure at this point I can get the LoRaWAN support down to one generic library and one source file.

I'm going to be reworking this to be slightly less "large"

PR #56, opened 2020-02-16
SNTP support

### LINKED ISSUE #33: Use NTP to set time on OpenEVSE controller
As a first step towards #5 and negating the need for a hardware RTC, NTP time from the ESP could be used to automatically set the time on the OpenEVSE controller.

We should keep the option to "sync" the time from a web browser for the use case where no web connection is present.

jeremypoulter:
Fixes #33 

To Do

- [x] Update time in UI when changed via NTP
- [x] Configure the timezone
- [x] Set ESP time from EVSE on start up
- [x] Wrap manual/browser time setting in HTTP API so ESP32 time is also set


jeremypoulter:
Hum... Yes, may need the timezones for NTP, will check on than. 

I think on a new install NTP is the default, but if you have enabled any of the services the "default" is disabled, this is a bi-product of using the `flags` area in flash.

Only the time on the EVSE will change, that is not currently reflected in the WebUI, can possibly make that change, send the updated time over the web socket. Or is it more than that?

glynhudson:
Just tested latest version, ntp and timezone selection works well. Is it possible to make Europe/London the default?

glynhudson:
I've noticed an issue that the GUI timezone display doesn't reflect the actual config timezone. After a refresh the GUI always reverts to showing `africa/abidjan`, however the times stays correct for `europe/london`, and the `/config` output is currently showing europe/london.

jeremypoulter:
This should be resolved now, please update

glynhudson:
Yes, it's fixed now. Nice work. I think this is ready to merge?

jeremypoulter:
I have not done anything around manually setting the time and the timezone so may need to do something there. it its current mode the time will be set as GMT then the configured time zone applied. Maybe as a quick fix we disable the time zone and only enable for NTP?

glynhudson:
That's a good idea, and make europe/london the default

On Fri, 13 Mar 2020 at 23:47, Jeremy Poulter <notifications@github.com> wrote:
>
> I have not done anything around manually setting the time and the timezone so may need to do something there. it its current mode the time will be set as GMT then the configured time zone applied. Maybe as a quick fix we disable the time zone and only enable for NTP?
>
> —
> You are receiving this because your review was requested.
> Reply to this email directly, view it on GitHub, or unsubscribe.



-- 
Glyn Hudson

https://openenergymonitor.org
http://zerocarbonadventures.co.uk

PR #70, opened 2020-04-21
lora: Initial work at LoRaWAN support. Empty announcements.

kallisti5:
This has been greatly simplified using a common LoRa library, and removing the Heltec code.

1) If enabled for the board, LoRa is setup using the network credentials hard coded within the code
   (these can be pulled via the OpenEVSE storage, but just not there yet)
2) If a LoRaWAN gateway is in range (a few miles), the device will begin to announce the OpenEVSE status.
  (the status is empty at the moment, need to pull from OpenEVSE still. The status information needs packed into as small a datastructure as possible)
3) The LoRaWAN network (in my case The Things Network) will receive the data, translate it, and pass it onto any configured integrations the user specifies. (aka, Plugshare??? :-) )

### DIFF
--- a/src/lora.cpp
+++ b/src/lora.cpp
@@ -1,0 +1,153 @@
+/*
+ * Copyright (c) 2019-2020 Alexander von Gluck IV for OpenEVSE
+ *
+ * -------------------------------------------------------------------
+ *
+ * Additional Adaptation of OpenEVSE ESP Wifi
+ * by Trystan Lea, Glyn Hudson, OpenEnergyMonitor
+ * All adaptation GNU General Public License as below.
+ *
+ * -------------------------------------------------------------------
+ *
+ * This file is part of Open EVSE.
+ * Open EVSE is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3, or (at your option)
+ * any later version.
+ * Open EVSE is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with Open EVSE; see the file COPYING.  If not, write to the
+ * Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ */
+#ifdef ENABLE_LORA
+
+#include <lmic.h>
+#include <hal/hal.h>
+#include <SPI.h>
+
+#include "emonesp.h"
+#include "lora.h"
+
+#include "app_config.h"
+
+
+#define LORA_HTOI(c) ((c<='9')?(c-'0'):((c<='F')?(c-'A'+10):((c<='f')?(c-'a'+10):(0))))
+#define LORA_TWO_HTOI(h, l) ((LORA_HTOI(h) << 4) + LORA_HTOI(l))
+#define LORA_HEX_TO_BYTE(a, h, n) { for (int i = 0; i < n; i++) (a)[i] = LORA_TWO_HTOI(h[2*i], h[2*i + 1]); }
+#define LORA_DEVADDR(a) (uint32_t) ((uint32_t) (a)[3] | (uint32_t) (a)[2] << 8 | (uint32_t) (a)[1] << 16 | (uint32_t) (a)[0] << 24)
+
+#define ANNOUNCE_INTERVAL 30 * 1000 // (In Milliseconds)
+
+
+// TODO: Store these via WebUI? We're doing (simple) ABP activation for now
+const char *devAddr = "00000000";
+const char *nwkSKey = "00000000000000000000000000000000";
+const char *appSKey = "00000000000000000000000000000000";
+
+
+// LoRaWAN credentials to use
+static uint8_t DEVADDR[4];
+static uint8_t NWKSKEY[16];
+static uint8_t APPSKEY[16];
+
+// Next LoRaWAN announcement
+unsigned long nextAnnounce;
+
+// LoRa module pin mapping
+const lmic_pinmap lmic_pins = {
+  .nss = LORA_NSS,
+  .rxtx = LMIC_UNUSED_PIN,
+  .rst = LORA_RST,
+  .dio = {LORA_DIO0, LORA_DIO1, LORA_DIO2},
+};
+
+// Used for OTAA, not used (yet)
+void os_getArtEui (u1_t* buf) { }
+void os_getDevEui (u1_t* buf) { }
+void os_getDevKey (u1_t* buf) { }
+
+
+void onEvent(ev_t ev) {
+  switch (ev) {
+    case EV_TXCOMPLETE:
+      DBUGF("LoRa: TX Complete.");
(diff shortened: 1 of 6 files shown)


src/lora.cpp
```diff
@@ -0,0 +1,152 @@
...
+// LoRaWAN credentials to use
+static uint8_t DEVADDR[4];
+static uint8_t NWKSKEY[16];
+static uint8_t APPSKEY[16];
+
+// LoRaWAN data packet
+unsigned long lastAnnounce;
+uint8_t dataPacket[9];
+
+// LoRa module pin mapping
```
kallisti5, line 61: We *might* be able to get these from HAL.. but not sure HAL is aware of them if we don't pull in the Heltec libraries.  Needs more investigation.  For now we can just define these per board.

src/lora.cpp
```diff
@@ -0,0 +1,152 @@
...
+  .dio = {LORA_DIO0, LORA_DIO1, LORA_DIO2},
+};
+
+// Used for OTAA, not used (yet)
+void os_getArtEui (u1_t* buf) { }
+void os_getDevEui (u1_t* buf) { }
+void os_getDevKey (u1_t* buf) { }
+
+
+void onEvent(ev_t ev) {
```
kallisti5, line 75: For now we just blast our status to the network every 30 seconds. We could add some confirmation code to "ensure the network accepted the update", but for now i'm keeping it simple.

src/lora.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+#define LORA_HTOI(c) ((c<='9')?(c-'0'):((c<='F')?(c-'A'+10):((c<='f')?(c-'a'+10):(0))))
+#define LORA_TWO_HTOI(h, l) ((LORA_HTOI(h) << 4) + LORA_HTOI(l))
+#define LORA_HEX_TO_BYTE(a, h, n) { for (int i = 0; i < n; i++) (a)[i] = LORA_TWO_HTOI(h[2*i], h[2*i + 1]); }
+#define LORA_DEVADDR(a) (uint32_t) ((uint32_t) (a)[3] | (uint32_t) (a)[2] << 8 | (uint32_t) (a)[1] << 16 | (uint32_t) (a)[0] << 24)
+
+#define ANNOUNCE_INTERVAL 30 * 1000 // (In Milliseconds)
+
+
+// TODO: Store these via WebUI? We're doing (simple) ABP activation for now
```
kallisti5, line 46: ABP activation is where "every device is configured"  OpenEVSE could move to a "OTAA" activation where devices "automatically register and activate themselves under an application" but this is overkill unless OpenEVSE wants to become a 'managed' product (plug it in, and it works under the OpenEVSE account)

kallisti5:
Testing has been limited thus far, without a spare OpenEVSE module testing is tough since the ESP32 enters a "Unable to communicate to OpenEVSE" loop.  The code was taken from another (working) project I have.

kallisti5:
Oh.. beyond The Things Network, this code could be used to connect to any other LoRaWAN network (915 Mhz in the US). one option is Helium.  https://developer.helium.com/console/cli

Not sure of the options there, however there are a *ton* of Helium devices out there today. https://network.helium.com/coverage

src/lora.cpp
jeremypoulter: Does this have to be anything specific? There is a hardware ID that can be obtained from the (ESP) HAL (the WiFi MAC)

src/lora.cpp
```diff
@@ -0,0 +1,152 @@
...
+  DBUGF("LoRa: Starting LoRaWAN broadcast...");
+  // Check if there is not a current TX/RX job running
+  if (LMIC.opmode & OP_TXRXPEND) {
+    DBUGF("LoRa: Modem busy. Retry later");
+    return;
+  } 
+
+  // TODO: Pack important OpenEVSE stats into dataPacket.
+
+  LMIC_setTxData2(1, dataPacket, sizeof(dataPacket), true);
```
jeremypoulter, line 139: When is the dataPacket filled in with the state?

src/lora.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+  lastAnnounce = millis();
```
jeremypoulter, line 141: IHMO it is better to work out what time the next packet should be, ie:
```suggestion
  nextAnnounce = millis() + ANNOUNCE_INTERVAL;
```

You only need to do the addition once rather than every check, but a minor point

src/lora.cpp
kallisti5: sorry, that comment was aimed at the LoRa module pin mapping.   The Heltec ESP32 V1 vs V2 has different pin mappings per device.  I've seen some references to the pin mappings being available from HAL, but there are no docs around using them.

https://github.com/mcci-catena/arduino-lmic/blob/master/src/hal/getpinmap_heltec_lora32.cpp

src/lora.cpp
kallisti5: > Does this have to be anything specific? There is a hardware ID that can be obtained from the (ESP) HAL (the WiFi MAC)

The three values identify + authenticate the OpenEVSE unit on the LoRaWAN network.  Pretty much the three values are generated at your LoRaWAN gateway provider and filled in. The values will be unique per OpenEVSE unit. (prime candidates for a webui setting)

Other things to adjust in the web ui in the future are the LoRa Spread factor. (7 = narrow bandwidth, but less prone to interference.  12 = wide bandwidth, but more prone to interference)

https://www.thethingsnetwork.org/article/how-spreading-factor-affects-lorawan-device-battery-life explains SF a bit.


>  When is the dataPacket filled in with the state?

That's what the todo is for :-)   We need to decide what's "important enough" to broadcast over LoRaWAN to a gateway potentially a few miles away.  Charger state, vehicle state of charge, etc.  Likely some subset of what you can send to MQTT today.  we should pack these as small as possible to improve reliability.  (two bits for charger state, etc)


Pretty much, data payloads that are sent over LoRa need to be as small as possible.  But essentially we're free to pack whatever OpenEVSE stats we want to.


Then, you add some "decoding logic" at the destination to unpack the lora data into whatever structure you want before passing it onto other integrations.

There is a review of the process here:
https://hackmd.io/@pmanzoni/Hy6qUmbA4


I'm using The Things Network here since it's free and has a low bar to entry. There are several however.  Essentially a LoRaWAN gateway such as The Things Network is an online API to:

1) Receive data from physical LoRaWAN gateway devices attached to the internet.
2) Translate the data from LoRaWAN, and send to configured integrations.

Think of LoRaWAN gateways as "cell phone towers".   Anyone can use the LoRaWAN gateway you provide, but each LoRaWAN gateway is configured for "a single gateway service" like The Things Network (free), helium.com (cryptocurrency based cost to use LoRaWAN), loriot.io (haven't looked at them yet)

src/lora.cpp
kallisti5: Ah. Good point :-)

kallisti5:
Ok. The latest patch has a bunch of fixes and actually pushes useful data over LoRaWAN :-)

The only change which is active when building non-lora enabled firmware is the addition of the "create_rapi_packed" function to input.cpp.  This takes important values and packs them into a bunch of uint8's to be broadcast over LoRa.  If a value could be reasonably > 255, I added a conversion factor (voltage, elapsed, temp)

Keep in mind this is all still untested. I need to short circut the OpenEVSE communication tests to try and fake the ESP32 to send some data over LoRaWAN.

jeremypoulter:
Replaced by #537

PR #84, opened 2020-06-10
Immediately post to the EmonCMS server when config changes.

### LINKED ISSUE #64: Improve responsivness of feedback when user enables / modifies emoncms HTTP posting
Currently, when a user enabled or modifies emoncms http posting and clicks 'save' there is not immediate feedback that the details entered are correct until the next post interval. 

Ideally clicking 'save' should trigger and immediate HTTP post which would verify successful connection. 

Ideally the debug output should display the server response, I'm still having trouble making connection to some emoncms.org accounts, this would be useful to debug.

jeremypoulter:
This gives the user immediate feedback as to if the settings are correct or not.

Also event the returned message to help the user figure out what is wrong if they do not enter the details correctly.

Fixes #64


glynhudson:
Fantastic, confirmed fixed #64

PR #111, opened 2020-08-25
Use the disable state to pause charge

### LINKED ISSUE #83: Think City charging problem; Option to disable pilot while sleeping?
Hi!

Proud new owner of [Think City](https://en.wikipedia.org/wiki/Think_City) electric car, 2010 model two seater with 23kWh Lithium battery. :) 

Originally car had a custom charging connector, but I was told that the signaling is the same as J1772. I changed the car side connector to J1772 so I can use same cables as with my Outlander PHEV.

Now the car charges fine with OpenEVSE, and does respect the pilot signal (6A-16A). Eco mode does work etc.

However, if OpenEVSE is left at "Sleeping" state for a long time, it's not possible to restart charging directly. Car simply refuses to use any power, it acts like it already has full battery. But if I send $FD rapi command to disable OpenEVSE, then wait 10 seconds and click "Start", it starts charging correctly.

This is a problem with Eco mode, because if there is no solar available it will go to sleep and does not wake again. 

The problem must be that the pilot signal is always active even if OpenEVSE is sleeping. Going to "Disabled" removes pilot and allows car to restart charging.

So.. Is it possible to use disable instead of sleep in Eco? Then I think it would work. Or is pilot signal needed to detect if car is plugged in or not?

jeremypoulter:
Fixes #83


glynhudson:
Nice work. I've just tested this, it worked as expected. Although, switching between paused / disabled had no effect on nissan leaf. I'll try and find some Zoe / ZS EV owners to test

PR #117, opened 2020-09-21
divert_PV_ratio: adjustable requirement on available PV power before starting charge or incrementing current

treepleks:
Adds a divert_PV_ratio parameter with a default value of 1.0 that 
    requires 100% of PV power for starting charging. There is a related wifi gui
    pull request. This compiles fines but has not been tested. Some feedback
    from the Open EVSE team would be welcome before I test flashing my
    ESP32.

    Motivation

    Because charge cannot start with less than 6A, and is incremented by
    steps of 1A, it is not possible for OpenEVSE to exactly consume the
    available excess of PV power. The divert_PV_ratio declares what
    minimum fraction of PV power is required to start charging or
    increment charging current.
    
    The default of 1.0 means that charging only starts (or is incremented)
    when all the required power is available as excess PV power. A higher
    value (1.1) would start only with an extra safety margin.
    
    Values below 1.0 are economically preferable if the kWh selling price
    is strictly less than the kWh night-price which is the usual
    situation. In this case, if the current fraction of excedent PV power
    is sufficient to make the extra imported power cost less than it would
    cost with the night price, it's better to charge now.
    
    As a rule of thumb, if the Daylight price is noted "D", the Night
    price "N" and the selling price "S", the optimal setting for the
    divert_PV_ratio is:
    
                      (D-N)/(D-S)
    
    If the actual (Daylight) price is 0.15/kWh but there is cheap Night
    charging price of 0.10/kWh and an actual selling price of 0.05/kWh
    then the best ratio to use is:
    
        (0.15 - 0.10)/(0.15 - 0.05) = 0.05/0.1 = 0.5

### DIFF
--- a/divert_sim/RapiSender.cpp
+++ b/divert_sim/RapiSender.cpp
@@ -1,6 +1,5 @@
 #include <Arduino.h>
 #include "RapiSender.h"
-#include "debug.h"
 #include "openevse.h"
 #include "input.h"
 
@@ -9,6 +8,7 @@
 #ifdef ENABLE_DEBUG
 #define DBG
 #endif
+#define DBUGF
 
 static CommandItem commandQueueItems[RAPI_MAX_COMMANDS];
 
--- a/divert_sim/divert_sim.cpp
+++ b/divert_sim/divert_sim.cpp
@@ -7,11 +7,10 @@
 #endif
 
 #include "Console.h"
-#include "emonesp.h"
+//#include "emonesp.h"
 #include "RapiSender.h"
 #include "openevse.h"
 #include "divert.h"
-#include "emonesp.h"
 #include "event.h"
 
 #include "parser.hpp"
@@ -25,6 +24,7 @@ long pilot = 32;                      // OpenEVSE Pilot Setting
 long state = OPENEVSE_STATE_SLEEPING; // OpenEVSE State
 String mqtt_solar = "";
 String mqtt_grid_ie = "";
+uint32_t flags;
 
 int date_col = 0;
 int grid_ie_col = -1;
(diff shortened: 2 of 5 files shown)


jeremypoulter:
It would be really handy if you could use the `divert_sim` to generate some graphs of how these changes alter the behaviour of the charging.

jeremypoulter:
FYI this is what I generated from the current divert module
[OpenEVSE Solar Divert Simulations - Final.pdf](https://github.com/OpenEVSE/ESP32_WiFi_V3.x/files/5279412/OpenEVSE.Solar.Divert.Simulations.-.Final.pdf)

treepleks:
Thanks a lot Jeremy for the feedback. I must confess I sort of hoped the Open EVSE team would be ready to do the simulation :-). I installed the full ESP32 compilation kit just to add this improvement, I can produce a firmware but am still reluctant to flash it on the wall box before it has been tested. 
Is there a guide on how to use/setup divert_sim? Do I need an ESP32 module for this (I have one or 2 left I think, but not the same branding company). Any pointer/URL would be handy.

treepleks:
Ok, looking to the code, I understand divert_sim is just a wrapper around divert. I added the required 

`+double divert_PV_ratio = 0.5;`

in divert_simp.cpp. Also added MicroDebug.h and StreamSpy.h. But make still complains at linking:
```
g++ -o divert_sim  divert.o WString.o RapiSender.o Console.o Print.o Core.o divert_sim.o -I. -I FakeDuino -I ../src -I../.pio/libdeps/openevse_huzzah32_idf/ArduinoJson/src -I../.pio/libdeps/openevse_huzzah32_idf/OpenEVSE/src -ggdb -DDIVERT_SIM -DRAPI_PORT=Console -pthread
/usr/bin/ld: divert.o: in function `config_pause_uses_disabled()':
/home/tom/Dev/ESP32_WiFi_V3.x/divert_sim/../src/app_config.h:105: undefined reference to `flags'
/usr/bin/ld: divert_sim.o: in function `__static_initialization_and_destruction_0(int, int)':
/home/tom/Dev/ESP32_WiFi_V3.x/divert_sim/divert_sim.cpp:22: undefined reference to `SerialEvse'
collect2: error: ld returned 1 exit status
```
I must be missing something. Any hint here?

jeremypoulter:
Oh, I will take a look at it, that is s related to some resent changes I made...

treepleks:
It's Ok. I could achieve it with a few changes:

RapiSender.cpp:
```
-#include "debug.h"
+#define DBUGF
```

divert_sim.cpp
```
-#include "emonesp.h"
-#include "emonesp.h" (yes, it's there twice :-)
+uint32_t flags;
+double voltage = 240; // Voltage from OpenEVSE or MQTT
+double divert_PV_ratio = 1.0;
```

And it worked. Simulations show it works as expected. Attached simulations for various values of PV_divert_ratio (GRID_IE_RESERVE_POWER  = 0.0  except last where it is 100 as in the vanilla version) 

- 0.05: wants to sell nothing to the grid provider
- 0.39: optimal for me
- 0.5: starts/increment as soon as 50% of the marginal difference is there
- 0.75: same with 75%
- 1.0: wants to avoid eating from the grid
- 1.1: same with a margin safety on transitions
- 1.0+100: same with a constant margin safety of 100W (vanilla OpenEVSE setting essentially)

I would suggest that when divert_PV_ratio >1, the decimal part should go to GRID_IE_RESERVE_POWER. This means that with divert_PV_ratio = 
- 1.1 it would wait for 100% solar and 144W margin. 
- 1.2 it would wait for 100% solar and 288W margin. 
- 0.9 it would wait for 90% solar and 0W margin. 

Just tell me what you think of this.
[Divert-1.0+100.pdf](https://github.com/OpenEVSE/ESP32_WiFi_V3.x/files/5286870/Divert-1.0%2B100.pdf)
[Divert-1.1.pdf](https://github.com/OpenEVSE/ESP32_WiFi_V3.x/files/5286871/Divert-1.1.pdf)
[Divert-1.0.pdf](https://github.com/OpenEVSE/ESP32_WiFi_V3.x/files/5286872/Divert-1.0.pdf)
[Divert-0.75.pdf](https://github.com/OpenEVSE/ESP32_WiFi_V3.x/files/5286873/Divert-0.75.pdf)
[Divert-0.5.pdf](https://github.com/OpenEVSE/ESP32_WiFi_V3.x/files/5286874/Divert-0.5.pdf)
[Divert-0.39.pdf](https://github.com/OpenEVSE/ESP32_WiFi_V3.x/files/5286875/Divert-0.39.pdf)
[Divert-0.05.pdf](https://github.com/OpenEVSE/ESP32_WiFi_V3.x/files/5286876/Divert-0.05.pdf)

jeremypoulter:
Thanks for doing those, I must admit I have read the description several times and I am not entirely sure I get the logic. The graphs certainly look better from the point of view of not starting/stopping the charge so much, which I guess is the point, to sacrifice using grid power to make sure all solar is used. I think I would prefer the default to be 1.1 do better match the existing logic, but with that change I would be happy for it to be merged. @glynhudson what do you think?

glynhudson:
Adding `divert_PV_ratio` as an advanced option looks like a good idea to me. The simulations looks good and make sense. 

Have you tested this on your ESP32?

treepleks:
Hello Glyn,

Yes. After I sent the simulations to Jeremy I actually flashed my Open EVSE with it and it works just fine. It changes a lot for me. I have 3kW panels, with work at home, we easily use 900W for various items (up to 4 computers are on) and on a sunny day it leaves around 1500W export and the open EVSE wallbox... would never start.

Now, setting the ratio to 0.5 (which is safe, above my supposedly optimal value of 0.39) my car has charged several hours today, at minimal marginal cost.

Just great :-)

glynhudson:
@treepleks could something have gone wrong with your PR? I've just checked and I can't see a PV ratio option running the latest version:
 
Screenshot 2021-01-03 at 01 44 25

treepleks:
Hello Glynn. I'm surprised because I flashed my OpenEVSE charger with your last release (3.3.2) when I saw my PR was inside and I have it. How can this be? 

Screenshot_20210103-150911

glynhudson:
Ah, got it. My mistake solar PV ratio is only displayed when `Grid +I-E` is selected

Issue #92, opened 2020-06-18
MCP9808 Temperature Sensor 


chris1howell:
Add support for MCP9808 temperature sensor.

Development OpenEVSE WiFi integrated 5 - 12v switch mode power supply, two neoPixels and MCP9808 temperature sensor.

Code successfully tested on development hardware.

````
#include <Wire.h>
#include <Adafruit_MCP9808.h>

Adafruit_MCP9808 tempsensor = Adafruit_MCP9808();

void setup() {
  Wire.begin();
}

void loop() {
  float c = tempsensor.readTempC();
}
````

![2020-06-18](https://user-images.githubusercontent.com/758844/85077813-148ce480-b1bb-11ea-9e2b-4c472ec62337.jpg)

jeremypoulter:
Should the temp just be exposed via MQTT/EmonCMS like the other temps, or are there additional requirements?

glynhudson:
@chris1howell do we need a RAPI command to pass the temperature reading into the open_evse controller so it can be used for in the normal way for over temperature throttling etc.

chris1howell:
It would probably be a good idea to pass the temp to the controller.

What I think we should do long term is have the controller only do the
safety shutdown at a single point 72C (or whatever we set) and have the
WiFi do the throttling.

The WiFi has a lot more resources to fine tune the current and will be
controlling for many other parameters.

PR #121, opened 2020-10-13
OpenEVSE WiFi v1 support

jeremypoulter:
Initial support for the OpenEVSE WiFi v1 board, #91 and #92

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -302,7 +302,7 @@ board_build.partitions = ${common.build_partitions_esp32}
 
 [env:openevse_wifi_v1]
 platform = ${common.platform_esp32}
-board = featheresp32
+board = esp32dev
 framework = arduino, espidf
 lib_deps =
   ${common.lib_deps}
@@ -318,10 +318,10 @@ src_build_flags =
   -DWIFI_BUTTON=0
   -DWIFI_BUTTON_PRESSED_STATE=LOW
   -DRAPI_PORT=Serial
-  -DDEBUG_PORT=Serial2
+  -DDEBUG_PORT=Serial
   -DSERIAL_RX_PULLUP_PIN=3
-  -DTX2=16
-  -DRX2=19
+  -DI2C_SDA=21
+  -DI2C_SCL=22
   -DENABLE_MCP9808
   -DENABLE_DEBUG_LED
   -DENABLE_DEBUG_INPUT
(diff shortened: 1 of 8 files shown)


jeremypoulter:
@chris1howell I have added the LED support and the temp sensor support although the latter does not look to be finding the sensor. Are you able to test?

chris1howell:
@jeremypoulter <https://github.com/jeremypoulter> Temperature sensor is
working by adding the following:

input .h
#ifdef ENABLE_MCP9808
Wire.begin(I2C_SDA, I2C_SCL);

platform.ini
-DI2C_SDA=21
-DI2C_SCL=22

LEDs do not seem to be working, I do see the LED Manager messages in Debug
but color stays white.

Additionally, I can not save MQTT in the UI. I keep getting "Failed to save
Config" messages.

jeremypoulter:
Added a fix for those issues, looks like using Serial2 breaks the NeoPixel, Could not get Serial1 to work so left the debug on Serial for now :( 

I2C now initialised on the correct pins, sorry I missed that in the sample code. Getting a value now, but looks to be 10 degrees higher than it should be...

chris1howell:
Great work.... Both LED and temp are working. I am also seeing the same
+10C on the temp sensor, will look into it.

LEDs look great, using one LED for WiFi status and the other for EVSE
status is really cool. Works perfectly when the station is connected to
WiFi as a client. I did note the station does not change color for EVSE
status when in AP mode (with no device connected) and in AP mode connected
as standalone. The LED stays green for "Ready", "Connected" and "Charging",
Red error states seem to work in all states.

jeremypoulter:
> LEDs look great, using one LED for WiFi status and the other for EVSE status is really cool

Not exactly what was intended, can you send a video... I did notice on my board there is a slight colour difference between the two LEDs... 

> Works perfectly when the station is connected to WiFi as a client. I did note the station does not change color for EVSE status when in AP mode (with no device connected) and in AP mode connected as standalone.

That was the intention, the LED should be used for WiFi or error states when connecting and the EVSE state when connected to the WiFi.

chris1howell:
The EVSE state must always be displayed visualy, it is a requirement of UL,
J1772 and IEC. WiFi State is not required by any regulation. When on the
road using a portable station (no LCD display) in "Standalone Mode" EVSE
status is still needed, there are many use cases where WiFi is not used or
avaliable.

My suggestion would be to use one LED for station status and the second for
WiFi visual feedback.

PR #123, opened 2020-10-20
Add new gui assets to fix failing TravisCI build

toofishes:
These assets were newly added with the last gui library update, but were
not checked into git. The latest master build in TravisCI is failing as
a result: https://travis-ci.org/github/OpenEVSE/ESP32_WiFi_V3.x/builds/737070599


glynhudson:
Thanks :+1:

PR #125, opened 2020-10-25
Configurable AP passphrase

ceesb:
I added a configurable passphrase for WiFi AP under the "advanced settings" menu.

### DIFF
--- a/src/app_config.cpp
+++ b/src/app_config.cpp
@@ -19,7 +19,12 @@
 
 // Wifi Network Strings
 String esid;
+
+// passphrase for Client mode
 String epass;
+
+// passphrase for Access Point mode
+String apass;
 
 // Web server authentication (leave blank for none)
 String www_username;
@@ -77,6 +82,7 @@ ConfigOpt *opts[] =
 // Wifi Network Strings
   new ConfigOptDefenition<String>(esid, "", "ssid", "ws"),
   new ConfigOptSecret(epass, "", "pass", "wp"),
+  new ConfigOptDefenition<String>(apass, "openevse", "apass", "apass"),
 
 // Web server authentication (leave blank for none)
   new ConfigOptDefenition<String>(www_username, "", "www_username", "au"),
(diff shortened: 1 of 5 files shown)


glynhudson:
I notice you have included a change to point the GUI sub-module to https://github.com/ceesb/openevse_wifi_gui, I don't think this was intentional. Can you amend?

ceesb:
Ah sorry, I fixed that. The sub-module stuff is confusing when you fork it ;)

There is another issue that I'm unable to fix, and that is the (client side) validation of input fields in the web interface. This is especially relevant for the AP passphrase, since you need more than 8 characters or the AP will not come up anymore. I looked into "knockout-validation" to get the input fields validated client side, but I lack the web development skills to enable this correctly. 

If you want, I can add another save-ap-passhphrase end-point in the webserver that does validation server side, but than I can't have the passphrase field under the "advanced settings" tab.

glynhudson:
What do you think of this PR @jeremypoulter? This would be a good feature to add. Would be best to wait until V4.0 is ready?

src/app_config.cpp
```diff
@@ -77,6 +82,7 @@ ConfigOpt *opts[] =
 // Wifi Network Strings
   new ConfigOptDefenition<String>(esid, "", "ssid", "ws"),
   new ConfigOptSecret(epass, "", "pass", "wp"),
+  new ConfigOptDefenition<String>(apass, "openevse", "apass", "apass"),
```
jeremypoulter, line 85: The second name should be short to save flash space but the first can be a bit more descriptive
```suggestion
  new ConfigOptDefenition<String>(apass, "openevse", "ap_pass", "ps"),
```

jeremypoulter:
Can add to v3 or v4 does not look to have anything that would conflict. just a minor change suggested

jeremypoulter:
Oh and the issue @ceesb mentions about validation, I will give it some consideration on how to add validation to the generic endpoint, probably more than just the AP password should be validated. In terms of doing it in the GUI, there is an example here: https://github.com/openevse/openevse_wifi_gui/blob/wifi_timer_updates/src/view_models/OpenEvseWiFiViewModel.js#L466-L477

jeremypoulter:
Superceeded by #595

PR #126, opened 2020-10-25
Fix CORS for zones.js and timeout using espota for upload

arjanmels:
Two minor fixes:
- Fix CORS for zones.js, by adding "Access-Control-Allow-Origin" header to all static requests
- timeout using espota for upload, by add feedLoopWDT in onProgress

### DIFF
--- a/src/ota.cpp
+++ b/src/ota.cpp
@@ -2,7 +2,7 @@
 #include "ota.h"
 
 #include <WiFiUdp.h>
-#include <ArduinoOTA.h>               // local OTA update from Arduino IDE
+#include <ArduinoOTA.h> // local OTA update from Arduino IDE
 #include <FS.h>
 
 #include "lcd.h"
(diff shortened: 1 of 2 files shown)


jeremypoulter:
Can you give some more details on the problems you where trying to solve? Particularly the need for CORS, that really shouldn't be needed any more.

arjanmels:
When connecting to "openevse.local", the lines starting at https://github.com/OpenEVSE/openevse_wifi_gui/blob/90b266bf71408ee5bef1da4cf1fcd4aefb13b33d/src/view_models/OpenEvseWiFiViewModel.js#L228 will replace the baseHost name ("openevse.local") with the ip address. Subsequently this ip address is used internally instead of the hostname and then the CORS problem with static files (zones.json) arises.

Second problem arises when using espota option to upload the firmware from visual studio code via platform.io: the watchdog gets triggered, because the update takes too long (and doesn't return to the main loop inbetween).

jeremypoulter (approved):
@glynhudson these changes look good to me

PR #127, opened 2020-11-01
Remove the legacy API and ENABLE_LEGACY_API flag

toofishes:
This flag was first introduced in 2017, and now that we've moved from
ESP8266 to ESP32, this feels like a good time to get rid of this code.
It is not enabled by default in any build configuration, and a lot of it
was gutted in an earlier commit 48cc8e064.

This endpoint has been merged with the status endpoint since 2017 in
commit 54987e6bdaa.

@jeremypoulter, I would definitely appreciate your input here on whether this is a good idea, and if not, why this should stick around as I'm not sure what would still be expecting to use this.


jeremypoulter:
As you mentioned a bunch of this has been removed and was of the same thinking but just slowly removing as I was changing similar areas, especially where it was making the code more complex.

PR #129, opened 2020-11-20
Adding support for smartEVSE

arjanmels:
The accompanying pull request for the gui: https://github.com/OpenEVSE/openevse_wifi_gui/pull/42

This pull request adds support for particular features of smartEVSE: https://www.smartevse.nl/
(The current/default version of EPS32_WiFi_V3.x works, but does not allow control of all smartEVSE features)

The RAPI interface for smartEVSE is implemented in: https://github.com/SmartEVSE/smartevse/pull/39

Let me know what you think about the pull request in its current form: would you indede be willing to consider merging this in one way or form into this repo, or do you prefer me to keep it in a separate fork?


glynhudson:
Hi thanks for your PR. It's cool that you have the WiFi module working on the smartEVSE. 

I've had a look at the changes and I think it would be best to keep them as separate forks, since quite a few changes are required. There also seems to be some changes in this PR which are not directly related to smartEVSE support.

What do you think @jeremypoulter?

jeremypoulter:
That may be wise for now, but we could consider adding official support for different controllers, I certainly have it in the back of my mind in the current work.

PR #132, opened 2020-12-07
Don't allow voltage to be set outside of 60-300 volts via MQTT

toofishes:
Voltage is used by divert mode, as well as sent along to the OpenEVSE
module for power calculations.

Setting this to zero could result in a lot of issues. The divert code
would see divide by zero happen, and the energy montoring code on the
OpenEVSE module would result in the recorded energy usage being
incorrect.

While this shouldn't happen reguarly, I found myself in this situation
by sending a null/zero-length message to the relevant topic that the
ESP32 was listening on. Because there is no error checking in the float
parsing code, a null value parsed to zero, which resulted in the voltage
value being set to zero.

An alternative here would be to reset the voltage to `DEFAULT_VOLTAGE`
if a zero or invalid value was received. As written, this will simply do
nothing and ignore the invalid/zero value. Thoughts here? Happy to
change this PR; straightforward to add one line to the else case:
`voltage = DEFAULT_VOLTAGE;`

Example after running `mosquitto_pub -n -t my/topic/vrms`:

zero-voltage

There might be additional validation/verification wanted around the solar and grid I/E feeds as well. However, this is a bit trickier, as zero and negative values are possible, so we'd likely want to move away from `toInt` and use `strtol` or something directly to be able to detect error cases. Alternatively, we can just ensure the value to be processed is not the empty string. Should I attempt to catch and handle these cases as well?

### DIFF
--- a/src/mqtt.cpp
+++ b/src/mqtt.cpp
@@ -57,11 +57,16 @@ void mqttmsg_callback(MongooseString topic, MongooseString payload) {
     divert_update_state();
   }
   else if (topic_string == mqtt_vrms){
-    voltage = payload_str.toFloat();
-    DBUGF("voltage:%.1f", voltage);
-    OpenEVSE.setVoltage(voltage, [](int ret) {
-      // Only gives better power calculations so not critical if this fails
-    });
+    double volts = payload_str.toFloat();
+    if (volts > 0.0) {
+      voltage = volts;
+      DBUGF("voltage:%.1f", voltage);
+      OpenEVSE.setVoltage(voltage, [](int ret) {
+        // Only gives better power calculations so not critical if this fails
+      });
+    } else {
+      DBUGLN("Ignoring voltage (was <= 0.0)")
+    }
   }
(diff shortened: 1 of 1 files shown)


toofishes:
It might also make sense to further restrict this to sane values, in case someone thought it should be specified in millivolts instead of volts, which would also cause quite a few issues? e.g. only update the internal value if this check passes:
`if (volts >= 60.0 && volts <= 300.0) {`
or similar. I'm not totally sure what sane cutoff values would be.

glynhudson:
Good idea, I think constraining to sane values makes sense to me. The values you choose above for sane values look sensible. What do you think @chris1howell @jeremypoulter ?

jeremypoulter:
I remember discussing doing this when we added the feature, I don't recall why it was not added, so I am fine with the change.

toofishes:
I've updated the check to accept a value in the 60-300 volts range. Thanks for the feedback, @glynhudson and @jeremypoulter !

PR #182, opened 2021-04-16
Integration of ArduinoOcpp

matth-x:
First draft integration of OCPP 1.6. This code is able to make a test-connection to SteVe, and to handle some basic OCPP operations (Start- and StopTransaction, MeterValues, SmartCharging)


jeremypoulter (commented):
Some initial comments

glynhudson:
Hi @matth-x 

Nice work with this OCPP implementation 👍 

I've had a go at testing it, I've been able to successfully connect it to an OCPP server. However, I've not had any luck sending commands to the unit, but this is probably because I'm not too familiar with using Steve and the options available on OpenEVSE OCPP config page. 

Would it be possible to arrange a video chat with you to go over the feature? I think it would be good to add some pop-out into boxes on the UI, to help explain what all the options do. 
 
Screenshot 2021-07-23 01 35 54

I've created a beta build (V4.1.0) with includes your OCPP PR: https://github.com/OpenEVSE/ESP32_WiFi_V4.x/releases/tag/4.1.0

matth-x:
Hi Glyn, thank you for the compliment :)

Yes, some documentation on the configurations would definitely be a good idea. We can have a call next week, I send you the possible time slots per pm.

The ID Tag is used to identify the user who is about to start the charging session. In our case that function is obsolete (because the CP doesn't have an RFID reader) but SteVe requires a valid user entry in its DB nevertheless. You can activate it in the ID Tag page by searching for the blocked ones (SteVe creates a "default entry" for each unknown tag) and setting its maximal concurrent usages from 0 to 1.

glynhudson:
Thanks! After creating a tag and user in Steve I was able to send a remote start request 👍 

Screenshot 2021-07-23 15 22 45

PR #198, opened 2021-06-11
Adding a GUI for configuring the vehicle data source

### LINKED ISSUE #71: Add Vehicle Config UI Tab and SoC display
After https://github.com/OpenEVSE/ESP32_WiFi_V3.x/pull/67 is merged we should make a 'Vehicle' tab on the web UI interface to interface with the Tesla integration e.g set enter username / password etc. I've added the API call for the Tesla integration as a comment in #67 

The vehicle tab should be structured to allow for multiple vehicle integrations, probably with only a single integration being used at a time..for the moment.  

Once the vehicle integration is active the current vehicle SoC can be displayed on the main page, it would probably be a good idea to provide an indicator as to when this was last updated (like we currently with solar PV divert). 

OVMS should be the next integration to consider adding.

jeremypoulter:
Fixes #71

Addin support for getting configuring the vehicle data source. Getting data from the Tesla API and MQTT are currently supported. 

Displayed on the UI in the new vehicle tab as well as on the LCD.


glynhudson:
Hi, @jeremypoulter this looks great. I'm keen to test it. How can I resolve these conflicts? Will rebuilding the GUI solve them?

jeremypoulter:
Ah... Had not noticed those, will check but yeah probably just a GUI build

Issue #205, opened 2021-07-04
Failure to connect to AP when poor signal


julianwb:
It seesm that this is a known problem with ESP-IDF and WiFi.
Symptoms: frequently attempts to connect to AP result in STA_START, STA_DISCONNECTED, AUTH_FAIL.   At the third attempt the code reverts to AP mode; behaviour I find to be somewhat obscure, as the button is still available for such a choice.
Resetting the WiFi 
>      WiFi.persistent(false);
>      WiFi.disconnect();
>      WiFi.mode(WIFI_OFF);
>      WiFi.mode(WIFI_STA);
and re-attempting the connection succeeds.

ref: <https://github.com/espressif/arduino-esp32/issues/2501>

chris1howell:
95%+ of OpenEVSE based stations do not have an external button, the button is optional and presents a significant security concern for publicly accessible stations. Not only can settings be changed but the WiFi control could be hijacked onto another network especially if the password is not set or discovered via bruteforce.

The automatic starting of AP mode is vital for users that incorrectly enter their password during initial configuration or take their station on the road.

I think it would be fine to try a few more times and wait a little longer between atempts especially if the change included the information on the display for each atempt and starting of AP mode.

PR #206, opened 2021-07-04
Reset WiFi on AP Connection failure

julianwb:
#205 Failure to connect to AP when signal is poor.
Replace existing behaviour: reset WiFi and retry connection at third failure instead of invoking AP mode (which may still be achieved through use of button).

NB This succeeds whereas 5 minute AP mode expiring and resetting the ESP32 does not, this results in a continuous loop as the failure to connect persists.
As an alternative (others may judge preferable) AP mode may be invoked for 5 mins as previously - then replace line #487 net_manager.cpp

> ESPAL.reset();

with

>       WiFi.persistent(false);
>       WiFi.disconnect();
>       WiFi.mode(WIFI_OFF);
>       WiFi.mode(WIFI_STA);
> 
>       DEBUG.println("Had 3 disconnects - resetting WiFi");
>       client_disconnects = 0;
>       startClient();
> 

should the temporary AP mode be judged important.

### DIFF
--- a/src/net_manager.cpp
+++ b/src/net_manager.cpp
@@ -467,11 +467,16 @@ net_loop()
   // Manage state while connecting
   if(isClientOnly && !WiFi.isConnected())
   {
-    // If we have failed to connect turn on the AP
-    if(client_disconnects > 2) {
-      startAP();
-      client_retry = true;
-      client_retry_time = millis() + WIFI_CLIENT_RETRY_TIMEOUT;
+    // Reset WiFi and try again
+    if (client_disconnects > 2) {
+      WiFi.persistent(false);
+      WiFi.disconnect();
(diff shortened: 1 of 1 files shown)


chris1howell:
The automatic starting of AP mode is vital for users that incorrectly enter
their password during initial configuration or take their station on the
road.

julianwb:
Thank you for the comprehensive response; I had not, as you have seen, considered the points you raise.
I will update this PR to perform as you suggest; further attempts with perhaps a longer delay between, I will also include the WiFi reset code, without which this does not work for me.
I don't have a display currently, if I can add one sensibly I will also implement your suggested updates to that,

jeremypoulter:
Superseded by #209, but thanks for the contribution this is a great find.

PR #209, opened 2021-07-11
Reset the WiFi state after receiving a disconnect event

jeremypoulter:
Fixes Failure to connect to AP when poor signal #205. Based on findings by @julianwb in #206.


jeremypoulter:
@julianwb I think this will resolve the reconnect issue, but I have not had a chance to test. It resets the WiFi after every disconnect, but still starts the AP. I also added the ability to (build time) configure the number of disconnects before starting the AP, `WIFI_CLIENT_DISCONNECTS_BEFORE_AP`

PR #221, opened 2021-07-24
Fixes to improve the (re)loading of the GUI after update

jeremypoulter:


glynhudson:
Nice work, I've just tested. The GUI didn't seem to automatically reload after updating, however when I manually reloaded the page didn't hang 👍 

However, I was updating from V4.1.0 beta

PR #242, opened 2021-08-30
Remove CORS and associated `.local` redirect

### LINKED ISSUE #231: CORS - OpenEVSE save broken when connecting using hostname
If you access the OpenEVSE unit via a hostname. saving settings fails due to CORS

Impacted versions:
* OpenEVSE 4.0.1
* OpenEVSE 4.1.0

Screen Shot 2021-08-03 at 10 47 46 AM

The cause seems to be the request using the IP address of the unit directly instead of the hostname the unit was accessed via.

jeremypoulter:
Fixes #231


glynhudson:
Great. Could you fix the conflict? Thanks

jeremypoulter:
@glynhudson done

PR #245, opened 2021-09-09
v4 API documentation

### LINKED ISSUE #9: API documentation
We should write some API documentation on the HTTP API, would probably be a good idea to use [OpenAPI](https://www.openapis.org/) and [Swagger](https://swagger.io/)

jeremypoulter:
Fixes #9


jeremypoulter:
@glynhudson / @chris1howell what do you think is the best way forward with this? Maybe we merge and then make sure new features are added to the docs and also flesh out with with other PRs, or is there anything that specifically needs adding first, maybe some content from the README/docs could do with merging/updating into this?

glynhudson:
I'm happy for it to be merged and we can add into as we go along. 

Should we remove all API documentation from the readme and direct all users to https://openevse.stoplight.io/docs/openevse-wifi-v4/ZG9jOjQyMjE5ODI-open-evse-wi-fi-esp-32-gateway-v4 for latest API docs?

jeremypoulter:
Yeah probably, I will tidy up th ReadMe probably wants splitting in to user guide, development guide and API (and probably also process doc as well)

jeremypoulter:
@glynhudson  I think this is ready for review now

glynhudson (approved):

PR #255, opened 2021-10-26
OCPP: integration of further functions

matth-x:


matth-x:
Thank you Jeremy! Then I can integrate it next week right away.

matth-x:
Should be ready now. Soft reset and FW update are integrated now as well.

chris1howell:
Did this PR also include the heartbeat fix?

matth-x:
Yes, both the WebSocket and OCPP heartbeat work, I've checked them. Probably the issue was that my library was ignoring the OCPP heartbeat interval when it was set with the BootNotification. I have already solved that. But I am not 100% sure if that was the actual reason.

PR #296, opened 2022-01-05
Updating the firmware to handle configuring the EVSE

### LINKED ISSUE #140: DEV V4.0: LCD does not update max current in 'Ready' mode
In 'Ready or mode the LCD displays "Ready 32A". However, if the maximum current is adjusted on the interface e.g 6A the LCD still shows 32A. I think this is because the Pilot value still seems to be 32A: 

OpenEVSE

### LINKED ISSUE #178: Sensor Value: Pilot not reflecting right value > 11A
Not certain what is the source data for the value Pilot under the Sensor block display but seems to be inconsistent and perhaps misleading since there is no sensor I am aware for the Pilot.

While operating in Eco Mode, with no timers, and a maximum Value set to 16A under Setup, noticed that the Session current reflects ~16A, but the Sensor Pilot Value seems not to increase above 11A. Checked my Tesla for charging current and it shows 16A.    Also Current Value under Session Status ~16A.

jeremypoulter:
Moved all the 'setting' related RAPI commands from the UI to the backend to fix various issues with the EVSE settings changing without the WiFi module knowing. This was mainly to resolve issues around the current not being displayed correctly on the LCD and in the UI.

- Fixes #140 
- Fixes #178

This change also include some updates to the RAPI tab. You can now click a button and call all the `$Gx` commands to get the current state of the EVSE module.


glynhudson:
Nice, I like the new feature of being able to see the result of all the $Gx commands 


However, the 'pause' button seems to be broken. When I click pause, nothing happens and the error `home.js:1 Uncaught TypeError: o.openevse.isPaused is not a function` is displayed in the console. 

Screenshot 2022-01-29 00 37 21

PR #322, opened 2022-02-27
Updated GUI with installer changes

### LINKED ISSUE #318: Home - Add Status Block and Max Current setting
Make the initial page more friendly for an installer.

Add status block and max current setting to home page, so an installer can take care of all mandatory initial configuration tasks on a single screen without requiring entering Standalone mode or restarting and reconnecting in Client mode.

Add The Status block:
Station State - Color
Current - Energy - Temp - Elapsed

Add Max Current setting 

This will allow an installer to verify station operation, set maximum current and connect the station to WiFi all from a single page. If the installer is a hired professional (not the station owner) it will allow all mandatory settings without having to connect the installer's device to the client's WiFi.

jeremypoulter:
Fixes #318 Home - Add Status Block and Max Current setting


chris1howell:
Looks great... One issue, the list of current will not go higher than the setting currently. The list should be MIN to MAX. Looks like it in Min to setting now.

glynhudson:
I've just tested this, it seems to work well. However, each time the captive portal loads for a few seconds and EVSE Error appears. This only appears for a split second and does not affect functionality, however it would be best if the error was not displayed at all. 

See video: https://drive.google.com/file/d/1NeK4fW2ZYyMN3Pi5RUCnX6L16oGinw8d/view?usp=sharing

Screenshot:

Screenshot_20220308-174032

jeremypoulter:
The current issue is a generic issue, #326 and fixed by #329, I will add a loading screen similar to the main UI to hide the errors. If/when we do a new UI I think we should just merge this into the same HTML code, getting a lot more cross over.

PR #329, opened 2022-03-11
Properly handle setting the max current by setting in the value in the EVSE EEPROM rather than the WiFi config

### LINKED ISSUE #326: V4.1.2 limits max current selection
I just downgraded from 4.1.2 to 4.1.1 because somehow, the power dropdown decided to limit itself to 15A as a maximum.  Before doing the update, the power was set to 15A for an overnight charge session, but then I did the update and that was the highest I could go.

jeremypoulter:
Fixes #326


DariusPL:
6A is being set as default during each WIFI loading regardless of actual max. current settings.

PR #341, opened 2022-04-04
LCD Display: Display Energy formatted to significant figures instead of decimal digits.

rbsexton:
The existing FW scales energy to SI notation and then places one digit after the decimal point.

This doesn't work so well for low values in kWh, ie 1.3kWh.

The ADCs in the system deliver about 3 significant figures, so display this number to a total of four - ie:
```
1.305kWh
13.05kWh
130.5kWh
```

### DIFF
--- a/src/lcd.cpp
+++ b/src/lcd.cpp
@@ -675,6 +675,20 @@ void LcdTask::displayScaledNumberValue(int line, const char *name, double value,
   displayNumberValue(line, name, value, precision, newUnit);
 }
 
+// Display a scaled number and adjust output such that the user gets some number 
+// of significant figures eg, for 4: 1.00598kWh appears as 1.005 rather than 1.0
+// Overloaded on displayScaledNumberValue
+void LcdTask::displayScaledNumberValue(int line, const char *name, double value, const char *unit)
+{
+  
+  char newUnit[20];
+  sprintf(newUnit, "%s%s", ScaleNumberSI(&value), unit);
+
+  displayNumberValueSigFigures(line, name, value, 4, newUnit);
+}
+
+// Display a numeric value preceded by the quantity name, scaled for SI notation, 
+// and with a specific number of digits after the decimal point.
 void LcdTask::displayNumberValue(int line, const char *name, double value, int precision, const char *unit)
 {
   char number[20];
--- a/src/lcd.cpp
+++ b/src/lcd.cpp
@@ -643,6 +643,28 @@ const char *LcdTask::ScaleNumberSI(double *value) {
   }
 }
 
+// Format a number to a specif number of signficant figures. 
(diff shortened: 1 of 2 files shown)


jeremypoulter:
@rbsexton thanks for the contribution, been something I have been meaning to do this for a while. Will check it over.

src/lcd.cpp
```diff
@@ -619,30 +619,77 @@ void LcdTask::displayInfoLine(LcdInfoLine line, unsigned long &nextUpdate)
...
   {
-    value /= 1000;
-    index++;
+    static const char *mod[] = {
+      "",
+      "k",
+      "m",
+      "g",
+      "t",
+      "p"
```
toofishes, line 632: I opened another PR before seeing this one to correct the capitalization of these units - might be worth updating it here since you're moving these around? https://github.com/OpenEVSE/ESP32_WiFi_V4.x/pull/344

jeremypoulter: Not a fan of multiple changes per PR but #344 will conflict with this PR and visa-versa so ok if @rbsexton wants to add the unit update

src/lcd.cpp
```diff
@@ -619,30 +619,77 @@ void LcdTask::displayInfoLine(LcdInfoLine line, unsigned long &nextUpdate)
...
   char newUnit[20];
-  sprintf(newUnit, "%s%s", mod[index], unit);
+  sprintf(newUnit, "%s%s", ScaleNumberSI(&value), unit);
 
   displayNumberValue(line, name, value, precision, newUnit);
 }
 
+// Display a scaled number and adjust output such that the user gets some number 
+// of significant figures eg, for 4: 1.00598kWh appears as 1.005 rather than 1.0
+// Overloaded on displayScaledNumberValue
```
jeremypoulter, line 681: I think this example highlights a flaw with the implementation. `1.00598kWh` should be rounded up to `1.006kWh`

jeremypoulter (commented):
Thanks for the contribution, this is something I have been meaning to do for ages. I have added a few comments

src/lcd.cpp
rbsexton: This one has been laying around!   I'd forgotten all about it.      I'll update this so that it can get pulled.

rbsexton: Ok.   I'm pretty sure that the merge of PR#344 made mine unmergable.  I'm going to start over and I'll fix the rounding error while I'm at it.

src/lcd.cpp
rbsexton: Good catch.  printf() doesn't round.   I'll add that.

jeremypoulter:
We have replaced the local UI with an LVGL UI, so this is no longer applicable

PR #344, opened 2022-04-17
LCD info display corrections and date format tweak

toofishes:
The metric units fix here should be uncontroversial- we need to be using `MWh`, not `mWh`, to be megawatt-hours, not milliwatt-hours. https://en.wikipedia.org/wiki/Metric_prefix#List_of_SI_prefixes

The date one might be a bit less desired, but I'd advocate for making the change. The OpenEVSE web UI has the advantage of being able to use the browser/OS date format, but the LCD display format is hardcoded. By using the ISO date format, there should be no confusion with a date such as "04/07/2022" - is this April 7th (as a US resident would likely interpret it), or July 4th (as most of the rest of the world would use)? By displaying "2022-07-04", it is much less ambiguous.

### DIFF
--- a/src/lcd.cpp
+++ b/src/lcd.cpp
@@ -546,7 +546,7 @@ void LcdTask::displayInfoLine(LcdInfoLine line, unsigned long &nextUpdate)
       gettimeofday(&local_time, NULL);
       struct tm timeinfo;
       localtime_r(&local_time.tv_sec, &timeinfo);
-      strftime(temp, sizeof(temp), "%d/%m/%Y", &timeinfo);
+      strftime(temp, sizeof(temp), "%Y-%m-%d", &timeinfo);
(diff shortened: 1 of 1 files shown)


src/lcd.cpp
```diff
@@ -546,7 +546,7 @@ void LcdTask::displayInfoLine(LcdInfoLine line, unsigned long &nextUpdate)
       gettimeofday(&local_time, NULL);
       struct tm timeinfo;
       localtime_r(&local_time.tv_sec, &timeinfo);
-      strftime(temp, sizeof(temp), "%d/%m/%Y", &timeinfo);
```
jeremypoulter, line 549: I think this is more of a personal taste/country specific convention, so should probably handle separately, probably needs to be user configurable.

PR #345, opened 2022-04-17
Cleanups

toofishes:
Remove a few unused global variables, and mark a few other functions and variables static that are only used in a single file.


jeremypoulter (approved):
Changes look good to me, thanks for the contribution.

PR #363, opened 2022-05-31
Fixed broken/hi-jacked url from openevse.org to openevse.com

d3wy:
Just fixing up a broken link


jeremypoulter:
Good find, thanks for the contribution.

jeremypoulter (approved):

PR #367, opened 2022-06-02
fix #366 : prevent charge_rate to be over max charge current

### LINKED ISSUE #366: Divert mode set charge rate above L2 max current
When power available exceed max current in divert mode, charge_rate is not caped to max current.
Consequence is charge_rate over mqtt or api displays a value superior to max charge capability.

KipK:
fix #366

### DIFF
--- a/src/divert.cpp
+++ b/src/divert.cpp
@@ -178,9 +178,14 @@ void divert_update_state()
     DBUGVAR(smoothed_available_current);
 
     charge_rate = (int)floor(available_current);
+    
     // if the remaining current can be used with a sufficient ratio of PV current in it, use it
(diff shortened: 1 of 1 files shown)


jeremypoulter:
Thanks for the contribution, however it is not quite that simple. There are several factors that affect the max current. I also don't think this is a problem. The EVSE  manager should filter this value down so what actually gets set on the pilot should be correct. So this may just be a legacy/badly named field that needs removing/renaming. FYI the actual charge rate is `pilot`

KipK:
I see. 
So Pilot will be. :)

KipK:
this field is from the divert code. charge_rate is only published when divert is on.
Can we consider to remove its publication to mqtt ?

PR #207, opened 2021-07-04
Platform IO ESP32 core 4.4

jeremypoulter:
Uplift to the latest ESP32 Platform IO core, including ESP IDF 4, this has a lot better debug features, mainly builtin stack trace decoding. 

Fixes #353

PR #369, opened 2022-06-03
Extend MQTT to set state / max_current / charge_current

### LINKED ISSUE #358: MQTT 'API'
We are deprecating RAPI access, at least the `$Sx` and `$Fx` commands, and replacing with a higher level HTTP API because a lot of the 'smart' feature will now be implemented on the more powerful ESP32. 

Currently none of this is exposed to MQTT, and to a large extent probably won't be as MQTT is not really designed for APIs. However there are a few use cases where controlling the EVSE via MQTT may have advantages over HTTP so the following should be made available:

- Max current
- Charge current
- Manual Override - Start/Stop charging
- Divert mode (existing API)

KipK:
Fixes #358

mqtt: set override ( start/stop/delete)
mqtt: publish max_current
mqtt  set charge_current + max_current

 Not sure of my choices, let me know if some approach are bad.

perhaps should I rename /charge_current/set as /pilot/set to stay coherent ?

### DIFF
--- a/src/mqtt.cpp
+++ b/src/mqtt.cpp
@@ -10,6 +10,7 @@
 #include "espal.h"
 #include "net_manager.h"
 #include "web_server.h"
+#include "event.h"
 
 #include "openevse.h"
 
@@ -106,6 +107,43 @@ void mqttmsg_callback(MongooseString topic, MongooseString payload) {
     if ((newdivert==1) || (newdivert==2)) {
       divertmode_update(newdivert);
     }
+  }
+  else if (topic_string == mqtt_topic + "/charge_current/set")
+  {
+    int newchargecurrent = payload_str.toInt();
+    DBUGF("Set charge_current: %d", newmode);
+    //using override, the only interest of this is having priority above everything else.
+    //We have a confusion in web interface to solve as details of the override are not displayed
+    EvseProperties props(EvseState::Active);
+    props.setChargeCurrent(newchargecurrent);
+    evse.claim(EvseClient_OpenEVSE_Manual, EvseManager_Priority_Manual, props);
+  }
+  else if (topic_string == mqtt_topic + "/max_current/set")
+  {
+    //not using override here, this is a main config change.
+    int newmaxcurrent = payload_str.toInt();
+    DBUGF("Set max_current: %d", newmode);
+    evse.setMaxConfiguredCurrent(newmaxcurrent);
+  }
+  else if (topic_string == mqtt_topic + "/manual_override/set") 
+  {
+      DBUGF("Set evse state: %d", newmode);     
+      if (payload_str.equals("stop")) {
+        // stop/pause using override like web interface 
+        EvseProperties props;
+        props.setState(EvseState::Disabled);
+        evse.claim(EvseClient_OpenEVSE_Manual, EvseManager_Priority_Manual, props);
+      }     
+      else if (payload_str.equals("start")) {
+        // start/unpause using override like web interface 
+        EvseProperties props;
+        props.setState(EvseState::Active);
+        evse.claim(EvseClient_OpenEVSE_Manual, EvseManager_Priority_Manual, props);
+      }
+      else if (payload_str.equals("delete")) {
+        // remove override
+        evse.release(EvseClient_OpenEVSE_Manual);
+      }      
   }
   else
   {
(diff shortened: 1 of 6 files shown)


jeremypoulter:
Thanks for your contribution and tackling one of the bigger issues. I will review it shortly.

KipK:
I have rebased it in one commit. There's probably some choices to discuss.

src/mqtt.cpp
```diff
@@ -107,6 +108,43 @@ void mqttmsg_callback(MongooseString topic, MongooseString payload) {
...
   }
+  else if (topic_string == mqtt_topic + "/charge_current/set")
+  {
+    int newchargecurrent = payload_str.toInt();
+    DBUGF("Set charge_current: %d", newmode);
+    //using override, the only interest of this is having priority above everything else.
+    //We have a confusion in web interface to solve as details of the override are not displayed
+    EvseProperties props(EvseState::Active);
+    props.setChargeCurrent(newchargecurrent);
+    evse.claim(EvseClient_OpenEVSE_Manual, EvseManager_Priority_Manual, props);
```
jeremypoulter, line 119: I think the priority should be the same as the HTTP API and there should probably be a specific client for these changes, eg

```suggestion
    evse.claim(EvseClient_OpenEVSE_MQTT, EvseManager_Priority_API, props);
```

src/mqtt.cpp
```diff
@@ -107,6 +108,43 @@ void mqttmsg_callback(MongooseString topic, MongooseString payload) {
       divertmode_update(newdivert);
     }
   }
+  else if (topic_string == mqtt_topic + "/charge_current/set")
+  {
+    int newchargecurrent = payload_str.toInt();
+    DBUGF("Set charge_current: %d", newmode);
+    //using override, the only interest of this is having priority above everything else.
+    //We have a confusion in web interface to solve as details of the override are not displayed
```
jeremypoulter, line 116: The GUI could be updated to show that the EVSE is being controlled be an external source and use the `/claims` API to get the info

src/mqtt.cpp
```diff
@@ -107,6 +108,43 @@ void mqttmsg_callback(MongooseString topic, MongooseString payload) {
       divertmode_update(newdivert);
     }
   }
+  else if (topic_string == mqtt_topic + "/charge_current/set")
```
jeremypoulter, line 111: This does not match the subscribed topic:

```suggestion
  else if (topic_string == mqtt_topic + "/pilot/set")
```

jeremypoulter (changes requested):

src/mqtt.cpp
jeremypoulter: Also the GitHub workflows look to have been broken by Platform IO so there is a dependency on #207

jeremypoulter:
Would also be great to add a doc for the API as `docs/mqtt.md`

src/mqtt.cpp
KipK: Agreed, that is the way. I have not gave a look yet the the javascript gui part as it's embedded in .h file, looks unreadable :)

src/mqtt.cpp
KipK: you mean I can merge #207 in my base main ?

KipK:
> 

was on it ;) it's pushed

src/mqtt.cpp
KipK: I was using a dedicated MQTT id first, then removed it, to match what http interface did :)
It seems http://openevse/override use EvseClient_OpenEVSE_Manual, EvseManager_Priority_Manual ( I've copied it )

KipK: I've pushed it with new EvseClient_OpenEVSE_MQTT, EvseManager_Priority_MQTT (500)

KipK:
Also, MQTT data should be updated just after a set. For now we need to wait for the new loop to have feedback.
What would be the best way to do it in a centralised way, without rebuilding a shaped event request for each function ?

src/mqtt.cpp
KipK: I've tested, but it has collaterals.
Now, as priority has changed, it is not considered as an override, but just a claim.
So on web interface we can't see and delete it.
Also manual_override is not set to 1 then on mqtt. 

Should we expose claims over mqtt ? 
Do you have some suggestions ?

KipK: I've opened an issue for this to not spam too much this page : https://github.com/OpenEVSE/ESP32_WiFi_V4.x/issues/373

KipK:
I'm closing it to open a wip new one based on http API for discussion.

PR #378, opened 2022-06-11
fix priority issue between scheduler and divert

KipK:
this revert https://github.com/KipK/ESP32_WiFi_V4.x/commit/82e52c4fdaa0e338ea4617bacdf9d09b12a7decd for introducing priority issue


jeremypoulter:
@KipK can you merge in `master` now the ESP32 core updates have been merged

glynhudson:
Hi @KipK, could you provide a bit more context about the priority issue this PR is addressing?

KipK:
@glynhudson , it's discussed here https://github.com/KipK/ESP32_WiFi_V4.x/commit/82e52c4fdaa0e338ea4617bacdf9d09b12a7decd

jeremypoulter:
@KipK I think the current code is working as expected, both the scheduler and divert have a weak disable to allow each other (or any other systems) to activate the charging when appropriate. Can you describe the usecase that you think is not working?

KipK:
@jeremypoulter , using divert mode as dynamic charge current balancing. Evse has lowest priority over my house consumtion, so I let it modify charge current using what available amp are left. But I only let the charge possible between 22:30 and 6h30 as energy is half the price. 

So can't use divert mode with timers withotu changing this.

jeremypoulter:
@KipK ah I thought that might be the case, I have a different plan for implantation of that feature. To do that we need some code similar to the divert code but needs to have some subtle differences.

1) The MQTT feed should provide the available current
2) Should set the max current 
3) Should have a high priority disable and a low priority (default) active
4) Probably no smoothing 
5) should revert to a failsafe mode when no MQTT signal

jeremypoulter:
Happy to accept a PR with just the ESP32 part mostly do that bit first anyway, as long as we have example .http files, it is easy enough to develop/test without the UI bit.

KipK:
@jeremypoulter , First iteration here : https://github.com/OpenEVSE/ESP32_WiFi_V4.x/pull/382

I've made it a Class, event if Divert is not.

Issue #350, opened 2022-04-30
Default time schedule and random delay UK legislation


glynhudson:
[New UK legislation](https://www.legislation.gov.uk/uksi/2021/1467/contents/made) which is taking effect from June 2022 stipulates that all EVSEs should come pre-configured by default to charge outside of peal hours. “peak hours” means 8am to 11am and 4pm to 10pm.

The charge point should also be set by default to turn on with a randomised delay of up to 10min. if the timer delay is set to switch on at 10pm the random delay will mean it should switch on randomly at some time from 9:50pm - 10:10pm. 

The user should be able to override the timer and the delay if immediate charging is required. The user can opt out from timed charging if they wish. 

To comply with legislation, I propose we preset a time charge on the EVSE between 10pm-8am and create a 10min randomised delay as mentioned above. The user should be able to switch off the random delay via the UI and override the timer and delay using a physical push button.

jeremypoulter:
That is an interesting requirement, kinda makes sense. Don't think it will be too difficult.

glynhudson:
Indeed, but I can understand the rational. The gov are keen to try to persuade EV owners to charge at off-peak times. Having all new EVSEs to have off-peak timers switched on by default will help with this. The randomised delay is there to avoid having a large load hitting the grid exactly at the same time. 

All this makes sense and in the long run will save everyone money, user are free to opt out of all of this if they would prefer to charge immediately.

Tifaifai:
After thinking about it, I understand the logic.
For the physical push button, it is perhaps the occasion to think about the integration of a RFID / NFC reader to be the entry point of authorizations and / or forced triggering of a recharge. (Little kit ;) )

jeremypoulter:
@chris1howell will this be an OK default for the USA? @glynhudson Same question for other EU countries?

@Tifaifai Initial RFID support has been merged to the main branch

jketcham:
I think if perhaps during initial setup of the EVSE, users could be presented with the default schedule and customize it (or disable it) that would be nice. Or at least make it very obvious how to change the default schedule.

My power company in Phoenix, AZ has an opt-in EV charging program with cheaper rates from 11pm to 5am year round, but no stipulation on a random delay, although I think that's a good idea too.

zymurgic:
[...]

PR #379, opened 2022-06-11
Mqtt reworked API version

### LINKED ISSUE #358: MQTT 'API'

### LINKED ISSUE #373: MQTT manual modes , also Reorganising MQTT topics
Let's discuss about https://github.com/OpenEVSE/ESP32_WiFi_V4.x/pull/369#discussion_r890166699 here, as it involve bigger changes.

Here is my proposal

So when MQTT set something that is not a configuration data, it should be a claim, that could be considered on mqtt side as such "manual_override" , let's call it "manual" then, to not be confusing with manual_override from the api ( using the highest priority manual level ).

/manual[0,1]				(0: auto/no manual 1: either manual charge_current or status are active)
/manual/set 0               (0: delete all related claims, can't set 1, there's no purpose I think?)
/manual/charge_current      (0: auto/no claim, other value in Amp )
/manual/charge_current/set  ( set to 0 delete the charge_current claim, other value in Amp )
/manual/status 				( 0: auto/no claim, 1: enabled/start, 2: disabled/sleep )
/manual/status/set			( set to 0 delete the enabled/disabled claim, 1: enabled/start, 2: disabled/sleep )

Considering this, we could also reorganize a bit the mqtt topics and remove duplicated values.
 Something like that ? 
 
/announce
/announce/[id]

/state [0,1,...]
/state/colour [0,1,..]
/state/flags [int]
/state/vehicule [0,1]
/state/manual_override [0,1]

/smartmode [0,1]  are smart features enabled or not ( could be solar divert / Grid +i-E  / other incoming modes]
/smartmode/set [0,1] enable/disable smart features. Replace divertmode/set( could be solar divert / Grid +i-E  / other incoming modes]
/smartmode/divert_update
/smartmode_grid_ie
/smartmode/available_current
/smartmode/smoothed_available_current
/smartmode/charge_rate ( should be deleted, just rounded available_current, confusing with pilot ) 

/charge/pilot
/charge/current
/charge/session_kwh
/charge/total_kwh

/system/temp
/system/tempwifi
/system/voltage
/system/max_current
/system/max_current/set
/system/srssi
/system/freeram

### LINKED ISSUE #376: claim engine should deep merge properties
When making a new claim over an existing one, it doesn't keep previous properties.
It should deep merge data to avoid sending again the whole json. This will be a pain to integrate mqtt feeds to smart home solutions if we need to resend all claim data for each separated command.

related to this WIP here : https://github.com/KipK/ESP32_WiFi_V4.x/commit/d9979dd18a95e9ccda0d8777304f45270707d693

KipK:
Fixes #358 
Fixes #376 
Fixes #373 

Following previous discussion here https://github.com/OpenEVSE/ESP32_WiFi_V4.x/pull/369 
Based on espidf4 branch.
 
Edit: all functional


jeremypoulter:
I haven't forgotten about this PR, just trying to fix a few issues with #207

KipK:
I need some helps on this one. 
I'd like to integrate real time update of modified values on mqtt instead of the 30sec loop pushing the whole dataset.
For ex, "pilot" value should be notified when setting a charge_current override or claim. waiting 30 sec for feedback is inconvenient.
Where should I start ? 
This will will prolly involve also this https://github.com/OpenEVSE/ESP32_WiFi_V4.x/issues/376

jeremypoulter:
Just calling event_send should be all you need. That will then send the new values.

The double polling/web socket is just a legacy thing, going to remove the polling at some point.

docs/user-guide.md
```diff
@@ -136,6 +136,8 @@ Data can be posted using HTTP or HTTPS.
 
 ### MQTT
 
+Refer to MQTT.md documentation
```
jeremypoulter, line 139:
```suggestion
Refer to [MQTT API documentation](mqtt.md)
```

src/input.cpp
```diff
@@ -91,6 +94,7 @@ void create_rapi_json(JsonDocument &doc)
   doc["amp"] = evse.getAmps() * AMPS_SCALE_FACTOR;
   doc["voltage"] = evse.getVoltage() * VOLTS_SCALE_FACTOR;
   doc["pilot"] = evse.getChargeCurrent();
+  doc["max_current"] = evse.getMaxCurrent();
```
jeremypoulter, line 97: This can be retrieved via the `/config` endpoint. Shouldn't be updated that often, most use cases should set the `max_current` via a claim.

src/mqtt.cpp
```diff
@@ -99,14 +105,51 @@ void mqttmsg_callback(MongooseString topic, MongooseString payload) {
...
       divertmode_update(newdivert);
     }
   }
+  // Config: main Max current
+  else if (topic_string == mqtt_topic + "/max_current/set")
+  {
+    int newmaxcurrent = payload_str.toInt();
+    DBUGF("Set max_current: %d", newmode);
+    evse.setMaxConfiguredCurrent(newmaxcurrent);
+  }
```
jeremypoulter, line 122: This can be removed, max current can be set via a claim

src/input.cpp
KipK: ok. I'll add the /config endpoint to mqtt to later. Do you think we should expose complete settable config too ? I'm wondering about security vs functionality

KipK:
To Do for later: 

1/ rewrite mqtt.cpp/.h as a class.
2/ publish everything has "retained",  and push only updated data as soon as they have changed. ( need point 3 ).  Current implementation spam smart home solution database uselessly.
3/ remove the 30sec loop, we need a better event driven status system to publish data when a value has changed. some values like "pilot" are not evented taking too long to update on mqtt side. This will also allows to remove the status polling on the web ui. . I have no idea of the change involved, I'd rather let you guys taking care of this one. I'm still blurry with the main codebase.
4/ publish /config topic ( read only or settable? )

src/input.cpp
jeremypoulter: That isn't what I meant, was just meaning this value can be retrieved by that HTTP endpoint and doesn't really fit in MQTT. Config updates are events via the `config_version` topic, when this changes the config can then be read via HTTP.

KipK: Yes I got it later. I've removed it. Can you give a last check to see if it's readty to merge? I will prolly be busy next week.

jeremypoulter:
> 2/ publish everything has "retained", and push only updated data as soon as they have changed. ( need point 3 ). Current implementation spam smart home solution database uselessly.

Possibly this could be an option, may need to be optional though, maybe more relevant for the values that are not evented every 30 seconds, but mostly they where targeted at the HTTP/WS API, IE you get the initial state via the `/status` endpoint then updates are sent via the WebSocket. 

> 3/ remove the 30sec loop, we need a better event driven status system to publish data when a value has changed. some values like "pilot" are not evented taking too long to update on MQTT side. This will also allows to remove the status polling on the web ui. . I have no idea of the change involved, I'd rather let you guys taking care of this one. I'm still blurry with the main codebase.

The 30 second loop is needed as the values are not evented from the EVSE module, guess we could keep a record of the last value and only send when they change. This may need to also be optional as so use cases it is good to get the regular updates, ie to make sure the device is still online.

> This will also allows to remove the status polling on the web ui.

This could be removed already, there may a few bugs, where values may not be evened, but sure these can be resolved. In reality this will probably just not be done when doing a new UI.

I would raise a separate ticket for all of these.

jeremypoulter (commented):
Code changes look good, will give it a test then can be merged.

KipK:
`The 30 second loop is needed as the values are not evented from the EVSE module, guess we could keep a record of the last value and only send when they change. This may need to also be optional as so use cases it is good to get the regular updates, ie to make sure the device is still online.`

On MQTT we already have the last will message to know if the device is connected or not so that shouldn't be a problem. Also we could add an heartbeat somewhere.

KipK:
Can you first merge the current_shaper pull request before this one  ? 
I willthen add the mqtt retain setting. Can't merge it now, I have too much conflicts from my branch with booth :)

KipK:
Do you plan to merge those soon ? 
Maintaining 2 branches for those pull requests and a third one with booth, for my usage, is getting somehow confusing. :)

jeremypoulter:
@KipK really sorry, yes, definitely intended on reviewing/merging this as soon as #350 is complete but I have to focus on that as it is now blocking sales in the UK

src/mqtt.cpp
```diff
@@ -294,9 +413,36 @@ mqtt_loop() {
     }
   }
 
+  if((millis() - mqttStatusLoopTime > MQTT_STATUS_LOOP_TIME) || !mqttStatusLoopTime ) {
+    mqttStatusLoopTime = millis(); // reset
+    mqtt_pub_status();
+  }
```
jeremypoulter, line 419: These values are evented as fast as they are read from the EVSE module so I don't think running a faster loop here will get you anything.

If you need faster updates, maybe the polling time can just be made configurable? but I think that should be a separate PR

KipK: It does. I can see fluctating values each 2 sec for live Amp , and pilot get refreshed in time. If not doing that it can takes 30 sec. 
You can see I only push few essential values in this loop to not spam the mqtt server. The others are refreshed by the main loop. I would prefer those to be pushed directly by the EVSE module when value has changed, but for now I haven't found a better way to do without modifications on the EVSE firmware.

I can make the poilling configurable in the UI

KipK: Seems I have a better way of doing this, instead of polling, can I for example, add a send_event in the EvseMonitor::setPilot, just after we receive the RAPI_RESPONSE_OK ?

KipK:
this [0b3c212](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/pull/379/commits/0b3c212f0e7c6be7b86cea1f2a7bafc52dfda8ca) works even better than the polling.

src/mqtt.cpp
jeremypoulter: Hum... Oh right yeah some of those values are read more frequently... Yeah I guess we can do that. Can always make it configurable  if people complain

jeremypoulter (commented):
Gave this a test, looks good, just one open query

jeremypoulter (approved):
Think this is good to go, would at least be good to get it out there for wider comments. Thank you very much for your contribution, it is really appreciated

KipK:
You're welcome. I'm laying around. There's also a pull request for the gui related to this one.

PR #382, opened 2022-06-23
Current shaper module

### LINKED ISSUE #65: Grid current limiting in DIVERT_MODE_NORMAL
At least in the Netherlands, but probably also other places, it is fairly common to have grid connections with rather low current limits on the main breakers, e.g. single phase 35A or 40A or 3 phase 25A for the whole household. If there are other significant loads present, this means that the EVSE would have to be set to a low pilot current in order to avoid tripping the main breaker for the house.
Some commercial EVSEs support a "load balancing" mode, where they monitor the grid current and dynamically adjust the EVSE current to not exceed the limit. I would like to suggest adding this feature to OpenEVSE as well.

This is quite similar to the existing eco divert mode and the changes implemented in #54, but there are a few differences:
* This would always be active, also (especially) when not in eco mode.
* This would be based on the grid I/E information, but instead of aiming for 0 import, you want to keep the imported grid current below a configurable limit.
* This limit needs stricter enforcement: If you exceed it for too long (on the order of seconds, depending on how much it is exceeded?), the house is left without power.
* Because of the previous point and related to #59, I would suggest that the EVSE enters a "safe" state when no message has been received on the grid I/E topic for a certain amount of time. The safe state could be a low current limit or completely disabling charging until information becomes available again.
* The maximum current allocated to the EVSE should still not exceed the configured value. E.g. house has 40A main fuse, EVSE is on a 32A circuit. Then we should never go over 32A pilot current, even when more is available from the grid.

KipK:
Current Shaper Class WIP 

configuration:
mqtt_live_pwr : mqtt topic for instantaneous power taken from the grid 
current_shaper_max_pwr : max power you can get from your energy plan in Watts

Fixes #65


jeremypoulter:
Does the max power need to be a feed? Would it not be a fixed value?

KipK:
> Does the max power need to be a feed? Would it not be a fixed value?

not really for my use case. 
I've kept it dynamic just in case some countries has different energy plan depending of season, time and co.
This info is already provided by my counter and I use it on other scripts so I use the mqtt feed to nbot have to change the value at different places.

src/evse_man.h
```diff
@@ -44,9 +45,10 @@ typedef uint32_t EvseClient;
 #define EvseManager_Priority_Boost     200
 #define EvseManager_Priority_API       500
 #define EvseManager_Priority_Ohm       500
+#define EvseManager_Priority_Shaper    600
```
jeremypoulter, line 48: I don't think you need an explicit priority, you should just use `EvseManager_Priority_Limit`, I would view this as a safety feature so should override everything. Although I guess there may be some cases where you want to allow manual override so maybe a config option to allow switching between this priority and limit depending on the specific usecase.

KipK: Totally agree on that.

KipK:
latest commit seems functional. I've tested it for few hours, from another branch merged with my mqtt mods.
There's no config yet from the web ui, but works with the api.

src/evse_man.h
KipK: > > Does the max power need to be a feed? Would it not be a fixed value?
> 
> not really for my use case. I've kept it dynamic just in case some countries has different energy plan depending of season, time and co. This info is already provided by my counter and I use it on other scripts so I use the mqtt feed to not have to change the value at different places.

I've falled back to a simple variable, the Mongoose Mqtt lib don't have any method to manually read a value from a topic . The max_power topic is only updated rarely if not never ( it's a retained value ),  can't  count on the mqtt callback.

jeremypoulter: Interesting... I would expect you to get called back when you subscribe to the topic...

KipK: not 100% sure, I'll have to retry later, I've had some bugs laying I've solved since.

KipK:
I think we have everything needed but the web UI parts now.

jeremypoulter (commented):
Code changes look good, will give it a test

KipK:
I've added the UI part. There's a pull request on the gui repo too.
Probably something to change in config_save_current_shaper() in app_config.cpp, line 422.
I had to force config_changed(), haven't understood how it should be done properly.

KipK:
I don't get why it says it has conflicts with the static files here. 
Does it need the gui branch to be merged first on your side?

KipK:
> updates to the GUI so you need to merge `master` on both the GUI a

missed those, thanks.

src/app_config.cpp
```diff
@@ -402,6 +419,25 @@ config_save_ohm(bool enable, String qohm)
...
+  config.set("current_shaper_max_pwr", maxpwr);
+  config.commit();
+
+  // please fix this: I'm pretty sure it's not the way to do here, but I haven't got clear the subscription thing probably linked to this :)
+  // so forcing config_cha nged here:
+  config_changed("mqtt_live_pwr");
+  config_changed("current_shaper_max_pwr");
+
+}
+
```
jeremypoulter, line 440: Should not create explicit functions, should just use the generic save

src/web_server.cpp
```diff
@@ -393,6 +396,48 @@ handleDivertMode(MongooseHttpServerRequest *request){
...
+  String qshaperlivepwr =  request->getParam("livepwr");
+
+  config_save_current_shaper(qshaperenabled, qshapermaxpwr, qshaperlivepwr);
+
+  response->setCode(200);
+  response->print("saved");
+  request->send(response);
+}
+
+
```
jeremypoulter, line 440: Should be no need foe explicit endpoints, the generic config endpoint can be used

src/app_config.cpp
KipK: Do you have any example somewhere ?

src/web_server.cpp
KipK: I've made it as a Service in the web UI, so i've tried to match what has been done with others. They don't seem to use /config endpoints there
sshot

KipK: sorry, I just got what you've asked. :)
I'm trying something.

KipK:
should be ok now

KipK:
you need first to get the gui pull request https://github.com/OpenEVSE/openevse_wifi_gui/pull/79 to make it compile on the CI/CD

src/app_config.cpp
jeremypoulter: All these save functions are lagacy and shouldn't be used for new stuff if you look at the JS they are mostly using the /config endpoint. There is a ticket to remove all of these. Sorry for the confusion.

KipK: yep I saw that later, it's fixed now.

jeremypoulter:
Hiya, can you merge master into this and the GUI? I will fix the GUI ref after merging this PR

KipK:
> 
Seems we're good

jeremypoulter:
Yeah, was about to say something has gone very wrong ;-)

src/event_log.cpp
```diff
@@ -168,8 +169,9 @@ void EventLog::enumerate(uint32_t index, std::function<void(String time, EventTy
         double temperature = json["tp"];
         double temperatureMax = json["tm"];
         uint8_t divertMode = json["dm"];
+        uint8_t shaper = json["sh"];
```
jeremypoulter, line 172: Have you tested for backward compatibility with logs that do not have the shaper info?

src/evse_man.h
```diff
@@ -48,7 +49,7 @@ typedef uint32_t EvseClient;
 #define EvseManager_Priority_Ohm       500
 #define EvseManager_Priority_Manual   1000
 #define EvseManager_Priority_RFID     1030
-#define EvseManager_Priority_Ocpp     1050
+#define EvseManager_Priority_Ocpp     1050
```
jeremypoulter, line 52: No need to add trailing spaces

src/current_shaper.cpp
```diff
@@ -0,0 +1,127 @@
...
+				if (_chg_cur < evse.getMinCurrent() ) {
+					// pause temporary, not enough amps available
+					props.setState(EvseState::Disabled);
+				}
+				else {
+					props.setState(EvseState::None);
+				}
+				_changed = false;
+				_timer = millis();
+				evse.claim(EvseClient_OpenEVSE_Shaper,EvseManager_Priority_Limit, props);
```
jeremypoulter, line 34: I think maybe we should be using a new priority for safety features at say `5000`, Limit is still a user choice and we will be using it in the future for the limiting features (like time and energy limits) so don't want those to be able to override the current shaper

```suggestion
				evse.claim(EvseClient_OpenEVSE_Shaper,EvseManager_Priority_Safety, props);
```

src/evse_monitor.cpp
```diff
@@ -507,6 +507,7 @@ void EvseMonitor::setVoltage(double volts, std::function<void(int ret)> callback
         StaticJsonDocument<128> event;
         event["voltage"] = _voltage;
         event_send(event);
+
```
jeremypoulter, line 510: random new line, please remove

jeremypoulter (commented):
I don't think we need additional APIs for enabling/disabling the shaper, this can entirely be done via the config. To my mind it is a safety feature so should be setup and enabled then left. I don't think it should have the same MQTT/HTTP APIs to 'easily' switch on/off from external sources that the divert mode does, but other than that I think this is looking really good, will do some testing , hopefully tomorrow.

KipK:
Looks good now

> I don't think we need additional APIs for enabling/disabling the shaper, this can entirely be done via the config. To my mind it is a safety feature so should be setup and enabled then left. I don't think it should have the same MQTT/HTTP APIs to 'easily' switch on/off from external sources that the divert mode does, but other than that I think this is looking really good, will do some testing , hopefully tomorrow.

I have some use case where it's needed. Sometime I override the shaper to let the car having max priority on my house load. Then the house heaters are shaping the current instead.

Therefore, I've made a quick live switch to override it without writing in the config.

src/event_log.cpp
KipK: > I don't think we need additional APIs for enabling/disabling the shaper, this can entirely be done via the config. To my mind it is a safety feature so should be setup and enabled then left. I don't think it should have the same MQTT/HTTP APIs to 'easily' switch on/off from external sources that the divert mode does, but other than that I think this is looking really good, will do some testing , hopefully tomorrow.

I have some use case where it's needed. Sometime I override the shaper to let the car having max priority on my house load. Then the house heaters are shaping the current instead.

Therefore, I've made a quick live switch to override it without writing in the config. 



> Have you tested for backward compatibility with logs that do not have the shaper info?

not tested that.

jeremypoulter:
Ok fair enough.

src/current_shaper.cpp
KipK: > I think maybe we should be using a new priority for safety features at say `5000`, Limit is still a user choice and we will be using it in the future for the limiting features (like time and energy limits) so don't want those to be able to override the current shaper

Yeah good point, I've changed that.

jeremypoulter:
I have tested this and it looks to work well. I did find a bug in the EVSE Manager when the current sharper was the only claim, and have pushed a fix.

jeremypoulter (approved):

PR #391, opened 2022-07-18
Apply a random start delay to the timer

### LINKED ISSUE #350: Default time schedule and random delay UK legislation

jeremypoulter:
Fix is a UK requirement to help prevent large loads being applied to the grid.

Fixes #350


glynhudson:
Nice work! I've just had a quick test, and it seems to work great. 

The only issue I've noticed is the 'next event' time requires a page refresh to update after the delay timer is updated. Would it be possible to update the 'next event' time without a page refresh? 

I think the Current Event / Next Event text should only be shown when advanced display is togged on. 

Is it possible to switch off the random delay? 

What is the procedure for setting a default schedule before units are shipped?

glynhudson:
> The only issue I've noticed is the 'next event' time requires a page refresh to update after the delay timer is updated. Would it be possible to update the 'next event' time without a page refresh?

This is now fixed 👍 

>I think the Current Event / Next Event text should only be shown when advanced display is togged on.

Confirmed fixed 👍 

>In advanced mode you can change the max time, just set to 0

Got it, I'll add a note to say this. 

> TBD, still looking at that, worst case you can use the HTTP API to set

Sure, that would work. Or we could just set it manually, it would only take a few seconds. 

I think this is ready to merge?

PR #406, opened 2022-08-02
added /scheduler /scheduler/set {json} to MQTT.

KipK:
/scheduler : get scheduler data ( same json as api )
/scheduler/set {json} : set scheduler data
/schedule/clear <ID> : clear related timer

Also fix failed retain switch since virtualBool switch


KipK:
There was a bug since the virtualBool switch, causing retained flag not working. I have added the fix here..

src/app_config.cpp
```diff
@@ -334,19 +333,22 @@ void config_save_emoncms(bool enable, String server, String node, String apikey,
...
+  uint32_t newflags = flags & ~(CONFIG_SERVICE_MQTT | CONFIG_MQTT_PROTOCOL | CONFIG_MQTT_ALLOW_ANY_CERT | CONFIG_MQTT_RETAINED);
   if(enable) {
     newflags |= CONFIG_SERVICE_MQTT;
   }
   if(!reject_unauthorized) {
     newflags |= CONFIG_MQTT_ALLOW_ANY_CERT;
   }
+
+  if (retained) {
+    newflags |= CONFIG_MQTT_ALLOW_ANY_CERT;
```
jeremypoulter, line 345:
```suggestion
    newflags |= CONFIG_MQTT_RETAINED;
```

models/Config.yaml
```diff
@@ -40,7 +40,6 @@ x-examples:
     mqtt_server: homeassistant.lan
     mqtt_port: 1883
     mqtt_topic: openevse
-    mqtt_retained: false
```
jeremypoulter, line 43: This is still valid because of the `new ConfigOptVirtualBool(flagsOpt, CONFIG_MQTT_RETAINED, 0, "mqtt_retained", "mrt"),`

src/web_static/web_server_static_files.h
```diff
@@ -10,8 +10,6 @@
 #include "web_server.lib.js.gz.h"
 #include "web_server.mqtt.png.h"
 #include "web_server.ohm.jpg.h"
-#include "web_server.shaper.png.h"
```
jeremypoulter, line 13: Don't think these are needed

jeremypoulter (commented):
Looks good, I think the `gui` is not required in this PR, ideally the `mqtt_retained` changes should be a separate PR as well, or can you add some more details on the reason for those changes to the PR description?

KipK:
I will do separated PR

PR #411, opened 2022-08-09
Fix for broken EmonCMS updates

### LINKED ISSUE #404: Error from emoncms
I'm running the latest development release. The "Energy Monitoring" area in OpenEVSE says it's not connected.   I have an error coming back from emoncms when data is sent to it.   Here is what I see in the OpenEVSE serial console debug:

`{"amp":0,"voltage":244.9400024,"pilot":6,"wh":2017162,"session_energy":0.258333333,"total_energy":2017.162,"temp":300,"temp_max":300,"temp1":false,"temp2":300,"temp3":false,"temp4":495,"state":254,"flags":1280,"vehicle":1,"colour":6,"manual_override":0,"override":"{\"state\":\"null\"}","freeram":163232,"divertmode":1,"srssi":-71}
{"emoncms_connected":0,"emoncms_message":"Input in not a valid JSON object"}`

I would expect the error is caused by this: `"override":"{\"state\":\"null\"}"`

jeremypoulter:
The new MQTT updates sent the override JSON in the regular updates to EmonCMS and EmonCMS can not handle this soremoved. I think this is not needed as it is handled directly in the MQTT module now.

Fixes #404

### DIFF
--- a/src/input.cpp
+++ b/src/input.cpp
@@ -69,7 +69,6 @@ class InputTask : public MicroTasks::Task
         event["colour"] = evse.getStateColour();
         event["pilot"] = evse.getPilotState();
         event["manual_override"] = manual.isActive() ? 1 : 0; //TODO: remove this
-        event["override"] = getOverride(); 
         event["session_energy"] = evse.getSessionEnergy();
         event_send(event);
       }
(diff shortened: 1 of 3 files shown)


jeremypoulter:
@KipK Can you check this, I don't think that code was needed, but if it is I can add a filter on the data we send to EmonCMS

KipK:
> @KipK Can you check this, I don't think that code was needed, but if it is I can add a filter on the data we send to EmonCMS

I have to check, but I think it's there in case manual override is activated from another channel than mqtt. To get the feedback.
It could be changed in override function to push it when an override is called then.

But can't emoncms ignore unneeded topics ( or better follow new api  ? As you will also have this problm for /claims /schedule and all other fields we've added.

jeremypoulter:
That was my original thinking, but this seemed like the easier fix... You are right about the manual override being triggered from different sources as well as the claims and schedule. It is probably better to drive the MQTT update from those modules directly so the MQTT topics are updated when the data changes.

KipK:
> It is probably better to drive the MQTT update from those modules directly so the MQTT topics are updated when the data changes.

I can do that. Unfortunately, this will prevent us to profit of those new events in a new UI then.

jeremypoulter:
For the UI I have been adding version numbers of the different resources to the status. When these change the UI goes and fetches the appropriate response. Seemed like the best way to do it without exposing all the resources via the status, really need to change the Websocket protocol to be something a bit more flexible

jeremypoulter:
But I can look at the updates as part of this ticket.

KipK:
Just to be clear, I add the getoverride() to manual.cpp and add an mqtt push ( also in manual.cpp ) of this event only when override change ( no evevnt_send() ? )

jeremypoulter:
@glynhudson let me look at the few cases @KipK has mentioned

KipK:
I've made a new PR for this without loosing override change from different channels.
https://github.com/OpenEVSE/ESP32_WiFi_V4.x/pull/416

jeremypoulter:
Superseded by #416

PR #412, opened 2022-08-09
fix retain flag faulty after VirtualBool switch.

KipK:


KipK:
seems something is still missing. I will have to find some time to check.

KipK:
ok I've missed a little stuff, but got a bug with the scheduler in openevse.

Seems If I set only one schedule ( a disabled state only for ex ) , then I see the 30 sec input.cpp loop not publishing anymore.

jeremypoulter (approved):

PR #415, opened 2022-08-11
Issue #187 - Feed watchdog in http_update_write()

### LINKED ISSUE #187: HTTP update failes on ESP32 Etherent gateway
Testing trying to update from 3.3.2 to 4.0.1 

Selection_436
Selection_435

ianrrees:
I've only spent a couple minutes testing this on an ESP32-PoE-ISO board, but it seems to work, and it compiles OK for the targets on the releases page.

Fixes #187

### DIFF
--- a/src/http_update.cpp
+++ b/src/http_update.cpp
@@ -88,6 +88,11 @@ bool http_update_start(String source, size_t total)
 
 bool http_update_write(uint8_t *data, size_t len)
 {
+  // Issue #187 "HTTP update fails on ESP32 ethernet gateway" was caused by the
+  // loop watchdog timing out, presumably because updating wasn't bottlenecked
+  // by network and execution stayed in the Mongoose poll() method.
+  feedLoopWDT();
+
   DBUGF("Update Writing %u, %u", update_position, len);
   size_t written = Update.write(data, len);
   DBUGVAR(written);


jeremypoulter:
Thanks for the contribution, I will check it out

ianrrees:
Thank you for identifying the problem!  I'm hoping to look at upgrading Mongoose this weekend.

There is something strange going on with the relationship between WiFi and Ethernet that I don't fully understand yet.  I think maybe the issue is if the WiFi is connected, and ethernet too, sometimes things get in to a bad state - I've just been erasing the whole flash to nuke the saved WiFi info.  Probably not worth digging too deep before a big networking change anyway.

ianrrees:
Also, just to be clear, I don't think this really fixes the underlying problem - it's just a hacky fix for that one symptom.  Really, the networking stuff shouldn't block the main loop for so long.  Haven't had a chance to look at the current version of Mongoose yet, but perhaps it provides a nice way to resolve that issue.  Or, I believe the ESP32 Arduino framework is actually running FreeRTOS under the hood, it might be more effective to have the OS handle the lower-level networking stuff.

jeremypoulter:
Looks to work well.

jeremypoulter (approved):

PR #447, opened 2022-10-16
fix  #446 remove force auto_release = true for override

### LINKED ISSUE #446: Can't set auto_release parameter with /override api endpoint
Override has auto_release always true whatever you set over the API.
Works using /claims endpoint btw

KipK:
remove force auto_release = true for override.
add default auto_release to true for EvsProperties

fix https://github.com/OpenEVSE/ESP32_WiFi_V4.x/issues/446

### DIFF
--- a/src/evse_man.h
+++ b/src/evse_man.h
@@ -70,7 +70,7 @@ class EvseProperties : virtual public JsonSerialize<512>
     uint32_t _max_current;
     uint32_t _energy_limit;
     uint32_t _time_limit;
-    bool _auto_release;
+    bool _auto_release = true;
   public:
     EvseProperties();
     EvseProperties(EvseState state);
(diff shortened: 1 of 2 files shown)


jeremypoulter (commented):
This is part of the design of the manual override. It is supposed to auto clear at the end of the session as it is only supposed to be used to override the current session. 

What is the problem you are trying to solve and why can't the /claim API be used?

jeremypoulter:
I also think setting the default to true could break a lot of things, the default was true and I think most modules make use of this.

KipK:
> This is part of the design of the manual override. It is supposed to auto clear at the end of the session as it is only supposed to be used to override the current session.
> 
> What is the problem you are trying to solve and why can't the /claim API be used?


I thought it was a feature has it publish auto_release key with /override but I haven't checked the POST part of the manual that hasn't. So it's intentional, thanks.

I totally can use Claims for the Activate/Auto/Disable button of the new UI, and that's what I already use without issue with my MQTT widgets so far.

PR #449, opened 2022-10-18
update /override mqtt topic when external manual override is added

KipK:


KipK:
Do not commit yet, I have to rethink it, there's still some case where it's not updated when change comes from outside source than mqtt

KipK:
It's better like this. Works in all tested case now.

PR #454, opened 2022-10-24
Few fixes

KipK:
some minor changes.
- Doc updates
- Updated `shaper` type to match other modules
- a few fixes found while testing OpenEVSE/openevse_wifi_gui#87

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -151,14 +151,14 @@ Data can be posted using HTTP or HTTPS.
 
 OpenEVSE can shape charge current according to your real time house load, preventing to exceed the maximum power your energy plan can handle.
 Once the module is toggled on, it will have highest priority to other claims.
-However it's possible to temporary disable it if needed with HTTP or MQTT
+However it's possible to temporary disable it if needed with HTTP or MQTT. For HTTP, send a text/plain HTTP POST request to /shaper with body containing shaper=value
 
 **Note #1**: this service is dependant of an external feed containing the  household live power in watt.
 It can come from an MQTT topic or an HTTP POST request.
  *** MQTT ***
  Set the topic in the shaper configuration page ( 'Live power load MQTT Topic' )
  *** HTTP ***
- Send periodically an HTTP POST request to http://{openevse_host}/status containing JSON formatted as /status GET endpoint.
+ Send periodically an HTTP POST request to http://{openevse_host}/status containing TEXT formatted like in /status GET endpoint.
 Body should contain {"shaper_live_pwr": value} in watt. Can be combined with other settable values.
   
 **Note #2**: set 'Max Power Allowed' according to your energy plan.
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -4,25 +4,32 @@
 
 <!-- toc -->
 
-* [Hardware](#hardware)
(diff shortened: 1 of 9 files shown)


docs/user-guide.md
```diff
@@ -88,15 +96,20 @@ A [OpenEnergyMonitor Solar PV Energy Monitor](https://guide.openenergymonitor.or
...
 ### Eco Mode Setup
 
+#### using MQTT
 * Enable MQTT Service
 * [emonPi MQTT credentials](https://guide.openenergymonitor.org/technical/credentials/#mqtt) should be pre-populated
 * Enter solar PV generation or Grid (+I/-E) MQTT topic e.g. Assuming [standard emonPi Solar PV setup](https://guide.openenergymonitor.org/applications/solar-pv/), the default MQTT feeds are:
   * Grid Import (positive Import / Negative export*): `emon/emonpi/power1`
   * Solar PV generation (always postive): `emon/emonpi/power2`
 
+#### using HTTP
```
jeremypoulter, line 106: Can you also update `api.yml` ?

src/web_server.cpp
```diff
@@ -709,7 +709,7 @@ handleScheduleGet(MongooseHttpServerRequest *request, MongooseHttpServerResponse
     response->setCode(200);
     serializeJson(doc, *response);
   } else {
-    response->setCode(404);
+    response->setCode(200);
```
jeremypoulter, line 712: The `404` is correct in this case, the requested event does not exist and the `404` allows you to detect this

src/web_server.cpp
```diff
@@ -811,7 +811,7 @@ void handleOverrideGet(MongooseHttpServerRequest *request, MongooseHttpServerRes
     manual.getProperties(props);
     props.serialize(response);
   } else {
-    response->setCode(404);
+    response->setCode(200);
```
jeremypoulter, line 814: Same as above, the `404` is correct

src/web_server_claims.cpp
```diff
@@ -29,7 +29,7 @@ handleEvseClaimsGet(MongooseHttpServerRequest *request, MongooseHttpServerRespon
     response->setCode(200);
     serializeJson(doc, *response);
   } else {
-    response->setCode(404);
+    response->setCode(200);
```
jeremypoulter, line 32: Same as above the `404` is correct

jeremypoulter:
`/claims` returns 200 with an empty array when no claims are set. We can also add a `claim_version` to the status, same as the schedule if you need an event to update the claims

src/web_server.cpp
KipK: I think the REST API should only answer 404 for not found element like /claims/{id}  /schedule/{id}
But not for getting claims/ or /override as it is still a real answer with needed info ( array is empty )

KipK:
> `/claims` returns 200 with an empty array when no claims are set. We can also add a `claim_version` to the status, same as the schedule if you need an event to update the claims, but the `404` is expected and the correct response in those cases.

ho I see, I'm calling /claims/{id} from the UI to only get the claim from Manual clientid.
I can solve that by grabing the whole package.

KipK:
a claim_revision in status could be usefull yes

jeremypoulter (approved):

PR #471, opened 2022-11-09
add upload_progress event to websocket

KipK:
This send "upload_progress" value to websocket while uploading
It gives better upload file progress bar as it seems to be broken using classic progress event from xmlhttprequest ( also can't work correctly behind proxy ) 

I'm using it on wip UI , demo at the end of video here https://watch.screencastify.com/v/EP73NnuERNFHosoQAR92

### DIFF
--- a/src/http_update.cpp
+++ b/src/http_update.cpp
@@ -79,7 +79,9 @@ bool http_update_start(String source, size_t total)
 
     lcd.display(F("Updating WiFi"), 0, 0, 10 * 1000, LCD_CLEAR_LINE | LCD_DISPLAY_NOW);
     lcd.display(F(""), 0, 1, 10 * 1000, LCD_CLEAR_LINE | LCD_DISPLAY_NOW);
-
+    StaticJsonDocument<128> event;
+    event["ota"] = "started";
+    web_server_event(event);
     return true;
   }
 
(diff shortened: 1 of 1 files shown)


jeremypoulter:
Interesting, the current upload progress should not be effected by a proxy, should be based on the progress of the upload from the browser POV, IE how much has been sent. That being said doing it this way will also include the local processing  as well so no problem with it

KipK:
> Interesting, the current upload progress should not be effected by a proxy, should be based on the progress of the upload from the browser POV, IE how much has been sent. That being said doing it this way will also include the local processing as well so no problem with it

Seems in some case, the browser send  big chunks of file quickly to the proxy.
progress value is jumping to 100% in no time on browser side, while proxy is slowingly finishing the task.
This is the case with some node.js proxying like  Vite use.

jeremypoulter:
Ah that makes sense the changes look reasonable, will test and merge.

jeremypoulter (approved):

KipK:
Please do not merge yet, I'll rename it to ota_progress, and also had an "ota" event for "started" "success" "fail" "error" status

KipK:
It's ok now.

jeremypoulter (approved):

PR #475, opened 2022-11-15
add auto_release to /override

KipK:
After many considerations, 'auto_release' missing from /override will be needed for the uses cases we have talked about.
I've tryied to trick by using /claim with manual_override client ID, but priority is then HTTP Api one ( 500 ) .

Should be done the proper way, default to true if there's no auto_release property in the override claim

### DIFF
--- a/src/current_shaper.h
+++ b/src/current_shaper.h
@@ -17,6 +17,7 @@
 #include "http_update.h"
 #include "input.h"
 #include "event.h"
+#include "divert.h"
 
 class CurrentShaperTask: public MicroTasks::Task
 {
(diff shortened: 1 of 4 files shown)


KipK:
I've added the ignored claims fix to this PR, it has sense.

KipK:
> I've added the ignored claims fix to this PR, it has sense.

seems it's not sufficient, I still get the /target without the override props....

KipK:
it stills don't take the override props if there's no state width it. But I don't see where it comes from. I'm a bit lost, if you see a hint somewhere...

KipK:
I've found it, I've effectively introduced this in a previous commit.

KipK:
I'm gonna resubmit in separated pr

PR #485, opened 2022-11-17
improvment & fixes of MQTT /override topic

KipK:
[add auto_release to /override](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/pull/485/commits/793797ed593b8aeef423e327b744838092f087d5)

[fix override publication missed on mqtt](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/pull/485/commits/c3dbf83194365db88b168055b0ab2cfb63c4e184)

[fix mqtt /override topic publishing too much props](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/pull/485/commits/71663dd4c4c7004429c367b8da85ed3884790178)

### DIFF
--- a/src/evse_man.cpp
+++ b/src/evse_man.cpp
@@ -10,6 +10,7 @@
 #include "event_log.h"
 #include "divert.h"
 #include "current_shaper.h"
+#include "manual.h"
 
 static EvseProperties nullProperties;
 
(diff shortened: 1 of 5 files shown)


src/evse_man.cpp
```diff
@@ -445,6 +439,12 @@ unsigned long EvseManager::loop(MicroTasks::WakeReason reason)
   {
     _evaluateTargetState = false;
     setTargetState(_targetProperties);
+
```
jeremypoulter, line 442: I am not super keen on this, will leave it for now but we need to consider how to make this more generic, I would prefer not to have these special cases in the EVSE manager... We probably need to give the clients text names and event them  all or something like that...

PR #487, opened 2022-11-20
Adding building of v2 UI

jeremypoulter:
Can now set the `GUI_NAME` env var to the directory containing the GUI. The script is a lot more flexable now and handles nested resources, as well as a mix of compressed and uncompresssed resources


KipK:
Thanks for this. I'll try soon . 

I think webserver_static.cpp needs a bit of modification to handle the index.html and the /assets subfolder 
We also have some .webmanifest file used for the webapp.
We can't use it as true Progressive Web App yet, as it require the files to be hosted over HTTPS.

KipK:
@jeremypoulter , how I should use this now to build the fw with new UI ?

jeremypoulter:
There are some additional envvars you need to set to build the UI: https://github.com/OpenEVSE/ESP32_WiFi_V4.x/pull/487/files#diff-d0777657fa3fd81d23aaf7273e58aee453b04e67882517900c56daeef9b3e4c1R102-R104

so assuming you have teh v2 GUI source checked out in the `ESP32_WiFi_V4.x/openevse-gui-v2` directory (and built from an NPM/JS POV) you would set the vars for the specific command

```shell
GUI_NAME=openevse-gui-v2 PLATFORMIO_BUILD_FLAGS='-D DISABLE_WIFI_PORTAL -D WEB_SERVER_ROOT_PAGE_INDEX' pio run ...
```

eg if you want to switch from one GUI to another, or

```shell
export GUI_NAME=openevse-gui-v2
export PLATFORMIO_BUILD_FLAGS='-D DISABLE_WIFI_PORTAL -D WEB_SERVER_ROOT_PAGE_INDEX'
pio run ...
```

KipK:
Can I setup those vars in platoformio.ini directly ?

PR #489, opened 2022-11-22
publish override_version after claim engine has treated it.

KipK:


jeremypoulter (commented):
Will do for now, but not really that clean. I think we may need a mechanisam to event the state changes internally, but that will probably be a bit of a larger shake up...

PR #495, opened 2022-12-09
Current Shaper now use "max_current" instead of "charge_current".

KipK:
WARNING! If you were manually changing max_current using HTTP API or MQTT, don't forget to switch to charge_current now


jcsogo:
> WARNING! If you were manually changing max_current using HTTP API or MQTT, don't forget to switch to charge_current now

The text in the comment doesn't match the text in the title and (I think) the code. The change is to deprecate "charge_current" and use "max_current" instead, if I am not mistaken.

Thanks.

KipK:
Nope it's for users that were using max_current before to throttle their charge_current because shaper was using charhe_cirrent with high priority. 
So now they should use charge_current and won't be interfered by Shaper ( using now max_current". )

PR #498, opened 2022-12-12
toggle divertmode between normal/eco when scheduler change state

KipK:
removed charge_current claim when scheduler get active
toggle divertmode between normal/eco when scheduler change state


jeremypoulter:
I would really like to avoid this sort of thing, the EVSE manager is there specifically so different modules do not need to know about each other to make it easier to expand in the future, is this just a UI display issue you are trying to solve or is there another problem?

KipK:
Yeah I'm agree it was a try out to disable the Divert button when Timers are active, but I can do it other ways.

PR #503, opened 2022-12-20
Ui2 build

KipK:


jeremypoulter:
@KipK I think your source branch was behind the `build_v2_gui` branch, can you pull the latest changes? Also please can you remove the V2 static files from `src/web_static` and leave with the original v1 files? would like to leave that as the 'default' UI while we are still testing.

KipK:
I've rebased the branch. Should be good now.

jeremypoulter (approved):

PR #515, opened 2023-01-15
Quick and dirty filter to ensure no >127 chars are sent over the websocket and hense valid UTF-8 is sent

### LINKED ISSUE #509: /evse/console websocket issue
Websocket with evse console close with thrown error:
` Could not decode a text frame as UTF-8. `

Seems /evse/console endpoint should filter its output from those unkown characters

jeremypoulter:
On a quick scan there does not look to be a readily avalible library for ASCII to UTF-8 conversion, so as we don't have all that many cases where this is an issue, basically just the space on the LCD, just a quick hack to ensure valid UTF-8 is sent.

Fixes #509

### DIFF
--- a/src/web_server.cpp
+++ b/src/web_server.cpp
@@ -1202,6 +1202,19 @@ void onWsConnect(MongooseHttpWebSocketConnection *connection)
   web_server_event(doc);
 }
 
+/*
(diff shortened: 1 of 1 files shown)


KipK:
Works fine here thanks

Issue #452, opened 2022-10-24
max_current in claim or override not taken into account if set before starting charge


KipK:
spotted a bug on claim engine, if you set a claim with max_current set to a lower value than max_current_soft , then when upgrading the claim state to active, it will start the charge with full power ignoring the max_current properties in the claim.
Resending a claim after set state: active will then works.

how to reproduce:
- POST a claim ( over http or mqtt) with {state: "disable"}
- plug the car
- POST a claim with only {max_current: 6}
- POST a claim with {state: "active"}
Charge will start at 32A or what your max_current_soft is set in config instead of 6A
- POST a claim with {max_current: 6}
- now it charges OK at 6A

pdhoogh:
This looks similar to #450 ?

KipK:
@jeremypoulter nope it's a new issue.
it the same result with HTTP or MQTT.
Forget the merged point ( for HTTP i send all the attributes ), I've checked the claims before and after suing http api, all the parameters are ok in the claims but it always ignore the max_current settings first.

For info, we use max_current to set a lower charge max for the session, while shaper use charge_current ( and if there's a max_current claim, it will modulate inbetween this new max_current )

Try to set a claim with disable and max_current set to a lower value.
Then set the claim to enable
You can see the

jeremypoulter:
The current shaper should really be using the `max_current` as well so it can interoperate with the solar PV for example. Maybe the `max_current` should behave slightly differently to the rest of the properties and the lowest value should win rather than the highest priority? I couldn't replicate the issue however, there is defiantly some weird interaction between starting charge and the current

KipK:
That could be a good idea yes. I don't know if there could have some collaterals therefore.

Here are the claims when stopped: 
```

   {
        "client": 65540,
        "priority": 100,
        "state": "active",
        "auto_release": false
    },
    {
        "client": 65548,
        "priority": 5000,
        "charge_current": 34,
        "auto_release": false
    },
    {
        "client": 65537,
        "priority": 500,
        "state": "disabled",
        "max_current": 6,
        "auto_release": true
    }
```
Then the claims when activated ( start charging at 32A)
	
```
	[
    {
        "client": 65540,
        "priority": 100,
        "state": "active",
        "auto_release": false
    },
    {
        "client": 65548,
        "priority": 5000,
        "charge_current": 39,
        "auto_release": false
    },
    {
        "client": 65537,
        "priority": 500,
        "state": "active",
[...]

PR #518, opened 2023-01-16
do not set Scheduler claim if evse module communication is not ready

### LINKED ISSUE #517: Timer set charge_current claim with 0 amp
I see Timer is setting a charge_current claim with getMaxHardwareCurrent() if EVSE is active when timer going to active.
But mose of the Time the claim is posted with "0" as max_current , so pilot is locked at 6A )
It looks like    properties.setChargeCurrent(_evse->getMaxHardwareCurrent()); gives 0 for whatever reason;

This is part of code in scheduler.cpp:
    ``` if(EvseState::Active == currentEvent.getState())
      {
        priority = EvseManager_Priority_Timer;
        properties.setChargeCurrent(_evse->getMaxHardwareCurrent());
      }
      _evse->claim(EvseClient_OpenEVSE_Schedule, priority, properties);```

I'm publishing getMaxHardwareCurrent() in /status to follow it, but it's always at 32A, I don't get why this part of code fails on scheduler.cpp

To reproduce:

- Set a timer to have OpenEvse Active at this time, reboot wifi module, at boot it will publish a timer claim with charge_current at 0.
ps: I have also tryied replacing getMaxHardwareCurrent() by getMaxConfiguredCurrent() and getMaxCurrent() and same thing happend

KipK:
fix #517

### DIFF
--- a/src/evse_monitor.cpp
+++ b/src/evse_monitor.cpp
@@ -164,7 +164,7 @@ EvseMonitor::EvseMonitor(OpenEVSEClass &openevse) :
   _min_current(0),
   _pilot(0),
   _max_configured_current(0),
-  _max_hardware_current(0),
+  _max_hardware_current(80),
   _data_ready(EVSE_MONITOR_DATA_READY),
   _boot_ready(EVSE_MONITOR_BOOT_READY),
   _session_complete(EVSE_MONITOR_SESSION_COMPLETE_MASK, EVSE_MONITOR_SESSION_COMPLETE_TRIGGER),


KipK:
I displays the claims/target in a tab in the UI, to keep it clear for users I'd rather have ok values than hardcoded ones displayed.

This issue probably hide some others.  I temporise on scheduler, but it's probably on the evseManager side that things should be handled isn't it ?

KipK:
@jeremypoulter I think we are close to solve many other issues as #452 , seeing my latest comment here https://github.com/OpenEVSE/ESP32_WiFi_V4.x/issues/517#issuecomment-1384523287

If it's the case this PR has no reason te be merged and we should investigate on evse_manager

jeremypoulter:
I think the correct way to solve this would befor the scheduler (and all the other modules) to wait for the evse to be active before sending the claims... but that being said the EVSE managershould also handle the claims at any point. 

IMHO setting to a fixed value is not a problem, 80A is the maximum allowed by the spec so not just a random value, and after startup will get reset. potentially I could hide this in the web layer or EVSE layer. but for this issue specifically I think it is a better fix than polling until the value is read correctly.

I will also check on why 0 is returned under certain conditions, if that is just the default before the correct value is read then again 80 is as good as 0 (well probably better)

jeremypoulter:
Yeah I think it is just a startup timing issue, I think setting the defailt valueto 80 would work: 
https://github.com/OpenEVSE/ESP32_WiFi_V4.x/blob/master/src/evse_monitor.cpp#L164-L167 probably should be

```c++
  _min_current(6),
  _pilot(6),
  _max_configured_current(80),
  _max_hardware_current(80),
```

After the EVSE starts up the claims will be re-evaluated and the value will be clamped (or at least it should be)

KipK:
> I think the correct way to solve this would befor the scheduler (and all the other modules) to wait for the evse to be active before sending the claims... but that being said the EVSE managershould also handle the claims at any point.

Yeah this would imply patching all the services, and deserve evseManager purpose no ? 
I'd rather go for fixing evsemanager evaluateClaim and let it handle that

jeremypoulter (approved):

KipK:
There's still something strange, I can see the claim not being reevaluated sometime. It stays at 80.

PR #530, opened 2023-01-29
Enable Shaper with Divert mode active

KipK:
Now shaper can work while divert is active.
produced energy or grid export is added to shaper max_pwr value
This is mostly necessary when Divert stays at 6amp waiting for smoothed current to get down before shutting down.
If the grid doesn't has enough amp available it can overload. This should prevent it.

and other little commits


KipK:
Do not merge yet, I have something to fix first

KipK:
it seems ok for what I could test with virtual solar stuff.

jeremypoulter (approved):

PR #532, opened 2023-01-29
fix #452 reevaluate claims if amps > pilot

### LINKED ISSUE #388: Divert mode current set initially set to maximum level regardless of generation
(Bug present in release version but also confirmed present on latest dev build)

My system is configured in "SolarPV-gen" mode with the MQTT topic reflecting generation published to every 10 seconds. Attack/decay are at their defaults.

When divert mode is activated and sufficient generation becomes available (>= 6 Amps), ESP32 Wifi will set the pilot to the maximum charge current regardless of actual generation. It will remain at maximum until either that setting is changed or the generation available would cause the pilot to be set to another integer current level.

Both the "Charge rate" (both main and service tabs) and "Pilot" values in the web GUI reflect what should have been sent to EVSE based on available generation but the actual current is still at the maximum level.

### LINKED ISSUE #452: max_current in claim or override not taken into account if set before starting charge

### LINKED ISSUE #481: Controller restart ignored eco divert mode setting at boot
If OpenEVSE is in Eco (PV Divert) mode and the controller is restarted, those settings are ignored (but are successfully restored) and the car starts charging right away:

**Before restart** 

image

**After restart**

image

As you can see, Eco mode is still active after restarting the controller, but the boot logic ignored it was active and there was no energy being exported at all.

### LINKED ISSUE #513: v4.1.7 Charge starts randomly when in eco mode
Since I updated to v4.1.7 (4 days ago), the charger starts charging itself even when it's on eco mode and there is no sun production.
I've started noticing strange behavior since the current shaper was implemented. In previous versions the shaper woke up the car several times a day, depleted the 12v battery on my Tesla M3 so the car went into failsafe/service request.
I saw it live and the effect was that it was waking the car and proceeding with a micro-charge of 2-3 seconds, then stopped. The car stood awaken. So I disabled the current shaper when I noticed it.

I don't know if its related but now it stays charging.
Yesterday:
Screenshot_20230110_091155_Always starts when connected

Today:
Screenshot_20230111_082739_Starts on it's own with no limit

As you can see, the solar excess topic is supplying negative (consumption) with a max rate of 5s, that it is on eco mode and that when it starts like this, it goes up to the charge limit that is set, in my case 28A.
It only stops charging if the wifi module is rebooted.

This is really bad. Today has done it 4 times, depleted my home battery to 20% (limit), and the last time I only noticed because it caused the inverter overload.

KipK:
Until the bug is corrected on evse fw, this solves the issue and prevent the charge_current going to max when Evse goes from Disabled to Active state ignoring claims & pilot values
fix #452 and should also fix #513 fix #388 fix #481

Edit: It seems those bugs are related to PP_AUTO_AMPACITY beeing set. Posted an issue there: https://github.com/lincomatic/open_evse/issues/156?fbclid=IwAR37MVzzv2WZ7iDPPr3lkryBSUIvsDBsbPJkFu5PcZG1JgMT7e2LF6KxFzQ

So this is not really an OpenEvse bug, but more that OpenEvse and Wifi module don't share the non tethered chargers checking cable ampacity state and we have confusions here.

### DIFF
--- a/src/evse_monitor.cpp
+++ b/src/evse_monitor.cpp
@@ -309,7 +309,6 @@ void EvseMonitor::updateEvseState(uint8_t evse_state, uint8_t pilot_state, uint3
       }
       _session_complete.update(getFlags());
     });
-  verifyPilot();
   }
 }
 
--- a/src/evse_monitor.cpp
+++ b/src/evse_monitor.cpp
@@ -163,6 +163,7 @@ EvseMonitor::EvseMonitor(OpenEVSEClass &openevse) :
   _stuck_count(0),
   _min_current(0),
   _pilot(0),
+  _pilot_is_wrong(false),
   _max_configured_current(0),
   _max_hardware_current(80),
   _data_ready(EVSE_MONITOR_DATA_READY),
@@ -308,7 +309,26 @@ void EvseMonitor::updateEvseState(uint8_t evse_state, uint8_t pilot_state, uint3
       }
       _session_complete.update(getFlags());
     });
+  verifyPilot();
   }
+}
+
+void EvseMonitor::verifyPilot() {
+    // After some state changes the OpenEVSE module compiled with PP_AUTO_AMPACITY will reset to the maximum pilot level, so reset to what we expect
+    _openevse.getCurrentCapacity([this](int ret, long min_current, long max_hardware_current, long pilot, long max_configured_current)
+    {
+      DBUGLN("Check pilot is ok");
+      DBUGVAR(pilot);
+      DBUGVAR(getPilot());
+
+      if(RAPI_RESPONSE_OK == ret && pilot != getPilot())
+      {
+        DBUGLN("####  Pilot is wrong set again");
+        _pilot_is_wrong = true;
+        setPilot(getPilot());
+        _pilot_is_wrong = false;
+      }
+    });
 }
 
 void EvseMonitor::updateCurrentSettings(long min_current, long max_hardware_current, long pilot, long max_configured_current)
@@ -319,7 +339,6 @@ void EvseMonitor::updateCurrentSettings(long min_current, long max_hardware_curr
   _pilot = pilot;
   _max_configured_current = max_configured_current;
 }
-
 
 unsigned long EvseMonitor::loop(MicroTasks::WakeReason reason)
 {
@@ -358,6 +377,10 @@ unsigned long EvseMonitor::loop(MicroTasks::WakeReason reason)
   if(0 == _count % EVSE_MONITOR_ENERGY_TIME) {
     getEnergyFromEvse();
   }
+
+  // Check if pilot is wrong ( solve OpenEvse fw compiled with -D PP_AUTO_AMPACITY)
+  if (isCharging())
+    verifyPilot();
 
   _count ++;
 
@@ -464,7 +487,7 @@ void EvseMonitor::setPilot(long amps, std::function<void(int ret)> callback)
     amps = _min_current;
   }
 
-  if(amps == _pilot)
+  if(amps == _pilot && !_pilot_is_wrong)
   {
     if(callback) {
       callback(RAPI_RESPONSE_OK);
--- a/src/evse_monitor.cpp
+++ b/src/evse_monitor.cpp
@@ -317,13 +317,11 @@ void EvseMonitor::verifyPilot() {
     // After some state changes the OpenEVSE module compiled with PP_AUTO_AMPACITY will reset to the maximum pilot level, so reset to what we expect
     _openevse.getCurrentCapacity([this](int ret, long min_current, long max_hardware_current, long pilot, long max_configured_current)
     {
-      DBUGLN("Check pilot is ok");
-      DBUGVAR(pilot);
-      DBUGVAR(getPilot());
-
       if(RAPI_RESPONSE_OK == ret && pilot != getPilot())
       {
         DBUGLN("####  Pilot is wrong set again");
+        DBUGVAR(pilot);
+        DBUGVAR(getPilot());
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -198,6 +199,23 @@ loop() {
...
+          DBUGVAR(current);
+          DBUGVAR(pilot);
+          evse.evaluateClaims();
+          MicroTask.wakeTask(evse);
+          
+        }
+        Timer4 = millis();
+      }
+  
+
```
jeremypoulter, line 218: This really should be hidden away in the EVSE manager/monitor if it is really needed, however getAmps just gets a cached value, so I think ou just need to tag this code into getChargeCurrentAndVoltageFromEvse() which is regularly called

jeremypoulter:
Actually I don't think getAmps is what you should be checking against, that is the live amp reading. What may be an idea is calling openevse.getCurrentCapacity after a state change/current change then doing the check to see if the pilot on the EVSE is the same as what the Monitor thinks it should be, so somewhere here, https://github.com/OpenEVSE/ESP32_WiFi_V4.x/blob/master/src/evse_monitor.cpp#L478, and here, https://github.com/OpenEVSE/ESP32_WiFi_V4.x/blob/master/src/evse_monitor.cpp#L280

KipK:
Do you think then we should reevaluate claim or just set the pilot to evse ?

jeremypoulter:
I think just setting the pilot should be sufficient

KipK:
I have rewrote it on getChargeCurrentAndVoltageFromEvse()

Can't make the change using openevse.getCurrentCapacity , i'm lost with those callbacks.
But it works good as it, ( still the 2-3 sec glitch )

KipK:
Ok I got it thanks. 
I've made the change, gotta test first.

KipK:
Jiust tested but it doesn't work now. Amps stays at 32

KipK:
I have added some debug:

#####   CHECK BUG #452    #####
pilot = 32
getPilot() = 6
####  Pilot differents set again

So the check works ok, but setCurrentCapacity seems to not have effect here.

KipK:
Ok it works now. I've rebased the commits in one.

KipK:
After few tests, I can say it reacts faster than the claim approach I did first. I don't see spikes over the set limit and it stabilise quiet fast.

Until we got a proper integration of auto_ampacity between openevse fw and ESP32 one, this should fit.

src/evse_monitor.cpp
```diff
@@ -359,6 +375,10 @@ unsigned long EvseMonitor::loop(MicroTasks::WakeReason reason)
     getEnergyFromEvse();
   }
 
+  // Check if pilot is wrong ( solve OpenEvse fw compiled with -D PP_AUTO_AMPACITY)
+  if (isCharging())
+    verifyPilot();
```
jeremypoulter, line 380:
```suggestion
  if (isCharging()) {
    verifyPilot();
  }
```

src/evse_monitor.cpp
```diff
@@ -464,7 +484,7 @@ void EvseMonitor::setPilot(long amps, std::function<void(int ret)> callback)
     amps = _min_current;
   }
 
-  if(amps == _pilot)
+  if(amps == _pilot && !_pilot_is_wrong)
```
jeremypoulter, line 487: Think it would be cleaner to add a parameter to the method with a default value, eg `bool force = false`

src/evse_monitor.cpp
```diff
@@ -311,6 +312,22 @@ void EvseMonitor::updateEvseState(uint8_t evse_state, uint8_t pilot_state, uint3
   }
 }
 
+void EvseMonitor::verifyPilot() {
+    // After some state changes the OpenEVSE module compiled with PP_AUTO_AMPACITY will reset to the maximum pilot level, so reset to what we expect
+    _openevse.getCurrentCapacity([this](int ret, long min_current, long max_hardware_current, long pilot, long max_configured_current)
+    {
+      if(RAPI_RESPONSE_OK == ret && pilot != getPilot())
```
jeremypoulter, line 319: May need to just check for larger pilot values to handle case where EVSE has loimited the current to a lower value than set.
```suggestion
      if(RAPI_RESPONSE_OK == ret && pilot > getPilot())
```

src/evse_monitor.cpp
```diff
@@ -311,6 +312,22 @@ void EvseMonitor::updateEvseState(uint8_t evse_state, uint8_t pilot_state, uint3
...
+      {
+        DBUGLN("####  Pilot is wrong set again");
+        DBUGVAR(pilot);
+        DBUGVAR(getPilot());
+        _pilot_is_wrong = true;
+        setPilot(getPilot());
+        _pilot_is_wrong = false;
```
jeremypoulter, line 326: See other comments but should be something like this

```suggestion
        setPilot(getPilot(), true);
```

src/evse_monitor.cpp
KipK: yes probably better. Change made

src/evse_monitor.cpp
KipK: ok I've added the param to setPilot

jeremypoulter (approved):

PR #533, opened 2023-01-29
Re-work of the network manager

### LINKED ISSUE #420: Wifi connection loop: possible improvement.
I have a poor wifi connection (RSSI around -80) and the network connection is unstable. By default, I understand the net loop restarts OpenEVSE in AP mode every 2 reconnect retries, for 5 minutes.

In my case, this ended up in a situation where the station remained in AP mode most of the time. Because I use the divert mode, this was creating issues with no update of the solar output while in AP mode. I often had to go to the garage and use my phone to reconnect the box to my network. A constant pain.

I modified the net loop and recompiled the firmware: after 2 retries, I instead reset wifi and let the box retry STA mode. I understand the AP mode can still be reached by pressing the touch button for a sufficiently long period of time. 

This has been a game changer. Since then, the whole system has been working flawlessly. I do not need to go to the garage anymore. Night and day really. I imagine I'm not the only person in this situation. I could use a wifi repeater, but this uses money and power, and maybe a source of failure. Instead, a small code change does the job.

I can imagine that the OpenEVSE team wants to stick to this AP approach to facilitate the usage of the box when it is used as a mobile charger. But clearly, this destroys the reliability of the system for all people with weak wifi (which may be more numerous than mobile users in practice). 

My suggestion would be to do the "try AP after 2 STA connect retries" only once at boot time, and then shift to a pure STA loop (with a long press or reboot still allowing to shift to AP mode). In mobile usage, when you use the box, you necessarily reboot (plug the box) when you want to use it, so you'll have the AP mode accessible. If you miss the AP mode time window, you can always unplug the box and retry or long press. For static users (most users I assume), this would lead to much more stable network connections.

Here are my naive changes (for a pure static user). If needed I could try to spend more time to code the proposal above and submit a PR, but it could take more time from me than from OpenEVSE experts.

```
if (isClientOnly && !WiFi.isConnected())
   {
     // If we have failed to connect turn on the AP
-    if(client_disconnects > WIFI_CLIENT_DISCONNECTS_BEFORE_AP) {
-      startAP();
-      client_retry = true;
-      client_retry_time = millis() + WIFI_CLIENT_RETRY_TIMEOUT;
+    if (client_disconnects > WIFI_CLIENT_DISCONNECTS_BEFORE_AP)
+    {
+      net_wifi_restart();
     }
   }
 
-  // Remain in AP mode for 5 Minutes before resetting
-  if(isApOnly && 0 == apClients && client_retry && millis() > client_retry_time) {
-    DEBUG.println("client re-try, resetting");
-    net_wifi_restart();
-  }
```

### LINKED ISSUE #460: ESP gateway issues with factory config over serial
The PR which added serial config seems to be causing an issue with ESP gateway https://github.com/OpenEVSE/ESP32_WiFi_V4.x/pull/443. I've tested on Rev G using the new rev G environment. V4.1.5 works fine on the ESP gateway using the new rev G environment. 

Could there be a serial port conflict?

```
OpenEVSE WiFI fe2c
Firmware: 4.1.5
Build date: Oct 27 2022 17:21:13
IDF version: v4.4.1-1-gb8050b365e
Free: 265712
[    76][E][vfs_api.cpp:104] open(): /littlefs/schedule.json does not exist, no permits for creation
[    77][E][esp32-hal-adc.c:180] __analogRead(): Pin 1 is not ADC pin!
AP IP Address: 192.168.4.1
E (10434) task_wdt: Task watchdog got triggered. The following tasks did not reset the watchdog in time:
E (10434) task_wdt:  - loopTask (CPU 1)
E (10434) task_wdt: Tasks currently running:
E (10434) task_wdt: CPU 0: IDLE
E (10434) task_wdt: CPU 1: IDLE
E (10434) task_wdt: Aborting.

abort() was called at PC 0x4012f4cc on core 0


Backtrace:0x400836e9:0x3ffbee1c |<-CORRUPTED
```

jeremypoulter:
This is a rework of the code to manage connecting to the network, WiFi or Wired. This code better manages the state we are in so hopefully there are less issues and is also more agressive at staying connected to the WiFi access point.

Fixes #420, fixes #460


KipK:
Does the wifi module switch now to AP + STA when wifi has been configured and previous mode was AP ?
I still have to finish the Wizard when this is ok

jeremypoulter:
Going to do some work cleaning up the endpoints based on this and I will work on that area then

glynhudson:
Hi @jeremypoulter I've just done some testing and noticed that when a long press on the front push button is used to force the module to create a WiFi AP captive portal is not active. 

Also, the reported RSSI of a network shows a couple of points worse (more negative) when running this PR, maybe this is not significant?

jeremypoulter:
Will take a look

KipK:
I'm testing this now, and making change to UI wizard ( moving to /config endpoint )
When I set the Wifi + Pass while in AP mode, , it correctly connect to the Wifi network while keeping the AP mode.
However, I see in the log it's looping in reconnection until Wifi module is rebooted.
After we reboot the Wifi module, it then goes in Client only and works ok:

`Connecting to SSID: WifiNetwrk
Got Network event ARDUINO_EVENT_WIFI_STA_START
WiFi State AccessPointConnecting
WiFi Mode AP+STA
Set host name to openevse-c6cd
Got Network event ARDUINO_EVENT_WIFI_STA_DISCONNECTED
WiFi State AccessPointConnecting
WiFi Mode AP+STA
WiFi dissconnected: WIFI_DISCONNECT_REASON_NO_AP_FOUND
Got Network event ARDUINO_EVENT_WIFI_STA_STOP
WiFi State AccessPointConnecting
WiFi Mode AP
Connecting to SSID: WifiNetwrk
Got Network event ARDUINO_EVENT_WIFI_STA_START
WiFi State AccessPointConnecting
WiFi Mode AP+STA
Set host name to openevse-c6cd
Got Network event ARDUINO_EVENT_WIFI_STA_DISCONNECTED
WiFi State AccessPointConnecting
WiFi Mode AP+STA
WiFi dissconnected: WIFI_DISCONNECT_REASON_NO_AP_FOUND
Got Network event ARDUINO_EVENT_WIFI_STA_STOP
WiFi State AccessPointConnecting
WiFi Mode AP
Connecting to SSID: WifiNetwrk
Got Network event ARDUINO_EVENT_WIFI_STA_START
WiFi State AccessPointConnecting
WiFi Mode AP+STA
Set host name to openevse-c6cd
Got Network event ARDUINO_EVENT_WIFI_STA_DISCONNECTED
WiFi State AccessPointConnecting
WiFi Mode AP+STA
WiFi dissconnected: WIFI_DISCONNECT_REASON_NO_AP_FOUND
Got Network event ARDUINO_EVENT_WIFI_STA_STOP
WiFi State AccessPointConnecting
WiFi Mode AP
`

KipK:
Other thing, when first scanning for network in AP only mode , client is disconnected from AP.

jeremypoulter:
@glynhudson I think I have resolved that issue, please can you check again

@KipK I think the connection issue you are seeing is expected behaviour. The configured SSID is not found (`WIFI_DISCONNECT_REASON_NO_AP_FOUND`)  so we keep on retrying.

I also can not reproduce the AP scanning issue, looks to work for me.

KipK:
@jeremypoulter the esp is 3 meters from the AP. signal is stroing.
 It just doesn't connect to AP until the ESP leave AP mode. 
At reboot it also  instantaneously connect to the router.  I think ESP32 should switch channel to the router channel when connecting in AP+STA mode. ( My wifi is channel 13 here, ESP seems to always be on lower channels in AP mode ) 

It found it  at first scan  in AP mode then all the other results return the same ( either empty or a network in the lower band channel ).

I've tried to debug that yesterday without success. Also tryied using WiFiManager lib on the same dev ESP and had no such problem.

KipK:
more info on this, first scan takes few seconds to answer and gives full result. Second scan and all followings answer instantaneously ( doesn't seem to scan really ) and always gives the same result thereafter ( can be empty or few networks around the same channel )

KipK:
I have submitted a PR on your branch allowing to restart mDNS at hostname change https://github.com/jeremypoulter/ESP32_WiFi_V4.x/pull/17

KipK:
so has discussed in pv, seems scanning in passive mode solve the scan issue. 

I now see another problem now since I've moved to passive scan, wonder if I'm alone. When in AP mode, wifi not configured yet, when saving the wifi / pass on /config endpoint, I'm kicked out of the AP,  AP seems disabled then and the ESP module don't connect to the WiFi network until I powercycle.

Expected behavior would be keep in AP+STA, dunno why, it was not doing this before I've set the scan to Passive. 

Other thing, when esp is in STA+AP and connected correctly to the wifi router, the IP adress from the /status endpoint is still AP one (192.168.4.1)
But I can reach the wifi module from its home network ip too.

jeremypoulter:
Oh I see, I fixed an issue with the AP not starting at all...

KipK:
pressing the button more than 2sec seems to not throw the AP portal.


> When the push button is held for 10s to force the module to create a WiFi AP this WiFi AP does not have a captive portal. 

@glynhudson , beware, 10sec is for factory resetting the esp not just set in ap mode.

jeremypoulter:
@KipK I think, that is correct, the user has explicitly changed to AP mode, so they would want to know the IP to then connect to. Technically the ESP has two IPs, maybe we could return both, but I think that can be done separately.

KipK:
but other apps/ui/home automation connected to the api will also receive this new ip , but the evse is still connected to the network and works normally.

KipK:
this https://github.com/jeremypoulter/ESP32_WiFi_V4.x/pull/17/commits/f7b10097cc495e31710b8f6348bae3b4f1184add fix @glynhudson captive portal not showing problem

KipK:
I've submitted you the compile fix, sorry for that. 

https://github.com/jeremypoulter/ESP32_WiFi_V4.x/pull/20

jeremypoulter:
Thanks, I knew I should have waited for those builds to finish ;-)

jeremypoulter:
@glynhudson can you test this again?

KipK:
Are we good for mergin it ?
I don't see any other issues than the one already fixed.

glynhudson:
Great, just tested. Working now, thanks a lot :+1:

PR #535, opened 2023-01-30
New Limit engine for Wifi module. 

### LINKED ISSUE #225: Add SoC and range limits
Now SoC integration is implemented it, SoC limit feature can now be implemented. The easiest way is probably a dropdown box under the energy limit with a few presets e.g 60%, 70%, 80%, 90%

KipK:
Limit engine for Wifi module. Replace limits from OpenEvse module.

New /limit API endpoint
Added classes:
Limit
LimitProperties
LimitType class

POST { "type": type, "value": value, "auto_release" : optional bool }
type actually possible are : 'time", "energy", "soc", "range" 

Disable state when time or energy session > defined limit.
Release disable claim when car is disconnected.

time_limit in minutes
session_energy in w/h

Todo: 
 - add configuration and feature on UI

fix #225


KipK:
Seems we are good for review now

KipK:
Mhh not ready to merge, something I've changed crash the fw. Will check tomorrow

models/Limit.yaml
```diff
@@ -0,0 +1,12 @@
+title: EVSE limit
+allOf:
+  - type: object
+    properties:
+      type:
+        type: string
+        description: 'Limit type: time, energy, soc, range'
```
jeremypoulter, line 7:
```suggestion
      type:
        type: string
        enum:
        - time
        - energy
        - soc
        - range
        description: 'Limit type: time, energy, soc, range'
```
Would also be nice to exten the description to cover what each of the different limits are, particularly the units

src/limit.cpp
```diff
@@ -0,0 +1,295 @@
+#if defined(ENABLE_DEBUG) && !defined(ENABLE_DEBUG_LIMIT)
```
jeremypoulter, line 1: Can you replace the TABs with 2 spaces?

src/limit.cpp
```diff
@@ -0,0 +1,295 @@
...
+					DBUGLN("Limit as expired, disable evse");
+					EvseProperties props;
+					props.setState(EvseState::Disabled);
+					props.setAutoRelease(true);
+					_evse->claim(EvseClient_OpenEVSE_Limit, EvseManager_Priority_Limit, props);
+				}
+			}
+		}
+
+		else if ( _has_vehicle && !_evse->isVehicleConnected()) {
```
jeremypoulter, line 196: There is an `onSessionComplete` complete event you should use for this for consistancy

jeremypoulter (changes requested):
Looks good, some great work. Just a few changes and I think this will work well

models/Limit.yaml
KipK: done

src/limit.cpp
KipK: ah good catch, far better.

src/limit.cpp
KipK: done

KipK:
All changes done, seems we are good to go. 
I have to think about the best UX to handle this on GUI now.

I think of a first selector for the Type, then a conditionned input ( time input for time, selector/ or number input for energy, slider for soc, selector/number input for range

KipK:
@jeremypoulter , I've finished to setup the UI2 for those new limits. Ready to push it when this is merged. 
there's a change requested marked but it's outdated i think.

jeremypoulter:
Mid reviewing/testing, hopefully a day or two

KipK:
I've pushed the gui limit branch here, if you want to test with : 

https://github.com/KipK/openevse-gui-v2/tree/Limit

KipK:
I"ve added new settings in the /config , it allows to set a permanent limit. On the UI side I'll place it in the configuration section, to differentiate from  limits from the main charging page.
Limits from the main page has auto_release = true, and can be deleted from the main page.
Those from the config, with auto_release = false have a lock icon instead of a delete one when displayed on the home.
They can only be deleted from the config. I haven't made the config part on UI yet

jeremypoulter (approved):

KipK:
thx UI merged to master.

PR #538, opened 2023-02-04
dynamically set build version using current git tag/shorthash

KipK:
if it's a realease tag, should name it normally with release name ( i.e 4.1.7 ) , if not, it display the current tag
Will stop confusion displaying always the same version when developping


KipK:
@jeremypoulter , it gets ok the correct version value and set the VERSION env var, but how do I get it to read this value only after the pre: extrascript ?

jeremypoulter:
Will have to check on that, but may have to be a different car as we need a version number in there for the official releases and that version number needs to be there before the tag (ideally), see https://github.com/OpenEVSE/ESP32_WiFi_V4.x/blob/master/docs/process.md. I am not opposed to changing the process, but it is critical that there is no rebuild after testing, and ideally I don't want to increase the version number after a testing failure and multiple builds with the same version number would defeat the purpose of the PR

jeremypoulter:
Will have to check on that, but may have to be a different car as we need a version number in there for the official releases and that version number needs to be there before the tag (ideally), see https://github.com/OpenEVSE/ESP32_WiFi_V4.x/blob/master/docs/process.md. I am not opposed to changing the process, but it is critical that there is no rebuild after testing, and ideally I don't want to increase the version number after a testing failure and multiple builds with the same version number would defeat the purpose of the PR.

KipK:
works ok

KipK:
we crossed posted, I've pushed a change.
But on what I have tested, if I use it with an official release tag, it just set 4.1.7 as version number.
If it's untaged commit then it's  the branch or "latest" + hash

jeremypoulter:
Looks to work, please update the [process docs](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/blob/master/docs/process.md) and you will probably need to update the release [validation workflow](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/blob/master/.github/workflows/release_validation.yaml).

Remember it is very important that their is no rebuild after testing, ie the tag has to be created before testing which will then create the release build and the testing will be done on that. What is the expected flow should the testing find an issue?

KipK:
I did the changes, but please check that it doesn't mess up before as I'm no expert in dev op github action

jeremypoulter:
@KipK I have not forgotten about this, I am just trying to figure out how best to work this into the release process

PR #544, opened 2023-02-11
Update to the latest ESP32 core

### LINKED ISSUE #352: LED Brightness slider bug
While the feature might not be used that much. I have noticed that if you slide the brightness to the lowest option, the LCD will be set to max brightness.

jeremypoulter:
As a consiquance removed the analogueWrite (not compatible) lib and moved over to the built in ledc.

Also fixed #352 LED Brightness slider bug.


KipK:
Thanks for this, tested ok here.

PR #550, opened 2023-02-23
Scheduler and A

miguelcrego:
Hi. Thanks for your great job!
It will be great that on scheduler you can select de A power you want on.
And it will be great to can have more than 2 scheduler. Then you can program different scenarios. For example.
Active 1: on Monday, wendnesday and Friday Active at 2:00 with 20 A.
Disable 1:  on Monday, wendnesday and Friday Disable at 6:00
Active 2: on Tuerday, Thusday and Sunday Active at 4:00 with 15 A.
Disable 2:  on Monday, wendnesday and Friday Disable at 8:00
Thanks.


jeremypoulter:
Please raise your suggestion in the issue section

PR #556, opened 2023-03-01
replace old counter from OpenEVSE module with new EnergyMeter class on ESP

### LINKED ISSUE #484: Feature Request : Daily Energy Counter
We have a session and total energy counter in kWh....
I'd really appreciate if we could also have a "daily" counter that resets daily at midnight.

KipK:
Depends new network branch to be migrated first.
now gives :
- session_elapsed
- session_energy 
- total_energy
- total_day
- total_week
- total_month
- total_year
- total_switches ( number of time the relay/contactor has turned on/off )

Old OpenEvse total is imported to new counter when migrating first time, then a flag prevent to import it again if counters are reset.

fix #484


KipK:
> ```c++
> return (int32_t) evse->getTotalEnergy();
> ```

done thanks

matth-x:
Didn't see that you already updated it in the meantime. Yes, looks perfect 👍

KipK:
I've rebased the branch for review not network manager has merged.

KipK:
I still have one last problem with this branch, the input.cpp loop doesn't seems to run as I don't get the create_rapi_json esach 30 sec anymore.
Haven't found yet what I have removed that create this issue

KipK:
> I still have one last problem with this branch, the input.cpp loop doesn't seems to run as I don't get the create_rapi_json esach 30 sec anymore. Haven't found yet what I have removed that create this issue

I have found, it's because I've removed the getEnergyFromEvse(), was called each 2 sec and was triggering:
 _data_ready.ready(EVSE_MONITOR_ENERGY_DATA_READY);

without this input.cpp is not waken up anymore while not charging 
My last commit is more a hack than a solution,  @jeremypoulter what would be the best way of fixing this ? 
The  getEnergyFromEnergyMeter I've added is just triggering _data.ready, I guess we can change this for a better way and remove the EVSE_MONITOR_ENERGY_TIME loop, has it's not needed  by EnergyMeter

jeremypoulter:
I need to check, but I think all you have to do is remove `EVSE_MONITOR_ENERGY_DATA_READY` from the bits that define when the data has all been fetched

KipK:
Thanks, looks clear now. I'm fixing this

KipK:
ok fixed & rebased

platformio.ini
```diff
@@ -45,7 +45,7 @@ lib_ignore = WebSockets ; ArduinoOcpp: don't compile built-in WS library
 extra_scripts = scripts/extra_script.py
 debug_flags =
   -ggdb
-  -D ENABLE_DEBUG
+  #-D ENABLE_DEBUG
```
jeremypoulter, line 48: Should be reverted

platformio.ini
```diff
@@ -77,7 +78,7 @@ build_flags =
   #-D MG_SSL_IF_MBEDTLS_MAX_FRAG_LEN=2048
   #-D ARDUINO_MONGOOSE_DEFAULT_STREAM_BUFFER=2048
   #-D ARDUINO_MONGOOSE_SEND_BUFFER_SIZE=2048
-  #-D ENABLE_DEBUG
+  -D ENABLE_DEBUG
```
jeremypoulter, line 81: Should be reverted

src/energy_meter.cpp
```diff
@@ -0,0 +1,514 @@
+#if defined(ENABLE_DEBUG) && !defined(ENABLE_DEBUG_ENERGY_METER)
```
jeremypoulter, line 1: Please use 2 spaces to indent

src/energy_meter.cpp
```diff
@@ -0,0 +1,514 @@
...
+	}
+	StaticJsonDocument<capacity> doc;
+	_data.serialize(doc);
+	// Keep previous "imported" property
+	if (_data.imported)
+	{
+		doc["imported"] = _data.imported;
+	}
+	String json;
+	if (serializeJson(doc, json))
```
jeremypoulter, line 458: No need to serialise to a srting first, can pass the file directly

src/energy_meter.h
```diff
@@ -0,0 +1,126 @@
+#ifndef _ENERGY_METER_H
```
jeremypoulter, line 1: Please use 2 spaces to indent

src/web_server.cpp
```diff
@@ -229,12 +229,7 @@ void buildStatus(DynamicJsonDocument &doc) {
   doc["evse_connected"] = (int)evse.isConnected();
 
   create_rapi_json(doc);
-
-  doc["status"] = evse.getState().toString();
```
jeremypoulter, line 233: I think this is still needed

src/web_server.cpp
```diff
@@ -229,12 +229,7 @@ void buildStatus(DynamicJsonDocument &doc) {
...
-
-  doc["elapsed"] = evse.getSessionElapsed();
```
jeremypoulter, line 235: Think we will need to keep these for backward compatibiity, at least for a short period

src/web_server.cpp
```diff
@@ -245,7 +240,6 @@ void buildStatus(DynamicJsonDocument &doc) {
   doc["charge_rate"] = divert.getChargeRate();
   doc["divert_update"] = (millis() - divert.getLastUpdate()) / 1000;
   doc["divert_active"] = divert.isActive();
-  doc["divertmode"] = (uint8_t)divert.getMode();
```
jeremypoulter, line 248: Still needed

src/web_server.cpp
```diff
@@ -256,8 +250,6 @@ void buildStatus(DynamicJsonDocument &doc) {
   doc["limit"] = limit.hasLimit();
 
   doc["ota_update"] = (int)Update.isRunning();
-  doc["time"] = String(time);
```
jeremypoulter, line 259: Still needed

src/web_server.cpp
KipK: oups clearly

src/web_server.cpp
KipK: I thought this one was a typo I think as the properties was "session_elapsed" in evse_mon. You're sur it is intended to have twice the same value in 2 different props on 2 different place ?

jeremypoulter: Hum... ok... LEts leave for now and group all the'depricated' ones together. Will discuss a plan for removing all of them over the next few releases but don't just want to remove with no notice.

src/web_server.cpp
KipK: I remember why I've removed thoses, they are already in create_rapi_json()

src/web_server.cpp
KipK: already in create_rapi_json

src/web_server.cpp
KipK: already in create_rapi_json

jeremypoulter (approved):

src/energy_meter.cpp
```diff
@@ -0,0 +1,508 @@
...
+    date.day = (uint8_t)timeinfo->tm_wday;
+    date.month = (uint8_t)timeinfo->tm_mon;
+    date.year = 1900 + (uint16_t)timeinfo->tm_year;
+
+    return date;
+};
+
+void EnergyMeter::increment_switch_counter()
+{
+  if (_switch_state != _monitor->isActive())
```
jeremypoulter, line 500: Just spotted a bug, I think this should be isCharging to reflect switches of the relay, isActive will increent the count in other cases, eg if the car is not plugged it and the state toggeles from active to disabled.

PR #561, opened 2023-03-08
The newer ESP32 IDF can do core dumps in the coredump partition

jeremypoulter:


KipK:
interesting, how do you read it if it boot loop ? 
Does it automatically output the core dump at next boot ( bootloader feature ? ) or does it has to be handled by the fw ?

KipK:
in the case of bootloop, wouldn't it write forever on the partition ?

jeremypoulter:
https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/core_dump.html

KipK:
We should add a new partition scheme too for the upcoming 16mb flash wifi modules.
I did that for now but not sure it's necessary to allocate this much for the app. 

# Name,   Type, SubType, Offset,  Size, Flags
nvs,      data, nvs,     0x009000, 0x005000,
otadata,  data, ota,     0x00e000, 0x002000,
app0,     app,  ota_0,   0x010000, 0x7F0000,
app1,     app,  ota_1,   0x800000, 0x7F0000,
spiffs,   data, spiffs,  0xFF0000, 0x00F000,
coredump, data, coredump,0xFFF000, 0x10000

KipK (approved):

KipK:
do you want this one to be merged ?

PR #562, opened 2023-03-08
Support for updating directly from GitHub

### LINKED ISSUE #259: Support upgrade directly from GitHub
In the UI download the latest release info (<https://api.github.com/repos/OpenEVSE/ESP32_WiFi_V4.x/releases/latest>) and then use the new HTTP client upgrade (#258) to download/upgrade from the release assets.

This will probably need support for upgrading from a Zip file and/or eventing the upgrade state.

### LINKED ISSUE #502: Add HTTP 302 Redirection
Moongoose lib doesn't follows HTTP 302 redirect.
This is needed for HTTP Ota over github links.

jeremypoulter:
Fixes #259 
Fixes #502

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -60,6 +60,8 @@ debug_flags =
   #-D ENABLE_DEBUG_TESLA_CLIENT
   #-D ENABLE_DEBUG_LIMIT
   #-D ENABLE_PROFILE
+  #-D ENABLE_DEBUG_HTTP_UPATE
+  #-D ENABLE_DEBUG_TIME
   #-D ENABLE_NOISY_PROFILE
   #-D ENABLE_DEBUG_MICROTASKS
 src_build_flags =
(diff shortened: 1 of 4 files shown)


KipK:
Great , thanks for this. I'm enabling it on the UI

KipK:
That works great, 

https://watch.screencastify.com/v/qHl5qp0bJliqgrgkNbth

KipK:
Perhaps one catch here. 

http_update_end doesn't seems to be called at the end ( or missing test if(Update.end(true)) ). not sure but the event["ota"] = "completed" is not published so I suspect this is the case

jeremypoulter:
Will take a look

KipK:
@jeremypoulter, forget it it works. It was testing the wrong ui branch

KipK:
for proper integration of having daily builds downloaded from the gui v2, I'll wait for ui v2 migration on next major release.
It will need to have the same file names as release to be consistant. 
We should also add the hashed version number as tag instead of "latest" on github this will feed the tag_name properties in : 
https://api.github.com/repos/OpenEVSE/ESP32_WiFi_V4.x/releases
and can be use to compare versions on the gui.

PR #564, opened 2023-03-10
ping pong over websocket

KipK:
let wifi module answer to ping over websocket
Handled on UIv2 (pingpong branch ) for proper disconnection/reconnection
( client send a ping if it hasn't received data for 5sec ) , tries 3 times each sec if it doesn't get an answer, then tries to reconnect to server ) . Behavior could be mirrored on UI1 too ( no need to send a ping each sec anymore  )

### DIFF
--- a/src/web_server.cpp
+++ b/src/web_server.cpp
@@ -1067,7 +1067,8 @@ void onWsFrame(MongooseHttpWebSocketConnection *connection, int flags, uint8_t *
     if (doc.containsKey("ping") && doc["ping"].is<int8_t>())
       {
         // answer pong
-        server.sendAll("/ws", "{\"pong\": 1}");
+        connection->send("{\"pong\": 1}");
+        
       }
   }
 }
--- a/src/web_server.cpp
+++ b/src/web_server.cpp
@@ -1060,6 +1060,16 @@ void handleNotFound(MongooseHttpServerRequest *request)
 void onWsFrame(MongooseHttpWebSocketConnection *connection, int flags, uint8_t *data, size_t len)
 {
   DBUGF("Got message %.*s", len, (const char *)data);
+  const size_t capacity = JSON_OBJECT_SIZE(1) + 16;
+  DynamicJsonDocument doc(capacity);
+  DeserializationError error = deserializeJson(doc, data, len);
+  if (!error) {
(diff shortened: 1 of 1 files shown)


src/web_server.cpp
```diff
@@ -1060,6 +1060,16 @@ void handleNotFound(MongooseHttpServerRequest *request)
...
 {
   DBUGF("Got message %.*s", len, (const char *)data);
+  const size_t capacity = JSON_OBJECT_SIZE(1) + 16;
+  DynamicJsonDocument doc(capacity);
+  DeserializationError error = deserializeJson(doc, data, len);
+  if (!error) {
+    if (doc.containsKey("ping") && doc["ping"].is<int8_t>())
+      {
+        // answer pong
+        server.sendAll("/ws", "{\"pong\": 1}");
```
jeremypoulter, line 1070: Should send to the specific client, not all

jeremypoulter (approved):

src/web_server.cpp
KipK: I haven't found the method for that in mongoose lib. However, the clients only send ping when it has not received one for 5sec , so if one client sent all received pong thereafter, and won't send on on their side.

jeremypoulter: `connection.send(....)` should do it

KipK: I think it's more efficient this way, or the server will always have to parse multiple ping requests from different clients as each will send their own then.

KipK: > `connection.send(....)` should do it

ah I was using server object instead of connection. 👍 
But do we really want that ? as above, with 5 ws connections, each will ping the server, and server will have to answer separately to all of them. 
I know it's not much processing, but the other way is using only 1 ping to serve them all

jeremypoulter: Not really one ping, the sendAll loops through all connections and sends a separate message to each. If we want a pong message, it should be in response to the pong message, to an explicit client

KipK: ok done . When you'll merge it, I'll need you to update the ui_v2 branch too.
I'll merge the pingpong ui branch thereafter, it would be good to pull the last version and rebuild dailies with it then

jeremypoulter (approved):

Issue #467, opened 2022-11-04
/settime endpoint problem


KipK:
There's somlething wrong with the time handling.

While integrating the set time using /settime endpoint to new UI, I've spotted a strange bug:

Example:
Current local time is 18h33

Sending to /settime:
ntp: false
tz: Europe/Paris|CET-1CEST,M3.5.0,M10.5.0/3
time: 2022-03-11T17:33:00.000Z  ( sending zulu time ISO format ) 

I receive fiirst from websocket:

{"time":"2022-11-03T**17:33**:26Z","offset":"+0100"}

then few seconds later receiving from /status:

time: "2022-11-03T**16:33**:27Z"

I can see the same problem with current UI when setting manually.

Same when setting with NTP, receiving first a time with one hour shift, then few sec later the time is back to ok.

But sending correct time manually ( with old UI too ) will end with a 2 hours shift 

issues probably related:
https://github.com/OpenEVSE/ESP32_WiFi_V4.x/issues/302
https://github.com/OpenEVSE/ESP32_WiFi_V4.x/issues/349

fhteagle:
I can confirm that issue has existed on the 4.x UI for quite some time. It gets positively maddening with UTC offsets in the -6 range. 

For example, manual time setting will even arbitrarily advance the day number to 1 more than what was entered in the GUI.

KipK:
Yes got this one with date too.

KipK:
> the acknowledgement is just that it has saved the time, not that it has retrieved it from NTP.

mhh , but in the last case, /settime only sent the ntp: true, and no time, so the time sent back should be the one retrieved by NTP isn"'t it ?

jeremypoulter:
Hum, will have to check..

KipK:
I have fixed this issue on UI2 side by setting /config time_zone to UTC when manual or browser time is set.

PR #569, opened 2023-03-11
Various Time zone fixes

### LINKED ISSUE #107: Timezone Issues
Selecting `Europe/London` timezone and NTP results in the correct time being set on the WiFi interface and on the controller :+1: 

IMG_0154

However, selecting `GMT+1` which should be same, results in the correct time being displayed on the WiFi interface, however the controller is 2hrs behind :-1: 

Could a solution just be to remove the GMT+X timezones? Since these seem to be a duplicate of the `Continent/City` timezones.  

IMG_0153

Thanks to Bill for reporting on the forum:https://community.openenergymonitor.org/t/can-openevse-default-be-set-to-eco-mode/11278/14
***

I've also noticed that the timezone dropdown menu does not work on iOS, tested on both iOS Chrome and iOS Safari on an ipad mini:

IMG_0152

jeremypoulter:
- Set the default time zone to the defined one
- skip `|` when setting the env var, stops the timezone from being recognised
- Filter out the Etc/... time zones

Fixes #107


KipK:
perfect, it solves the timer bug, and it probably fix #467 I have to check that later

KipK:
One should push a new release after this merge.  4.1.7 has too much problems. This was the last urgent one to fix.

KipK:
now this fix #467  :+1

jeremypoulter:
At the moment it forces UTC in the code, so it doesn't matter what time zone you send. I am going to look at setting the time zone for manual set times as the DST info from the timezone is handy. Also need to figure out how to handle a different timezone of the EVSE and the browser. The ESP now sends the local and GMT times (according to the ESP) but this is then displayed in the timezone of the browser by default. Not sure if that is much of an issue other than for testing.

KipK:
strage, because if i send to /settime 
ntp: false
tz: Europe/Paris|CET-1CEST,M3.5.0,M10.5.0/3
time: 2023-03-13T08:55:00.000+01:00

It then has 
local_time: "2023-03-13T08:55:00+0000"
time: "2023-03-13T08:55:00Z"

KipK:
corrected I sent it wrong, but same problem:

ntp: false
tz: Europe/Paris|CET-1CEST,M3.5.0,M10.5.0/3
time: 2023-03-13T09:19:33.023+01:00

gives:
local_time: "2023-03-13T09:19:33+0000"
time: "**2023-03-13T09:19:33Z"**

jeremypoulter:
Yeah, I think I need to update it to properly handle timezones for manual mode.... :) <https://youtu.be/-5wpm-gesOY>

KipK:
ahahah that's insane reality we have to deal with :)))

KipK:
Can it be merged yet? Even if you change the manual tz later, it already works ok

jeremypoulter:
Found many more timezone bugs, and fixed most of them, but this is still WIP

jeremypoulter:
I think the main task left is to work out how to set the time manually in the EVSE's timezone

KipK:
I've updated the UI2 for new /limit. 

Works ok for manual without tricking to UTC.
UI sends time as UTC + timezone and we're getting back correct local time

KipK:
all good here. Thanks for this

KipK:
@glynhudson can you merge this one ?

PR #580, opened 2023-03-27
add missing energyMeter data to emoncmss_publish()

KipK:
forgot this one

Also put the deprecated properties in create_rapi_json as they are missing for emoncms publish

( time to ask emoncms to use the new properties before deprecation )

edit: is there a good reason why data are sent separately to emoncms instead of using event_send like websocket & mqtt?
Also it seems we should send the full /status topic to emoncms instead of create_rapi_json as more data are now coming from ESP32 and not RAPI.

### DIFF
--- a/src/energy_meter.cpp
+++ b/src/energy_meter.cpp
@@ -8,7 +8,6 @@
 #include "app_config.h"
 #include "event.h"
 
-
 EnergyMeterData::EnergyMeterData()
 {
     total = 0;
--- a/src/input.cpp
+++ b/src/input.cpp
@@ -147,6 +147,12 @@ void create_rapi_json(JsonDocument &doc)
   doc["srssi"] = WiFi.RSSI();
 
   timeManager.serialise(doc);
+  // get EnergyMeter data
+  evse.createEnergyMeterJsonDoc(doc);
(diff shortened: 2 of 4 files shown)


jeremypoulter:
EmonCMS only accepts numeric data and does not necessarily need all the data exposed by event_send. Also as EmonCMS will typically be remote and WS/MQTT local sending less data is best.

KipK:
I'm agree, and not sure :) I've wanted to move it out of this create_rapi_json() because it's not RAPI data.
Saying this, there's other data not coming from RAPI in this function like manual_override, freeram, divertmode, srssi, time.
Wouldn't it be cleaner to start separate thoses in prepration of a future properties refactoring ?

jeremypoulter:
For now it is probably best to make that change, we can rename that function to make it clearer that it is not just RAPI if you want.

KipK:
it's ready for merge

KipK:
rebased and ready

jeremypoulter (approved):

glynhudson:
I've just tested this and `total_energy` doesn't seem to be posting to Emoncms.org

Emoncms-input-view

PR #583, opened 2023-03-29
some EnergyMeter fixes

KipK:
fix monthly counter wrong
session data initialised at reboot
fix time now using localtime for rotate()
publish energyMeter after counters reset

### DIFF
--- a/src/energy_meter.cpp
+++ b/src/energy_meter.cpp
@@ -79,10 +79,10 @@ void EnergyMeterData::deserialize(StaticJsonDocument<capacity> &doc)
         // daily
         daily = doc["dy"];
     }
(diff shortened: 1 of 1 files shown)


KipK:
I still have a problem with the en energyMeter rotate() with new time engine. 

Could you help me to solve this ? Seems the time I get now is GMT and not local:

I don't what to change here: 

https://github.com/OpenEVSE/ESP32_WiFi_V4.x/blob/2d34a3a9bc817aa77b02f8868a97c976d73eed5d/src/energy_meter.cpp#L485

KipK:
Can we merge this ?

jeremypoulter (approved):

Issue #587, opened 2023-03-31
Feature Request: Minimum Idle Time with reset for relay wear reduction and avoiding power peaks


pdhoogh:
The objective of this feature is to reduce the cost of using OpenEVSE for charging an EV by avoiding grid import power peaks and charger relay cycles. 

Avoiding grid import power peaks reduces the utility bill. Avoiding unnecessary relay cycles saves relay life and avoids the car being charged going in "Charge Error" mode. Avoiding the EV to go in "Charge Error" mode is also reducing cost as it is improving the ability to use Solar to charge the EV. In "Charge Error" mode, the Solar PV installation may produce power that needs to be injected in the grid as the EV is not accepting the charge. The EV then may need to be charged with import power later on.

I am re-iterating this as requested by @jeremypoulter and agreeing with @soloam.

The proposal is to turn off the charging (> Disable) as soon as the "available current" drops below 6 amps. A timer is started at the same time. The timer is reset every time the available current is calculated to be below 6 amps. When the timer runs out, the charge is resumed.

This will avoid import power peaks, because the EV charging is turned off as soon as the import power peak is detected. It will also save the relay in the OpenEVSE, as with this feature, OpenEVSE will wait for stable power availability before it resumes charging. A number of home appliances are cycling power with short intervals:

- Laundry Dryers
- Ironing stations
- Microwave ovens that run at "reduced power setting"
- Washing machines
- Dishwashers
- Solar PV installation with half open skies
- more I forgot to add...

With the current management of the relay, switching ON/OFF the charge may happen too frequently or to quickly, causing the EV to go in "Charge Error" mode. My Peugeot e-208 does this a lot. The current 10 minute minimum charge time feature is not always enough to avoid this.

I am repeating @soloam's example below:

Let's assume a idle time of 10 minutes.

10:00 - Goes bellow 6 - System goes idle
10:30 - Remains bellow 6 - Remains Idle
10:35 - Goes to 10 - Remains idle for 10 min
10:40 - Goes bellow 6 - Remains idle and reset timer
10:50 - Goes to 10 - Remains idle for 10 min
11:00 - Remains over 6 - Resumes Charging

One can implement this using MQTT or API commands, but native OpenEVSE would of course be better.

jeremypoulter:
That is what the smoothing attack and decay values do.

[...]

PR #590, opened 2023-04-01
New InputFilter class

KipK:
Following #587 issue + discussion
Changed for a new generic inputFilter class not dependant of fixed sample frequency.
This will calculate automatically previous filter ratios depending of data submission delta ( thanks to @pdhoogh for taking time to explain what looked like dark magic to me ) 
Should probably fix some of the issues posted with graphic showing unprevisible divert results

shaper ok
Divert ok
UI2 gui ok (shaper branch)
UI1 gui ok (https://github.com/OpenEVSE/openevse_wifi_gui/pull/98)


soloam:
This is great! Thank you! When will it hit a release? Thank you

jeremypoulter:
I think the reason for this change got lost in between the other conversations, could you clarify the use case? I am not totally sure this is an appropriate change. To my mind the primary use case for the shaper is for things like load balancing and controlling the maximum current that can be delivered.

It is much more of a safty feature and should be fast to react, both in terms of switching the current down and back up again. That is the reason we created this as a separate module rather than trying to make the divert fit that use case. 

At a minimum, I think to accept this PR it would need to be configurable and the default to be off, but waiting to hear the input on the use case.

soloam:
Like discussed, the main use case for this is to ensure that a safety feature is not responsible for a fast wear out of the relay! Attenuating the number of times that the relay goes on and off when the power consumption is going up and down! Like stated, a low pass filter would help, only when the relay is off and need to go on, because when the power is lowering, the ideia is that is instant! It's preferable that the power goes down and only some minutes latter goes up... Than waiting or delaying lowering the power and risk a power cut!

Tks for all the time and attention dedicated to this issue

jeremypoulter:
Then IMHO the solution is to add the smoothing on activation as with the divert, so off instantly then apply the smothing to the recovery. I don't even think we should apply any filtering to the reduction of the current at all, even optional.

jeremypoulter:
That being said is it not reasonable to expect that the EVSE should be installed in situations where it is gauranteed at least 6A? Hense the Shapper having to turn off the current should be an exception should it not?

soloam:
In my personal situation I get a lot turn off while charging and continuing to use the house to other activities! And I have 32 Amps... A lot of people here in my country don't go over 20! In that situations the shaper is a must have! Ensuring that while the house is consuming more that the charger is lower or off... And when late at night that most of the house is idle, the charger can go full power!

jeremypoulter:
Ok point accepted, 32A for the entier property is fairly easy to use up if that is typical in some locations. I very much think that the correct solution to this issue is a configurable smoothing on the attack. I have never found a case yet where an arbitory time is the correct solution to a problem.

soloam:
At this point I'm good with either solution... If that results on my relay not going on and off several times because of momentary spikes or lows!

src/current_shaper.cpp
```diff
@@ -18,21 +25,30 @@ void CurrentShaperTask::setup() {
...
+					_pause_timer = 0;
 					props.setState(EvseState::None);
 				}
 				_changed = false;
 				_updated = true;
 				_timer = millis();
-				evse.claim(EvseClient_OpenEVSE_Shaper,EvseManager_Priority_Safety, props);
+				// claim only if we have change
+				if (evse.getState() != props.getState() || evse.getChargeCurrent() != props.getChargeCurrent()) {
+					evse.claim(EvseClient_OpenEVSE_Shaper, EvseManager_Priority_Safety, props);
```
jeremypoulter, line 50: Not sure about this, I think it should just claim and let the EVSE manager manage what to do.

KipK: I've added that to not spam the claims, the UI was receiving lot of version change that we don't need.

KipK:
> I think the reason for this change got lost in between the other conversations, could you clarify the use case? I am not totally sure this is an appropriate change. To my mind the primary use case for the shaper is for things like load balancing and controlling the maximum current that can be delivered.
> 
> It is much more of a safty feature and should be fast to react, both in terms of switching the current down and back up again. That is the reason we created this as a separate module rather than trying to make the divert fit that use case.
> 
> At a minimum, I think to accept this PR it would need to be configurable and the default to be off, but waiting to hear the input on the use case.

It already goes to disable immediatly. It just takes 5mn to start again or more if needed until it's stable enough to start.
This is temporary to have at least this for the current prerelease until I find time to make it configurable.

I also have only 39A available and can override easily with my eletric heaters in winter.

KipK:
@jeremypoulter , thinking of future shaper rewrite to prepare it for handling thermal throttling too and beeing configurable.
Do you think I should move shaper class to evseManager like I did for energyMeter ?

KipK:
I did this to take update interval in the loop:

`	// filtering live_pwr for getting out of pause
	uint32_t updt_time = (millis() - _timer) / 1000;
	// default filtering would consider an update each 10sec. 
	// Calculating the smoothing factor considering current interval.
	double factor = (updt_time * current_shaper_smoothing_factor) / 10.0;
	_smoothed_live_pwr = (_live_pwr * factor) + (_smoothed_live_pwr * (1 - factor));`

jeremypoulter:
I wouldn't do either, the thermal throttling should just be a separate module and both should create claims so can live outside of the EVSE manager and call it

pdhoogh:
The math should be the same as already discussed elsewhere. The module needs to know the update interval of the power data in minutes and the time constant then needs to be a configurable value also expressed in minutes. Zero means no filter (immediate action). Every time the module wants to update the current setpoint it should get the last update interval in minutes and use this formula to come up with the filter factor: 
image
DeltaT is the update interval in minutes. Tau is the time constant of the exponential filter. From that formula it is clear a divide by zero needs to be avoided, so tau = zero must lead to a filter factor of 1 (no filter) and skip the computation of the formula above. A practical low limit for tau like 0.1 is a good value to avoid very large numbers generated by the division operation needed for the exponential. So the data validation should allow zero for tau, but values between zero and 0.1 should result in using 0.1. A time constant smaller than 6 seconds (0.1 minutes) does not make much sense with a sample rate of ten seconds or more ([Shannon et. al.](https://en.wikipedia.org/wiki/Nyquist–Shannon_sampling_theorem))

The result alpha is the filter factor used by the OpenEVSE today. So the implementation of the filter does not need to change. Only that the filter factor is now calculated instead of entered directly.

I have not yet tried the @jeremypoulter idea of using the attack and decay filter factors yet because the weather here is so bad my PV installation barely produces enough power to cover my home's low power state:
image
So unless we go to LIDL for groceries where I can charge the EV for free, we're back to using the Diesel Merc as long as the weather stays like this. Home import power is too expensive here for charging the EV, diesel is cheaper.

From @jeremypoulter simulation graphs I now think I better understand how the filters work and if my understanding is correct, I believe what @jeremypoulter proposes, using the attack/decay, is close to what is requested:

- The actual amp setpoint calculated from the +I/-E is unfiltered and aims to balance power immediately
- The Amps available is filtered using attack / decay exponential low pass filters
- The filtered Amps available is compared to the 6 amps minimum level to decide if the OpenEVSE is enabled or Disabled
- There is still the possibility for a chattering relay if the filtered Amps available hovers around 6 Amps. Therefore a configurable minimum charge time is implemented. That minimizes "wasting PV power". Elsewhere on this site we have a discussion where I request to replace this by a configurable minimum disable time. So we adapt to new legislation and tariff structures. Or avoid using more power than the home installation allows. This should not be hard, as it almost exists today. There is a random timer for enabling the OpenEVSE to implement some kind of reacceleration scheme, avoiding all OpenEVSE's to enable at the same time. The configurable minimum disable time can be just the minimum value used in the random number generator.

Is my understanding correct @jeremypoulter ? 

Based on the above understanding I have the attack filter factor set to 0.033 (5 minutes time constant) and the decay filter factor set to 0.5 (about 15 seconds time constant), allowing for some debouncing while high power duration is too short to trip safeties or get registered as a 15 minute sustained peak.

KipK:
Do we really need to let the Tau beeing configurable ? ( or just set in #define , i use 10 sec as time reference interval )

pdhoogh:
Yes, Tau needs to be configurable.

I will try to clarify. First recapping above:

_Every time the module wants to update the current setpoint it should get the last update interval in minutes and use this formula to come up with the filter factor: 
image
DeltaT is the update interval in minutes. Tau is the time constant of the exponential filter. From that formula it is clear a divide by zero needs to be avoided, so tau = zero must lead to a filter factor of 1 (no filter) and skip the computation of the formula above._ 

Tau is the exponential filter constant in minutes. This is essentially the same intended functionality as the filter factors that are configurable now for attack and decay. Only the new function for these configurable parameters would become the filter time constants in minutes rather than the filter factors straight.

In the time domain, the effect of the filter using only the filter factor as @jeremypoulter described depends on the rate at which the calculation is executed. This can easily be understood as follows: in the current implementation, when the filter calc is run, the new filtered value for a filter factor of 0.1, is calculated as 90% of the previous filtered value + 10% of the current unfiltered value. Every time this runs, the filtered value is updated accordingly. So one can see that the filtered value will move twice as fast in the time domain if the calculation is done at half the period. E.g. if you calc this every five seconds, the filtered value will move twice as fast as if the calc is done every ten seconds. So the behavior in the time domain also depends on the calc period.

To transform the response of the filtered value to a fixed behavior in the time domain, the filter factor needs to be compensated for the calculation period. E.g. the filter factor needs to be changed so that the effect of a 10% share in ten seconds is the same as two x% shares in 2 times 5 seconds.

So yes @KipK you could fix DeltaT to 10 seconds (0.166 minutes), but that assumes the update frequency of the data comes in exactly every ten seconds. And as you described before, that is not always the case. There is a timer that keeps track of the MQTT update period already. You can use that value just before the last update, in minutes, for the deltaT.

The exponential function is needed for accuracy with the smaller time constants, as the simplified expression is an approximation that is only valid for tau much larger than DeltaT. Below is a graph showing the exponential and the approximation 
image
It works well for a time constant of 1 minute or longer
image
But for faster filters, that deteriorates and that may be the practical range for at least the decay value
image

KipK:
@pdhoogh 
>So yes @KipK you could fix DeltaT to 10 seconds (0.166 minutes), but that assumes the update frequency of the data comes in exactly every ten seconds. 

in fact I did it a bit differently. 
The 10 sec is just a constant for the calculation. 
I've tested it and seems to works good like this 
 
Any cons ? 

```
#define EVSE_SHAPER_FILTER_TAU 10 // 10 sec

double factor = (updt_time * current_shaper_smoothing_factor) / EVSE_SHAPER_FILTER_TAU;
_smoothed_live_pwr = (_live_pwr * factor) + (_smoothed_live_pwr * (1 - factor));

```

KipK:
@jeremypoulter I've rebased the PR with filtering and configurable UI is ok on shaper branch too.
Tested for what I could, seems all ok. 

we now can just use timer based pause if the filter ratio is set to 1 using current_shaper_min_pause_time,
or use filtering + current_shaper_min_pause_time just in case the filter has wrong settings.

pdhoogh:
> @pdhoogh
> 
> > So yes @KipK you could fix DeltaT to 10 seconds (0.166 minutes), but that assumes the update frequency of the data comes in exactly every ten seconds.
> 
> in fact I did it a bit differently. The 10 sec ( EVSE_SHAPER_FILTER_TAU ) is just a constant for the calculation, but then I change the smoothing factor proportional to the time difference with EVSE_SHAPER_FILTER_TAU I've tested it and seems to works good like this I can spam data in short intervals or lagging data, it doesn't change the way it reacts

Do I see correctly that you switched the variable names and see tau as the calc period, not the filter time constant?

KipK:
no EVSE_SHAPER_FILTER_TAU is a constant of 10 sec ( ideal world update refresh )

current_shaper_smoothing_factor is configurable over ui, default set to 0.03
min 0.01
1 would mean it doesn't filter. 

"updt_time" is the real time we got between the last 2 update

KipK:
this is how it looks on ui

image

KipK:
It needs some smoothing when you are around the 6amp limit. Or it can switch back and forth quiet quickly. 
This is like for divert to prevent the relay wears.

When relay is switched off we needs to wait a bit of time before going on again. 
The smoothing stop when it's back to active

jeremypoulter:
Was thinking about this a bit, normally you would have a higher on value than off to stop this, eg off at 6A on at 6.5A. might be worth a try for both the shaper and the divert. Probably should think about splitting out the filtering/set point code into a common class...

KipK:
taking time to read @pdhoogh filtering explanations above ( quiet interesting thanks for that )
I like the idea of common users have a simple time parameter to configure instead of mystical ratio values.

I'm trying to implement that now

jeremypoulter:
Can we do that as a separate PR, and let's do it for both divert and shaper please.

soloam:
> this is how it looks on ui
> 
> image
> 
> 

Will this be the new layout for configuring the shaper?

KipK:
> Can we do that as a separate PR, and let's do it for both divert and shaper please.

yes.

Can you give a look on the filter implementation here ? 
So on UI now you only set the input filter time constant in seconds ( or whatever you want it to be called ) 👍 
image

src/current_shaper.cpp
```diff
@@ -118,20 +149,60 @@ void CurrentShaperTask::shapeCurrent() {
...
+			if (current_shaper_smoothing_time) {
+				if (current_shaper_smoothing_time < EVSE_SHAPER_MIN_FILTER) {
+						current_shaper_smoothing_time = EVSE_SHAPER_MIN_FILTER;
+				}
+				factor = 1 - exp((-1) * (updt_time / (current_shaper_smoothing_time)));
+			}
+			else {
+				// avoid divide by 0
+				factor = 1;
+			}
```
KipK, line 178: @pdhoogh here is the implementation

KipK:
Can't we merge it like this for now, and I'll switch to generic inputFilter class on the next PR

I have rebased this PR in one commit so a bit painfull to reparate thigs.

KipK:
I still have _smoothed_live_pwr going to undefined after a pause, have to find why

pdhoogh:
> no EVSE_SHAPER_FILTER_TAU is a constant of 10 sec ( ideal world update refresh )
> 
> "updt_time" is the real time we got between the last 2 update

@KipK that is indeed what I thought I understood from you. 
For clarification, the filter formula
image
reads alpha equals DeltaT divided by Tau. That is the approximation, but I use it now to show what Tau really is.

DeltaT is the scan period and is what you refer to as "updt_time". That can be 10 seconds fixed, assuming perfect updates.
Tau on the other hand is the "time constant" of the filter and - sorry for this - that is control theory lingo for describing the behavior in the time domain, but in our discussion it is by no means "constant". If Tau doubles, the filter slows down and will take twice the time to get to the unfiltered value.

In the currently released firmware, the attack and decay values are the alphas in the above formula. These are settable. So if we wish to move to settable time constants, we should replace those settable alphas by settable tau's and use the formulae as described above to calculate the alphas from the settable taus using the formula that takes into account the update period. That way, the settable time constants (taus) can be used to change the filter behavior in the time domain and make it independent of the update frequency (by compensating for it).

This is not a linear function. It is an exponential.

> 
> And there's also a current_shaper_min_pause_time set in the config, to add a minimal pause.

Great!

We have plenty of sun now and I'm charging the EV using Divert and with setting the attack / decay alphas as suggested by @jeremypoulter. So far it seems to work fine. I'll let it run for a few days to see what my evaluation is. It is promising and may well be all I need.

KipK:
> > no EVSE_SHAPER_FILTER_TAU is a constant of 10 sec ( ideal world update refresh )
> > "updt_time" is the real time we got between the last 2 update
> 
> @KipK that is indeed what I thought I understood from you. For clarification, the filter formula image reads alpha equals DeltaT divided by Tau. That is the approximation, but I use it now to show what Tau really is.
> 
> DeltaT is the scan period and is what you refer to as "updt_time". That can be 10 seconds fixed, assuming perfect updates. Tau on the other hand is the "time constant" of the filter and - sorry for this - that is control theory lingo for describing the behavior in the time domain, but in our discussion it is by no means "constant". If Tau doubles, the filter slows down and will take twice the time to get to the unfiltered value.
> 
> In the currently released firmware, the attack and decay values are the alphas in the above formula. These are settable. So if we wish to move to settable time constants, we should replace those settable alphas by settable tau's and use the formulae as described above to calculate the alphas from the settable taus using the formula that takes into account the update period. That way, the settable time constants (taus) can be used to change the filter behavior in the time domain and make it independent of the update frequency (by compensating for it).
> 
> This is not a linear function. It is an exponential.
> 
> > And there's also a current_shaper_min_pause_time set in the config, to add a minimal pause.
> 
> Great!
> 
> We have plenty of sun now and I'm charging the EV using Divert and with setting the attack / decay alphas as suggested by @jeremypoulter. So far it seems to work fine. I'll let it run for a few days to see what my evaluation is. It is promising and may well be all I need.

Yeah this comment was a bit old. I've implemented the exponential formula since and works good now so far.
You now set tau as second ( new defaults for divert are now 20 for attack and 120 for decay but they are temporary and need your input on this )

Do you want a build to test ?

pdhoogh:
Ah, great! Since you have DeltaT already in seconds it indeed makes sense to do the time constants tau in seconds. 

I'll wait for you to hammer out the issues.

In my ExxonMobil life I was also a trainer and I have taught many process control courses to our licensees. I am patient. And if the student misunderstood it is usually me not understanding what knowledge I can assume from the students. So don't hold back on asking for clarification. There are no stupid questions, only stupid answers.

As for the defaults, I have no preference, really, as it completely depends on the use case. I'm OK using the current defaults for the alphas back calculated to taus for a 10 second update period. As a provider of firmware the most important question to ask here is what default settings will cause the least reactions or questions from the new users. So the "common denominator" for the use cases should set defaults. I guess you guys have a lot more experience with this now than I do for the OpenEVSE.

If we assume the use case I have and @soloam has, it looks like a tau for decay of 10 to 20 seconds is what we look for, cutting out the EV charge fast when a power import peak is presented. For the attack, the idea is to wait for "stable power" as someone put it. So a tau for attack of 5 to 10 minutes would be more appropriate, i.e. 300 to 600 seconds.

The taus you propose are closer to the original intent of "not wasting solar power". That is fine as long as the utility companies let you without additional cost.

This remains a feedback control system and there is only so much you can achieve with it. Settings will never be perfect for all circumstances. To further improve things is beyond the scope of OpenEVSE. It requires a CEMS with predictive control. That is more my peace of cake :-).

KipK:
@pdhoogh , that's interesting. So I could propose on the UI some kind of presettings with booth behaviors.
Idea for merging this fast without too much users feed of question would be to have default values matchings behavior of the one we had before with the ratio. 

then I can propose other behaviors throught the UI for common use cases.

KipK:
Or let's make a release now with current UI2 master, and then we're good to move on for a major release with this one and UI switch, and takes time to test it

pdhoogh:
> So having theoretical default values that would mimic previous default settings + some different presettings hardcoded on UI, would be a good compromise for smooth transition.

That is an improvement, calling this a compromise does not do justice to your work. You're doing great!

KipK:
with presettings on UI, looks good:

image

pdhoogh:
C'est SUPERBE, Guillaume! Merci Beaucoup!

KipK:
> 

thanks, i've fixed it. Was a mistake.

KipK:
> This is great! Thank you! When will it hit a release? Thank you

@soloam you can test it on the build above, but this is clearly untested in live condition yet. 
No much consequencies anyway, you can roll back. 

https://drive.google.com/open?id=1FjYBbFW2SXE_5jxBuc-knzxWo3vFwJwK&authuser=hotlobster75%40gmail.com&usp=drive_fs

KipK:
@pdhoogh I've fixed the divert/shaper concurrency.
Now shaper add energy excedent to its total power available and is supervising over Divert. 
So booth can concurrency handle their work now, shaper manager max_current, and divert charge_current. It can't goes over it and overriding the circuit braker should not be possible ( if data are not lagging )

jeremypoulter:
Yeah that was the plan with the EVSE Manager, each module does it's thing and the manager sorts out the final state. As the shaper is a safety module so has a higher priority than the divert so will override it. Lots of great work here thanks, I do need to look at it a bit more in depth so will probably not be merged quickly. I need to make sure it operates as expected across all the situations, which means fixing up the divert simulator.

KipK:
Should we have the hysteresis configurable or constant of 0.5A seems universally acceptable?
It's not much to add, but it's another frightening setting on the interface.. :)
I'll make a simple/advanced version of some setup page someday

KipK:
@jeremypoulter 
I see in divert code the divert_pv_ratio is used in the calculation for booth production & I/E grid mode.
But UI1 is only displaying the divert_pv_ratio field in the config for I/E grid mode. 

double trigger_current = _evse->getMinCurrent() * min(1.0, divert_PV_ratio);

Do I have to reset divert_PV_ratio to 1 when user switch to Production mode ? Or does this field should be displayed for booth mode ?

jeremypoulter:
I think it is applicable to both, don't recall why it is only available for Grid I/E will check

KipK:
> I think it is applicable to both, don't recall why it is only available for Grid I/E will check

Also we should add a real setting for divert mode. As we can push http data now, we shouldn't test the mqtt fields anymore.

KipK:
That makes me think, shaper for now use filtering to get out of pause only. But perhaps it would be better to ramp up the input signal smoothly too.
Should take some time to get from 6 to 32 A. On the other way down, we could have some low Tau filter that reacts quick but can smooth a bit sudden spikes

pdhoogh:
@KipK said:

> We will probably switch to uiV2 for next major release anyway.
@pdhoogh That would be nice if you could test the new input filter and propose some good default values for the presettings ( the one I've set for now are just empirical based on your explanation but we can probably improve this.
The presettings I've made only change the filter values, but perhaps they should include the minimum pause time too? ( It default to 5mn for now perhaps it's a bit too much used with the filter.

I am now using GUIv2 downloaded 26 Feb with the filter factors. Which one would you like me to test?

Including the minimum charge time is a good idea, just make sure users know this.

The filter factors for the "No Waste" can be equivalent to the values originally used for defaults as that was the use case they were intended for. I don't remember what these were, but time constants can be calculated with this formula: 
image
Assuming 10 seconds DeltaT and time constant in seconds, this is what I get:
image
For the "No Import" case I'm currently using following time constants:

Divert smoothing attack (alpha = 0.033): 300 seconds (5 minutes)
Divert smoothing decay (alpha = 0.7): 8 seconds
Minimum charge time: 20 seconds

These work for me. The minimum charge time depends on the EV, how fast it can go through its charge initialisation. Mine is a Peugeot e-208. Others may be different. The aim is to avoid "Error" by the EV charger.

I'm not sure what the use case for "Average" would be?

And putting in defaults for "Custom" is a good idea to avoid crashing the code. These can be the same as "No Waste" to be friendly to @jeremypoulter. My opinion is "No Import" will gain traction in the future.

KipK:
Thanks a lot for this. About the Average, it's just a dummy setting I've put there, needs to be replaced by some real use case or remove it. 
It's free for changes ;)

KipK:
The default is based on your current settings. If it matches one presetting it will check this one if not it checks curtom

soloam:
> > This is great! Thank you! When will it hit a release? Thank you
> 
> @soloam you can test it on the build above, but divert part is clearly untested in live condition yet. 
> No much consequencies anyway, you can roll back. 
> 
> https://drive.google.com/open?id=1FjYBbFW2SXE_5jxBuc-knzxWo3vFwJwK&authuser=hotlobster75%40gmail.com&usp=drive_fs

I tested this out and looks OK! I'll do more tests, but so far, so good! Will this be merged?

KipK:
Not yet. This needs to be a bit more tested first.
And perhaps change a few things.

pdhoogh:
OK @KipK, I'll test the beta you want me to review when you tell me it is ready and where it is.

KipK:
@pdhoogh there's a wifi_v1 build here https://drive.google.com/drive/folders/1FjYBbFW2SXE_5jxBuc-knzxWo3vFwJwK
Tell me if you need other platofrm. 

I get your point on Shaper, that needs to be discussed, could be interesting to simplify the whole setup. A minimum import from grid parameter makes lot of sense.
That has a bit of collaterals for those who charge using booth divert & timer based import, and probably other use case. So there's some change in the current codebase to fit this new "vision".

jeremypoulter:
The key difference is more about what they allow/prevent other modules from doing.

Divert will be off by default and switch on when the power when there is sufficient capacity. Also importantly allows other modules to enable and/or override the current, eg timer over night, or manual full power boost.

The shapper on the other hand is active by 'default' and won't allow overriding, eg the timer overnight won't work if disabled and you can't boost above the current, for example. That is why they were implemented as separate modules in the first place.

pdhoogh:
@KipK : image

@jeremypoulter : that is perfectly fine as long as all this works reliably. And so far in my experience it does not. That is why I was trying to help. Simplifying may help.

KipK:
>  And so far in my experience it does not. That is why I was trying to help. Simplifying may help.

But it's supposed to be fixed since few monthes in dev build ( not in 4.1.7 ) .
Should be ok on the one you test here. You need to have cookie enabled on your browser to get in google drive.

KipK:
@pdhoogh I've put it here if you prefer: https://github.com/KipK/openevse-gui-v2/releases/tag/inputfilter-0.1.0

pdhoogh:
Thanks! I'm on it. No sun today, no charging. UI looks the part, great!
image

KipK:
@pdhoogh I've updated the build. It just fix a bug with energyMeter and UI update.

pdhoogh:
OK, same location as previous? I'm using the 7d3abd17 now.

KipK:
> OK, same location as previous? I'm using the [7d3abd1](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/commit/7d3abd173ec1a003ae381ae622476cf6a4ba9194) now.

yes I've updated booth gh and drive

pdhoogh:
OK. 7d3abd1 seems to work well. At first glance it looks like the attack time constant is a bit too fast. That is the downside of a filter. With available current at 10 amps, it takes some time to reach 6 amps, but with available current at 20 amps the filtered value reaches 6 amps a lot faster. It will take some time to find the "sweet spot" values. The challenge today is clouds and when no clouds, it is cold outside and very bright sunlight... the PV panels are having a hay day.
image

KipK:
Wouldn't it be better then to limit the max input to like 8-10 amps first until EVSE has gone in active state?

pdhoogh:
@KipK I would not encourage you to do this current value limiting for the filter.

I appreciate your zeal! I was not complaining. I know there is a difference between a timer and a filter approach and that is what I was pointing to. It's fine. What I will probably end up doing is taking a bit of a larger attack time constant and trade that off with "wasting" a little bit more when there is less power available. The important thing here is that it works perfectly as is with the small decay time constant, the power to the EV charger is quickly cut and that is what is important to me. Maybe we could have a bit different default values for the "No Import" case, but that would only work for me. Anyone's case is specific. So what works for me may not work for others. So even changing the default values is not really relevant. 

I like simplicity as you have probably gathered already, so I won't encourage you to make things more complicated. Your idea is a really good one to reduce the dependency of the available current on the behavior in the time domain. If that makes sense... It would require extra text in the GUI to explain so people would not get confused... these inner workings can get really mysterious to those who have not followed the development and may generate quite a bit of questions and "issues".

You already did a fantastic job. The GUI is much better, the time constants speak to me and the EVSE has been behaving really well on the last firmwares I've used. 

The only stinging issue left - that I no longer have - is that it seems WiFi can be a bit erratic when using channel 6. Channel 4 works fine and I saw it also did for someone else who had really strong WiFi at the OpenEVSE location. That probably is not a firmware issue but may be interference by the other electronics in the case as you proposed. I may at a later date test that hypothesis by wrapping  the other electronics in a grounded aluminium sheet.

pdhoogh:
Ah! I did see something that may be an issue and cause strange behavior. When the charge current is switched on, the calculated "available current" dips down for a cycle. That is an artifact that probably has to do with timing. If the EVSE would act on this spike down, it may be an issue. It does on the filtered value... that may induce ON/OFF/ON/OFF-behavior.
image

pdhoogh:
@KipK in the particular case above, it was when I had unplugged and re-plugged the EV. No changes in the EVSE. EVSE wanted to start charging and the EV accepted after being re-plugged. 

Usually, when in Divert and the EVSE sees not enough available current, it goes in "Disable" or "Pause". With the e-208 it has to be "Disable". When the EVSE decides there is enough current, it enables the charge. When the EV accepts the charge and starts drawing current, I sometimes see the same dip in "available current", but not always, mostly not actually. That has to do with - I guess - the EV going through an initialization where it draws about 0.7 Amps at first, before accepting the current proposed by the EVSE. That does not do anything meaningful to the "available current", but a small dip can sometimes be noticed. 

The EV is not always consistent in its behavior. Sometimes, when enabling, it first draws the full proposed current for some time, then remembers it should initialize, drops the charge current and goes through the normal procedure. So Stellantis are not bug free either. And that is precisely what happened in the trend example above. You can see the EV first draws a lot of current, than drops it to 0.7, initializes and goes back to the current proposed by the EVSE. And that seems to confuse the EVSE a bit. Everything recovers after one scan, so it seems all is OK in the end. But these transients might cause erratic behavior in the end and I'm not saying it is the EVSE or the EV to blame here :-).

KipK:
@pdhoogh this could also be the AUTO_AMPACITY bug in the evse fw. 
It has been fixed but there's no build yet available. I can make one for you if needed.
you have untethered cable on evse isn't it ?

pdhoogh:
@KipK I indeed untethered. No need for special treatment, I'll wait for the official build, unless you want me to Beta-test.

pdhoogh:
@KipK I have OpenEVSE.
image

KipK:
@pdhoogh so this is not AUTO_AMPACITY bug , you are tethered ( the charge cable from openevse to your car ) . 
Untethered has a TYPE 2 socket and needs to mesure cable ampacity. This is the part that has a problem and started charge full amp. But if you have 7.1.3.EU fw  this is shouldn't be the problem then.

pdhoogh:
Ah yes, sorry for the confusion :-). I unplugged the car and called it "unteathered the car". But indeed the charging station is one with the cable attached permanently to it. So the EVSE is a teathered one, indeed. 

Sun today and yesterday we went to Brussels with the EV, so charge needed... I'll see how it does today.

I have a dynamic tariff for import / export, prices have come down quite a bit. Import is now 15 c€, 30c€ with all the Belgian taxes and charges added, and export is a bit over 6c€. Km cost on the sun is now again a tenth of diesel so the EV is back in business !! 

I much prefer a teathered EVSE, you dont have to mess around with cables in your car. The e-208's cables - the included granny charger and a Type 2 Mennekes cable - are nicely tuck in the included bag and stuck to the side of the boot space. Charging is almost always on teathered chargers for me. The one at home for slow charging or DC chargers. Even the AC chargers at LIDL are teathered. Chargers in hotels are often teathered too. So I almost never have to take out the cables from the boot. Zalig.

KipK:
In France now all new AC evse installed even at home have to be untethered with a protection on the T2 socket. 
I thought it was the same for most Europe countries

pdhoogh:
Not in Belgium. Yet? Teathered AC home chargers are still sold here.

As good as all public AC charge posts are unteathered in Belgium. I rarely use them.

After having seen the mangled cables careless users destroyed and leave hanging around in rain puddles, I can see where the new guidelines come from :-). But DC chargers will have to remain teathered. Maintenance is a nightmare.

KipK:
@jeremypoulter could you review this yet?

jeremypoulter:
Working on it, fixing the divert simulator so we can test, #609 , nearly there just need to add a method to set the config vars so we can compare master and this PR

pdhoogh:
Still using this firmware and it works very well still. However, I saw 4.2.2 is now official but for some reason it does not let me load new firmware. When I hit the Update button, nothing happens?

image

KipK:
@pdhoogh yes it's a little bug since we've changed filename on the release. You can update manually by going to http://openevseIP/update it's fixed on stable

pdhoogh:
Thanks! Awesome support :-)

Since it is cloudy today I'm trying out Shaper. Works well so far. I disabled self production because there is not enough for 6 Amps. 

One minor minor minor nag: the timeconstant does not mention units of measure (secs, I guess). It does on the divert settings.

pdhoogh:
Ah, official 4.2.2 still has filter factors... ?

KipK:
Yes it still not merged yet.

pdhoogh:
> Yes it still not merged yet.

Pas de problème, tu es le meilleur, Guillaume ! Je ne savais pas, c'est tout...

I have been using local_532c73e4_modified for the best part of this month in divert mode. With @jeremypoulter's filter suggestion and the changes made by @Kipk I really do not have anything left to be desired as far as OpenEVSE's intended functionality is concerned in divert mode.

I have only begun to use shaper, so far it looked good, but it really has not had a lot of use time yet. I keep shaper off while using divert. I'm so glad divert works the way I want it to without much attention that I don't want to break it :-).

Most of my grief is with Stellantis, who's servers are very unreliable. Too many times Florian Bezannier's (@flobz) wonderful work with "PSA car controller" cannot do its job. I mainly use it to limit charge to 70% or 80% on my Peugeot e-208. As the spec for the type 2 charge cable does not allow direct comms on SoC, OpenEVSE does what it can and that too works well. I use the power limit.

I can forgive Stellantis, the Peugeot e-208 is a wonderful city EV, Stellantis nailed about everything else with it. We almost exclusively use it for our transport needs. So OpenEVSE is very important in our lives :-).

Not being able to set a battery charge limit native on the Peugeot e-208 with this battery chemistry is an important gap for me.
I do what I can and that is good enough, because this is a temporary problem anyway. The next generation EVs probably won't need that. LiFePO4 is already less sensitive and it looks like the sodium batteries that will come to market this year won't need as much babysitting either.

So I'm happy to see the OpenEVSE team putting its efforts where it matters for the future, integration in the power grid and trying to follow the erratic rule changes by the power grid operators. Way to go!!

KipK:
@pdhoogh I've updated the build with latest master & UI change: 

https://drive.google.com/open?id=1FjYBbFW2SXE_5jxBuc-knzxWo3vFwJwK&authuser=hotlobster75%40gmail.com&usp=drive_fs

pdhoogh:
You have values for 4 things now:
Divert default
Divert minimum waste
Divert peak import prevent
Shaper

So what would you like the values for?
20s attack and 300s decay look like a good set of values for "minimum waste" or the "default", With that we would need 600 secs minimum charging time.

Since I believe you provided "custom", the presets on the other choices may be fixed immutable values. Like the "Automatic" mode on a camera: the go to setting when you're lost.

KipK:
Changing presets should switch automatically to custom. Isn't it the case?

pdhoogh:
> Changing presets should switch automatically to custom. Isn't it the case?

I had not tried it. Indeed it does. Perfect!

KipK:
> > Changing presets should switch automatically to custom. Isn't it the case?
> 
> I had not tried it. Indeed it does. Perfect!

it pointed me to a little UI bug on those presets, it's fixed on the build in the [Drive](https://drive.google.com/open?id=1FjYBbFW2SXE_5jxBuc-knzxWo3vFwJwK&authuser=hotlobster75%40gmail.com&usp=drive_fs)

jeremypoulter:
@KipK I have fixed divert_sim to work with the InputFilter, see #626 

FYI here is this PRs results:

InputFilter_PR

vs master:

master

KipK:
here are the results for no-waste vs no-import presettings.

* no-import
 - settings:
  `{
  "divert_PV_ratio": 1.1,
  "divert_attack_smoothing_time": 300,
  "divert_decay_smoothing_time": 20,
  "divert_min_charge_time": 600
}
`
[inputfilter-noimport.pdf](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/files/11402355/inputfilter-noimport.pdf)


* no-waste
 - settings:
  `{
  "divert_PV_ratio": 1.1,
  "divert_attack_smoothing_time": 20,
  "divert_decay_smoothing_time": 300,
  "divert_min_charge_time": 600
}
`

[inputfilter-nowaste.pdf](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/files/11402360/inputfilter-nowaste.pdf)

KipK:
compared to current divert in master with ratio factors, default settings

[divert_master.pdf](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/files/11402413/divert_master.pdf)

KipK:
on what I see here is the default settings for no-waste are a little too conservative compared to old filter with default settings, 

for no-import it looks really good as it already

pdhoogh:
The time based filter constants can easily be calculated. What were the default filter factors of old?

KipK:
divert_attack_smoothing_factor: 0.4
divert_decay_smoothing_factor: 0.05
Min_charge_time 600

KipK:
so with 600s min_charge_time it gives this:
[inputfilter-nowaste2.pdf](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/files/11404169/inputfilter-nowaste2.pdf)

really near, I see it cutoff a bit more often than previous ratios in some case but nothing worriing

pdhoogh:
20 and 200 are probably good enough values, no need to be _that_ precise in my opinion.

KipK:
Those results looks quiet good. Seems the nowaste has a better cutoff than master and import a bit less from grid.

jeremypoulter:
But there are a lot more switches in the nowaste/default, I think the decay needs to be a little longer to bring it more inline with master, in terms of switching. I also wonder if no waste should also adjust the pv ratio to bring in more grid power and use more of the solar, so no wast will maximise the use of solar at the expense of using more grid, default should act similar to master and no import should sacrifice the use of solar so not to use any grid power ( think we also need to reduce/remove the min charge time as well for that config)

jeremypoulter:
well, maybe not remove, bt reduce. the current default is 5minutes IIRC. It would be great if you could supply CSV files of the data when there are errors so we can run it through the simulator and see if we get the same results.

KipK:
Can you explain the issue you have with timers and shaper using divert? For what I have tested here it was working ok

jeremypoulter:
> @pdhoogh Can you explain the issue you have with timers and shaper using divert? For what I have tested here it was working ok

Also, can we get some CSVs of the shaper feed? I would like to extend the simulator/testing to cover the shaper (and timer)

pdhoogh:
Not sure I am using the latest firmware you guys are testing. I have local_532c73e4_modified.

@KipK, what I noticed is the OpenEVSE started to swing between shaper limits and divert operation. Divert is supposed to keep import at 0W. I had set shaper limit to 800 Watt Import. The OpenEVSE was swinging the current between 0W and 800W import.

With timers, when I set them it works fine. When I remove them divert no longer works, I need to reset.

@jeremypoulter I would like to oblige, but I need to work to get a good datalogger set up here and I don't have it right now. Currently I am busy with other stuff so I'd like to do that later if that is fine with you. There is also the Error thing I need to investigate. But weather is good now and outside activities require my time: weeding, washing the paths, filling the gaps with polymer based sand, cleaning and restarting the swimming pool, changing the electricity distribution in the garden house, repairing broken lighting in the garden... gosh, I need to keep my dearest wife happy too :-). And as I said earlier, I'm quite happy for now where things stand with OpenEVSE. If there is a thing I can install really quickly on a Synology NAS using docker to log and create CSV files from MQTT data, I'd like to learn that. So far I have my sights on Grafana and that seems to be a fair bit of work. But maybe there is an easier way to get the CSV files you are asking for?

jeremypoulter:
> But maybe there is an easier way to get the CSV files you are asking for?

Should be fairly quick to setup with https://EmonCMS.org or https://data.openevse.com/emoncms/ if you don't mind posting data to a cloud service.

glynhudson:
If you want to use emoncms.org we can make a free account available for 1yr of data logging. If you want to go ahead with this Let me know your emoncms.org account username, and I'll apply some credit.

KipK:
>@KipK, what I noticed is the OpenEVSE started to swing between shaper limits and divert operation. Divert is supposed to keep 
> import at 0W. I had set shaper limit to 800 Watt Import. The OpenEVSE was swinging the current between 0W and 800W import.

That's odd oO. Shaper set only "max_current", divert set "charge_current" , they theorically can't fight together. We need to find why

pdhoogh:
@jeremypoulter and @glynhudson thank you so much for the very generous offer. My uid is "pdhoogh". Can you access my data directly from the cloud? Since I'm having a free trial, I'll have no problem at all with that. If needed I will curse in church (share my password). Of course not on the public forum. I will look into setting things up as soon as I get some more time.

@KipK, my observations were flaky at best. As I said, I need better data logging to make any claim. Also, what you suggest as a "malfunction" I would not have minded, really. But it is not necessary to have that "undocumented feature" (this is a joke from the process control world and is an alias for "bug") if the divert would work reliably with a "_Required PV power ratio_" < 1.0. 

It works very well with "_Required PV power ratio_" >= 1.0. But I never understood what it was trying to do with a value < 1.0. I tried 0.6 and I was expecting the charger in Divert mode would supply 6 Amps as soon as available current is 3.6 Amps. As more than 6 Amps is available, it would work as normal Divert (+I/-E = 0). And drop back to 6 Amps as available current would be between 3.6 and 6 Amps. But maybe that expectation is wrong. 

The described expected behavior is what I really want. It allows me to balance cost better when there is not enough solar. Instead of putting 800W in the grid and need to import 1320W afterwards to get the car charged, I'd rather use my 800W and add 520W import.

The observed behavior with "_Required PV power ratio_" = 0.6 is that it initially works as I expect, but when the available current goes above 6 Amps and later drops back to e.g. 4 Amps, the charger switches to 16 Amps (max current). That was very painful during my test, as the drop in available current was due to solar dropping off and coincidentally the swimming pool starting its cleaning cycle and gobble up 4600 Watts... the charger obliged with adding 3700 Watts, so I was importing 8000... OUCH! 

I have not tried it with the firmware I have now. The trial was too painful. But previous releases (4.1.x; x<=7) behaved that way.

KipK:
Ho there was so many changes since, I wouldn't keep those test on previous fw as factuals

pdhoogh:
Agreed as I think I understand what you mean. The test is still factual for 4.1.x, but should be retested on 4.2.2 or later, right?

I was not using Shaper during the test. Shaper was off. Just divert with Required PV Ratio =0.6. If it starts raining again here I will give it another go... now it is nice and sunny, no need.

src/divert.cpp
```diff
@@ -223,7 +223,7 @@ void DivertTask::update_state()
     double trigger_current = _evse->getMinCurrent() * min(1.0, divert_PV_ratio);
 
     // the smoothed current suffices to ensure a sufficient ratio of PV power
-    if (_smoothed_available_current >= trigger_current)
+    if (_smoothed_available_current >= trigger_current + EVSE_DIVERT_HYSTERESIS)
```
jeremypoulter, line 226: The `else` needs an occompianying `if` so there is a 'dead' zone between being active and disabled, eg

```C
    // the smoothed current suffices to ensure a sufficient ratio of PV power
    if (_smoothed_available_current >= trigger_current + EVSE_DIVERT_HYSTERESIS)
    {
      EvseProperties props(EvseState::Active);
      props.setChargeCurrent(_charge_rate);
      _evse->claim(EvseClient_OpenEVSE_Divert, EvseManager_Priority_Divert, props);
    }
    else if (_smoothed_available_current <= trigger_current - EVSE_DIVERT_HYSTERESIS)
    {
      if (EvseState::Active == _evse->getState(EvseClient_OpenEVSE_Divert) && 0 == min_charge_time_remaining)
      {
        EvseProperties props(EvseState::Disabled);
        _evse->claim(EvseClient_OpenEVSE_Divert, EvseManager_Priority_Default, props);
      }
    }
```
It is debatable if `EVSE_DIVERT_HYSTERESIS` should be applied to the upper level, lower level or both but the point of the hysteresis is that the oon value and off vale are not the same

src/app_config.cpp
```diff
@@ -177,12 +180,15 @@ ConfigOpt *opts[] =
 
 // Divert settings
   new ConfigOptDefenition<double>(divert_PV_ratio, 1.1, "divert_PV_ratio", "dpr"),
-  new ConfigOptDefenition<double>(divert_attack_smoothing_factor, 0.4, "divert_attack_smoothing_factor", "da"),
-  new ConfigOptDefenition<double>(divert_decay_smoothing_factor, 0.05, "divert_decay_smoothing_factor", "dd"),
-  new ConfigOptDefenition<uint32_t>(divert_min_charge_time, (10 * 60), "divert_min_charge_time", "dt"),
+  new ConfigOptDefenition<uint32_t>(divert_attack_smoothing_time, 20, "divert_attack_smoothing_time", "das"),
+  new ConfigOptDefenition<uint32_t>(divert_decay_smoothing_time, 200, "divert_decay_smoothing_time", "dds"),
```
jeremypoulter, line 184:
```suggestion
  new ConfigOptDefenition<uint32_t>(divert_decay_smoothing_time, 600, "divert_decay_smoothing_time", "dds"),
```
a value of 600 is much closer to the `master` defaults (at least for the data sets we have for the simulator) I would also change the `divert_PV_ratio` for the 'no waste` profile to 0.5, although I think we need to look at that value it looks to bedoing three different jobs...

src/divert.cpp
KipK: thanks, I've missed this one. Not sure we need to apply it when ramping down, or if we need to have it configurable yet. We'll see this with some more tests

src/app_config.cpp
KipK: nowaste profile is actually supposed to be the same as default config and match old default behavior. Use 0.5 pv ratio as default looks a bit aggressive :)

But I can add more presettings then. Should we start with default, no waste, no import ?

KipK: Here are the description I've put in UI for the "no waste" and "no import" presettings:
no waste: "No waste of produced energy. Slowly decrease charge rate, using grid to compensate, but increase faster when energy is going back."
no import: "Try to limit grid usage. Will slow down the charge rate quickly, but increase slower when energy is going back."

If we introduce a default one,  how would we describe it to differentiate it from no waste ?

KipK:
> IMHO I think these three functions should be split out into separate config values as I don't think they are always related

damn, we were trying to simplify the setup, and will end with more parameters than before :)

KipK:
Here are the presettings I propose then : 

```
presets = [
		{
			name: "Default",
			desc: "Slowly decrease charge rate, using grid to compensate, but increase faster when energy is going back.",
			id: 0,
			divert_attack_smoothing_time: 20,
			divert_decay_smoothing_time: 600,
			divert_min_charge_time: 600,
			divert_PV_ratio: 1.1
		},
		{
			name: "No waste",
			desc: "No waste of produced energy. Same as default but will import more from the grid to not waste any solar energy",
			id: 1,
			divert_attack_smoothing_time: 20,
			divert_decay_smoothing_time: 600,
			divert_min_charge_time: 600,
			divert_PV_ratio: 0.5
		},
		{
			name: "No import",
			desc: "Try to limit grid usage. Will slow down the charge rate quickly, but increase slower when energy is going back.",
			id: 2,
			divert_attack_smoothing_time: 300,
			divert_decay_smoothing_time: 20,
			divert_min_charge_time: 600,
			divert_PV_ratio: 1.1
		}
    ]
```

src/app_config.cpp
KipK: Here are the description I've put in UI for the "no waste" and "no import" presettings:
no waste: "No waste of produced energy. Slowly decrease charge rate, using grid to compensate, but increase faster when energy is going back."
no import: "Try to limit grid usage. Will slow down the charge rate quickly, but increase slower when energy is going back."

If we introduce a default one,  how would we describe it to differentiate it from no waste ?

pdhoogh:
I don't need more presets. The intention is now clear to me and I'm fine with it. But it has to work as intended. I believe you need to look at state space control, its transitions and what the charger does in each state.

I believe this may work (sorry, no graphcet :-)): 

State "Standby" >> charge is disabled or paused. After reset or intialization of the program, it begins in state "Standby"
Possible state transition is to "Initiate" only

State "Initiate" >> Charging at 6 Amps
Transition From "Standby" to state "Initiate" is when smoothed current reaches PV Ratio * 6 Amps 
Possible transitions are to "Standby" and "Divert"

Transition from "Initiate" to "Standby" when smoothed current < PV Ratio * 6 Amps * 0.9 (hysteresis 10%)
State  "Standby" discussed earlier

State "Divert" >> Divert charge control
Transition from "Initiate" to "Divert" when smoothed current > 6 Amps 
Possible transition only to state "Initiate"
Transition from "Divert" to "Initiate" when smoothed current < 6 Amps * 0.9 (hysteresis 10%)

This is a classic case where you get lost if not using state space control. This is old "serial" programming, the advantage is that it is more deterministic. Very difficult to predict behavior if this is deserialized...

I fully realize this is not the only sequence we need to have for the complete set of functions. This is the sequence for Divert operation only.

A similar sequence can be set up for the other functions such as Shaper.

Each sequence produces its own output current variable.

There is a "manager" sequence that then prioritizes and selects the charger state and current to use.

I know it is the old way, I have difficulties making sure what the behavior of a deserialized code is, too many possibilities.

jeremypoulter:
> Here are the presettings I propose then :

@KipK I think those will work. Can you update the expected test results then I think this is good to go.

KipK:
I'll try to find some time tomorrow to update those.

jeremypoulter:
Maybe we could write a script to generate those tests....

KipK:
> Maybe we could write a script to generate those tests....

yes if you can, I see there's a lot of data to update manually on this file.

KipK:
we're good to go.

jeremypoulter (dismissed):

jeremypoulter:
Good to go when the UI changes are merged

KipK:
> Good to go when the UI changes are merged

@jeremypoulter I've merged inputFilter changes in master gui, we're ready to go on this one. I'll rebase newflags thereafter.

jeremypoulter (approved):

jeremypoulter:
Thanks @KipK and @pdhoogh

Chaos-070799:
I need access to make and recieve phone calls to my boss and navigation for travel i will address other issues once i get home

PR #602, opened 2023-04-17
Add support for Wemos D1 Mini ESP32

mathieucarbou:
I can confirm that we can flash version 4.2.2 on the Wemos D1 Mini ESP32 with this profile and we can connect to the OpenEVSE card successfully. 
- TX pin of the ESP has to be connected to the RX pin of the OpenEVSE card
- RX pin of the ESP has to be connected to the TX pin of the OpenEVSE card

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -45,6 +45,7 @@ jobs:
       matrix:
         env:
           - openevse_nodemcu-32s
+          - openevse_wemos_d1_mini32
           - openevse_esp-wrover-kit
           - openevse_esp-wrover-kit_latest
           - openevse_huzzah32_dev
--- a/platformio.ini
+++ b/platformio.ini
@@ -139,6 +139,16 @@ build_flags =
   -D WIFI_LED_ON_STATE=LOW
   -D RAPI_PORT=Serial2
 
+[env:wemos_d1_mini32]
+board = wemos_d1_mini32
+build_flags =
+  ${common.build_flags}
+  ${common.src_build_flags}
+  ${common.debug_flags}
+  -D WIFI_LED=2
+  -D WIFI_LED_ON_STATE=LOW
+  -D RAPI_PORT=Serial2
+
 [env:openevse_esp-wrover-kit]
 board = esp-wrover-kit
 lib_deps =
--- a/platformio.ini
+++ b/platformio.ini
@@ -139,7 +139,7 @@ build_flags =
   -D WIFI_LED_ON_STATE=LOW
   -D RAPI_PORT=Serial2
 
-[env:wemos_d1_mini32]
+[env:openevse_wemos_d1_mini32]
 board = wemos_d1_mini32
 build_flags =
   ${common.build_flags}


KipK:
I guess you can already use 
env:openevse_nodemcu-32s
It use the same conf

Mathopenevse:
Hi, i confirm it's working for me with a D1 mini esp32 (cp 2104) and an openevse V.5.5 board 👍

jeremypoulter (commented):
Thank you for the contribution. Please also add the new env to the GitHub Workflow

zehn-cat:
This could be confusing to some people:
> 
> * TX pin of the ESP has to be connected to the RX pin of the OpenEVSE card
> * RX pin of the ESP has to be connected to the TX pin of the OpenEVSE card

The RX and TX pins for Serial2 are not the ones labeled RX and TX, but rather IO16 and IO17
I would suggest to use Serial0, then you can use the pins labeled RX and TX. However those pins are mislabeled on the Wemos D1 Mini ESP32 (at least the ones that I have- it's a known issue). The pins are swapped relative to the labels.  So need to connect "TX" to TX, etc.

Connected thusly, the pre-compiled builds for the Huzzah32 work fine for this board.
I use this same board so I also appreciate the effort to support it.

mathieucarbou:
> This could be confusing to some people:
> 
> > * TX pin of the ESP has to be connected to the RX pin of the OpenEVSE card
> > * RX pin of the ESP has to be connected to the TX pin of the OpenEVSE card
> 
> The RX and TX pins for Serial2 are not the ones labeled RX and TX, but rather IO16 and IO17 I would suggest to use Serial0, then you can use the pins labeled RX and TX. However those pins are mislabeled on the Wemos D1 Mini ESP32 (at least the ones that I have- it's a known issue). The pins are swapped relative to the labels. So need to connect "TX" to TX, etc.
> 
> Connected thusly, the pre-compiled builds for the Huzzah32 work fine for this board. I use this same board so I also appreciate the effort to support it.

Oh that's good to know and might explain why we were not able to make it work with the Huzzah32 build (we've tried).

mathieucarbou:
I am converting this to draft until I can do more testing.

jeremypoulter:
Most of the ESP32 boards with built in USB UART user Serial1 or Serial2 for the RAPI as Serial is connected to the USB and it is reall converient to use the USB for debug

mathieucarbou:
> Serial1

Yes that is why I created the `wemos_d1_mini32`  env like the `nodemcu-32s` one. Both have a usb uart.

But like @zehn-cat said about the tx-rx inversion, we found it strange because in some schematics rx goes to rx and in some photos rx goes to tx. Exemples:

- Doc here shoes rx - rx and tx - tx https://openevse.dozuki.com/Guide/Current+-+v5.5+OpenEVSE+Advanced+and+Value/26

- Image below (with the old esp8266) shows an inversion

image

I don't know if there is an officially documented way to wire. But if the official way to wire is to set rx - rx and tx - tx, and if what @zehn-cat said about the labelling being wrong, then the current PR makes sense.

But if the labelling is not inversed, then, maybe the rx and tx pins should be redefined and inversed in the env.

zehn-cat:
> Most of the ESP32 boards with built in USB UART user Serial1 or Serial2 for the RAPI as Serial is connected to the USB and it is reall converient to use the USB for debug

Yes, I understand that completely. Just though people might want to be able to use the released versions of compiled builds with this board - since there is not one that uses Serial2. I got really stuck on the pinout error. To make matters worse, there are two pins labeled "GND". Only the one closest to the pin "VCC" actually works.

jeremypoulter:
UART connection pinouts labels are allways subject to perspective, is it the TX/RX of the device or the expected input. In our case both boards the pinout is the TX/RX pins of the MCU so they need to be crossed over. 

Not sure about the Gnds which diagram are you referring to? 

There are several envs that use Serial2 and some that use Serial1 already so I don't have a problem with that, but what we probably do need is wiring diagrams for all the boards in the docs

mathieucarbou:
> There are several envs that use Serial2 and some that use Serial1 already so I don't have a problem with that, but what we probably do need is wiring diagrams for all the boards in the docs

I agree. Some wiring require rx-rx to match, some like d1 mini and esp8266 require rx-tx to be connected together.

The other alternative is to change the platformio file and redefine the RX & TX so that the doc can be consistant over all supported ESP in this file.

I thing @zehn-cat is referring to the 2 GND in this pinout diagram:

zehn-cat:
> I agree. Some wiring require rx-rx to match, some like d1 mini and esp8266 require rx-tx to be connected together.
> 
> The other alternative is to change the platformio file and redefine the RX & TX so that the doc can be consistant over all supported ESP in this file.
> 
> I thing @zehn-cat is referring to the 2 GND in this pinout diagram:

I would not suggest changing anything adapt to a board that has a design defect. You can see the error in the schematic relating to RX and TX here:
EpE4dGj
and noted here: https://github.com/MHEtLive/ESP32-MINI-KIT/issues/12

As for the GND pins, all I can say is that the one on the left side, closest to "RST", could not power the board all by itself- but there is no reason to power it that way

jeremypoulter:
As I said it is all about perspective, at the end of the day you have to connect the RX pin on one MCU to the TX pin on the other. The problem is that some people label the actual boards from the perspective of either the signal they are providing, or the signal they are expecting. I don't really think there is a correct or wrong answer in that context as the terms are ambiguous. What you really need is something like SPI with the MOSI and MISO labels which are a lot more explicit.

But to bring it back to this PR, I don't think any changes need making, but ideally it would be good to start on documenting the hardware connections, if a connection doc/diagram for the Wemos D1 Mini ESP32 could be added to the docs folder that would be great.

mathieucarbou:
> But to bring it back to this PR, I don't think any changes need making, but ideally it would be good to start on documenting the hardware connections, if a connection doc/diagram for the Wemos D1 Mini ESP32 could be added to the docs folder that would be great.

I will make that happen and squash the commits ;-)

mathieucarbou:
So after having tested with @Mathopenevse , D1 Mini ESP32 works fine as-is with Huzzah but rx and tx have to be inversed like @zehn-cat said. No need for a different env.

PR #614, opened 2023-05-01
Fix for working on different timezones

jeremypoulter:
maketime uses local time so changed to timegm so the input and output times are GMT

### DIFF
--- a/divert_sim/divert_sim.cpp
+++ b/divert_sim/divert_sim.cpp
@@ -70,7 +70,7 @@ time_t parse_date(const char *dateStr)
   time.tm_min = m;         // 0-59
   time.tm_sec = s;         // 0-61 (0-60 in C++11)
 
-  return mktime(&time);
+  return timegm(&time);
 }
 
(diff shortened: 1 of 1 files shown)


jeremypoulter:
@KipK Please take a look, should fix the time zone issues with the divert sim

KipK:
Date is good now, but I still get failed test:

```
01/01/2020 23:45:00,0,0,0,0,1440,254,0.0756085                | 01/01/2020 23:45:00,0,0,0,0,1440,254,0.0756085
Test Energy_and_Power_Day_2020-04-01 failed
Passed 0 tests, failed 12 tests
```

KipK:
using -Z ( ignbore trailing space ) on diff command solve the problem

jeremypoulter:
I wonder where the trailing spaces are coming from....

KipK:
when comparing files manually they don't have any trailing space, that's strange

jeremypoulter:
Yeah that could be it

Issue #512, opened 2023-01-09
Resolve issues waking some cars from sleep


KipK:
Some Zoe users needs to setup a special procedure to prevent the Zoe going in sleep state .
The /config set Sleep/Disable state for pause has been introduced for that.
But it doesn't seems to solve the issue on all Zoe.
They have a  start procedure on their smart home device to handle this:
- stop the car
- before it goes to sleep mode, plug the car and start a 30sec charge.
- set the EVSE to sleep ( 254 via $FS command)
This first procedure prevent the Zoe to go in sleep mode and not waking up later 
- when cheap rates  hours comes, put the EVSE to Disable ( 255 via $FD command) 
- wait few sec,
- Start charge ( $FE )

And then it seems to works.
The problem here is they still use RAPI commands for that, could be usefull to set an interface over the API to let them control that disabled state.

The /config one needs ESP32  to be restarted to let the disable state change effectively.

KipK:
Yep  have a way to dynamically set over the api the disable state from 254 to 255 or vis versa without having to reboot EVSE & write to the flash nor using RAPI commands.

mathieucarbou:
FYI I have a Zoe ZE50 E-Tech R110, and this vehicle is subject to the sleep issue.

Ref: https://github.com/OpenEVSE/ESP8266_WiFi_v2.x/issues/223
Ref: https://github.com/OpenEVSE/ESP32_WiFi_V4.x/issues/643

So I added an automation in HA which is sending through mqtt a sequence of `$FD $FE` when vehicle connects. I don't need to start a charge.

```yaml
automation:
  # 
  - alias: "OpenEVSE: Zoé Bugfix"
    trigger:
      - platform: state
        entity_id:
          - binary_sensor.openevse_vehicle_connected
        from: "off"
        to: "on"
    condition: []
    action:
      - service: mqtt.publish
        data:
          topic: "openevse/rapi/in/$FD"
      - delay:
          hours: 0
          minutes: 0
          seconds: 2
          milliseconds: 0
      - service: mqtt.publish
        data:
          topic: "openevse/rapi/in/$FE"
```

So far, and like reported in the issue in reference, this is pretty reliable.

I've seen this option in OpenEVSE:

![image](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/assets/61346/31cf6442-7437-4a11-b9ab-2ffedc714d43)

But I don't want to use it because of this warning:

[...]

PR #618, opened 2023-05-02
New settings / flags (#595)

### LINKED ISSUE #479: Restarting controller via HTTP throws error 500
Hi,

I'm on 4.1.6-dev and attempting to restart the controller via HTTP (using the UI or simply doing a curl to `https://openevse/r?json=1&rapi=$FR`) results in an error 500 with response `{"cmd":"$FR","error":"RAPI_RESPONSE_TIMEOUT"}`.

Interestingly the HTTP request is working because looking at the controller's terminal I see `$FR^30` followed by the asynchronous boot command `$AB`.

### LINKED ISSUE #595: Add some new /config flags & /status property
/config:
- wizard_passed 
  feature: set default to 0, when wizard has been passed once, set the flag to 1.
 comment: find a way to not display the wizard for upgraded setup. ssid is not sufficient as it doesn't solve ethernet module. Can we detect the whole config file is still default ? 

- default_state: 0 | 1 ( active | disabled )
 feature: allows to choose default evse state (active or disabled) when no override/claim has been setup.
comment: I think EVSE fw should also be set to disabled by default, not sure this is the case or if it has any collaterals.

- divert_type: -1 | 0 | 1 ( unset | solar | grid i/e )
 feature: store divert type. Remove needs of mqtt topic ( needed for http push ) ( -1 needed to import previous fw behavior ) 

- vehicle_data_src: -1 | 0 | 1 | 2 ( unset | tesla | mqtt | http)
 feature: store vehicle data source. remove  needs on checking tesla credentials or mqtt feed. 

/status:

~~- mode_auto: 0|1~~
  ~~feature: match the auto mode from UI, if there's no "state" property set in /override , then auto is 1~~
  ~~goal: remove some boilerplate code from UI side & simplify other integrations to match current UI2 use case.~~ 
  ~~comment: probably also need an endpoint on api to activate auto ( if called, it remove /override "state" property if there's other properties in the /override claim, or clear the override if there's only the "state" prop set. For now it's done on UI side. )~~

KipK:
FIxes #595
needs UI branch "newflags"

Not to be merged before UI migration

done: 
 * wizard_passed ( fw + ui )
 * divert_type ( fw + ui ) 
 * vehicle_data_src ( fw + ui )
 * default_state (fw + ui )
 * set AP ssid & password ( fw + ui ) 
 * added /restart {"device": "gateway|evse"} to http api & mqtt + restartEvse() method in evseManager ( close #479 ,


KipK:
@jeremypoulter , I have segfault on divert_sim here. I have modified this (https://github.com/OpenEVSE/ESP32_WiFi_V4.x/pull/618/commits/c3c68b790f756a51897acf177fd60f5497d55e41) in divert _sim. Any idea on what I've missed here ?

src/evse_monitor.cpp
```diff
@@ -254,8 +254,16 @@ void EvseMonitor::evseBoot(const char *firmware)
...
   });
 
+  // ESP32_WiFi handle OpenEvse state, so let first put it to disabled before unlocking
+  if (config_pause_uses_disabled()) {
+    disable();
+  }
+  else {
+    sleep();
+  }
+
```
jeremypoulter, line 264: This is the source of the crash in the divert_sim at least, not quite sure why it is not causing a problem on the ESP, but these calls cause a state change, thus calling updateState and ultimately the _energyMeter.increment_switch_counter. However the evseBoot call is before the _energyMeter.begin:
https://github.com/OpenEVSE/ESP32_WiFi_V4.x/blob/master/src/evse_monitor.cpp#L383-L386

KipK: Ah that makes sense, thanks.
So what would be the best way to prevent evse going to active state first  it hokups the switch at boot ( I can see when Bootlock is disabled the evse switch to active and immediately being set disabled by wifi module.

jeremypoulter: Can you move the energy monitor begin earlier? Maybe pass the EVSE monitor pointer in the constructor?


That being said won't the main loop put the EVSE into the correct state? Having these lines here will only change the state a few fractions of a second sooner right?

KipK: Yes evsemanager put it back to correct state, but at startup it will fast switch to active then disabled within a sec. 
Or perhaps we should set evse firmware to default to disabled ?

KipK: I've moved the EnergyMeter start a bit sooner, seems to fix it.

src/app_config.h
```diff
@@ -85,6 +88,14 @@ extern uint32_t scheduler_start_window;
 //Shaper settings
 extern uint32_t current_shaper_max_pwr;
 
+// Vehicle
+extern uint8_t vehicle_data_src;
+
+#define VEHICLE_DATA_SRC_NONE 0
+#define VEHICLE_DATA_SRC_TESLA 1
+#define VEHICLE_DATA_SRC_MQTT 2
+#define VEHICLE_DATA_SRC_HTTP 3
```
jeremypoulter, line 97: Can we use an enumeration instead of #define?

src/divert.h
```diff
@@ -5,6 +5,10 @@
 #ifndef _EMONESP_DIVERT_H
 #define _EMONESP_DIVERT_H
 
+#define _DIVERT_TYPE_UNSET -1
+#define _DIVERT_TYPE_SOLAR   0
+#define _DIVERT_TYPE_GRID  1
```
jeremypoulter, line 10: Please use an enum rather than #define, similar to [EventType](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/blob/master/src/event_log.h#L19)

src/evse_monitor.cpp
```diff
@@ -254,8 +254,16 @@ void EvseMonitor::evseBoot(const char *firmware)
     _heartbeat = RAPI_RESPONSE_OK == ret;
   });
 
+  // ESP32_WiFi handle OpenEvse state, so let first put it to disabled before unlocking
+  if (config_pause_uses_disabled()) {
```
jeremypoulter, line 258: Not convinsed this is the correct place for this, this is only useful if the target state is disabled, if the target state is active, this will cause an unnecessary switch to sleep and then back to active/charging. In any case the monitor should just get the current state and let the manager deside what to do. Also FYI the EVSE manager has a local copy of the config. Another thing to consider is that we are going to have to extend the 'pause' behaviour to handle some of the weird things various vehicles do, so having this logic in multiple places will make this harder.

KipK:
It won't switch, it 's locked in Bootlock mode, it prevent it to switch to
active and then switch back. But on the other side it will just switch to
active once

Le mar. 9 mai 2023, 19:23, Jeremy Poulter ***@***.***> a
écrit :

src/evse_monitor.cpp
jeremypoulter: Only the latest firmware has the boot lock so still have to support firmwares with out bootlock support. Probably what needs to happen is to unlock the bootlock in the main loop after the state has been evaluated and set the first time

KipK: I've changed the bootlock method with latest commit https://github.com/OpenEVSE/ESP32_WiFi_V4.x/pull/618/commits/c1fba7de64e362394f56d58d29ef3f65434ae389

KipK: So i've pushed to your openevse_lib the missing vflags : https://github.com/jeremypoulter/OpenEVSE_Lib/pull/9
Now with latest [commit ](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/pull/618/commits/e27c9a3f424bde4a75eca0e64ca8ca4a9c892123) it tests correctly the bootlock condition

KipK:
@jeremypoulter , when review is ok, don't merge it yet ( same for inputFilter too ). I'll have to merge the gui related branch to master first.

Can we decouple the daily builds from the gui commits now ?

KipK:
ah also can you bump a new openevse lib version ?

jeremypoulter:
> Can we decouple the daily builds from the gui commits now ? 

Just need to remove the workflow in the GUI repo

jeremypoulter:
> ah also can you bump a new openevse lib version ? 

Did a new release this morning

KipK:
> > Can we decouple the daily builds from the gui commits now ?
> 
> Just need to remove the workflow in the GUI repo

I've removed it and just put a gui build for testing.
This remind me to add some unit test on the gui later.

KipK:
> Yeah, testing would be excellent

Cool thing we can do with svelte is testing each components easily. 
https://sveltesociety.dev/recipes/testing-and-debugging/unit-testing-svelte-component

mathieucarbou:
@KipK : what about the ammeter calibration (`#SA 227 0` for me). Will there be a way to set this calibration from the UI if RAPI si removed ?

KipK:
@mathieucarbou , I have changed here the /restart to /restart {payload} , with payload = "gateway" or "evse"

mathieucarbou:
> @mathieucarbou , I have changed here the mqtt /restart to /restart {payload} , with payload = "gateway" or "evse"

oh, nice :-)

mathieucarbou:
> @mathieucarbou , I have changed here the mqtt /restart to /restart {payload} , with payload = "gateway" or "evse"

FYI I've [grabbed the restart changed in a branch](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/compare/master...mathieucarbou:ESP32_WiFi_V4.x:restart-evse) and I have updated my automation when Zoé vehicle is connected.

I will report back this week-end hopefully is the sun is helping.

src/divert.h
```diff
@@ -5,11 +5,18 @@
...
+
 #include <Arduino.h>
 #include <MicroTasks.h>
 
 #include "evse_man.h"
 
+enum divert_type {
+  _DIVERT_TYPE_UNSET = -1,
+  _DIVERT_TYPE_SOLAR = 0,
+  _DIVERT_TYPE_GRID = 1
```
jeremypoulter, line 17: Can we remove the `_` at the start? Still would like to use the more C++ style as well, IE `Divert type::Unset` but can do that at another point

src/web_server.cpp
```diff
@@ -1128,6 +1145,7 @@ web_server_setup() {
   server.on("/settime$", handleSetTime);
   server.on("/reset$", handleRst);
   server.on("/restart$", handleRestart);
+  server.on("/restartevse$", handleRestartEvse);
```
jeremypoulter, line 1148: Should mirror the MQTT API, but will need a little bit of backwards compatibility, so I would make a GET request to /restart restart the gateway but posting to /restart with either `gateway` or `evse` restart the appropriate device.... That being said that doesn't feel very RESTful.... I guess the RESTful thing to do would be to POST `restart` to /evse or /gateway   or maybe this is just one of those cases that doesn't fit into that model...

src/evse_man.h
```diff
@@ -449,6 +452,9 @@ class EvseManager : public MicroTasks::Task
     void enableTemperatureCheck(bool enabled, std::function<void(int ret)> callback = NULL) {
       _monitor.enableTemperatureCheck(enabled, callback);
     }
+    void restartEvse() {
```
jeremypoulter, line 455: @matth-x I think there was an outstanding requirement for restarting the EVSE for OCPP, does this meet that requirement?

src/web_server.cpp
KipK: would like but if we do the /gateway /evse way, then it would have sense to move all endpoints related to /gateway on it. I guess /restart gateway|evse follows more actual structure.  Shouldn't we post it as json also ? /restart {"dev": "gateway"}

src/divert.h
KipK: i've renamed them. I let you change those to C++ way or if you have an example

src/web_server.cpp
jeremypoulter: yeah i think probably at this point sticking to /reset is the best. Using a JSON payload is also fine, but should also have parity with the MQTT API

KipK: I've updated booth api & mqtt to /restart {"device": "gateway|evse"}

@mathieucarbou

matth-x:
@jeremypoulter Thanks for the notification! Will add this to the OCPP integration

mathieucarbou:
@KipK @jeremypoulter : do you guys plan on merging this PR soon ? 

If not, would it be possible to extract the restart work in another PR (would close ocpp requirement, fix bug, allow evse restart, would also provide a workaround for the zoe sleep bug, and I need that in my HA integration also)

Thanks :-)

api.yml
```diff
@@ -529,6 +530,28 @@ paths:
...
+  /restart:
+    post:
+      summary: restart gateway or evse modules
+        requestBody:
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                device:
```
mathieucarbou, line 542: `device` should be indented

KipK:
Would like to get this PR merged first, I have some work to rebase the inputFilter branchs on fw+ui, the merge of those 2 branchs won't be smooth I think.

jeremypoulter:
Think everything is done, will check over again then should be good to go

KipK:
@mathieucarbou I don't know wich fw version you have on booth evse + gw, but I've fixed yeserday an issue with bootlock on this branch, so if you merged this one before, there's chance the bootlock unlock method was failing at evse restart. 

If everything works as expected, your evse should never report wrong values fro min max current . Or there's something else happening you have to fix first, but rebooting each 2 hours the evse is clearly not a solution ;)

mathieucarbou:
@KipK : i actually updated my automation that was doing a sequence of FD FE for Zoé to use that considering the rapi support will be removed. What other solution otherwise? Using the disable sleep option has a warning saying the detection when the car is connected won't work anymore.

src/mqtt.cpp
```diff
@@ -95,7 +95,7 @@ void mqttmsg_callback(MongooseString topic, MongooseString payload) {
     DBUGF("voltage:%.1f", volts);
     evse.setVoltage(volts);
   }
-  else if (topic_string == mqtt_vehicle_soc)
+  else if (topic_string == mqtt_vehicle_soc && vehicle_data_src == VEHICLE_DATA_SRC_MQTT)
```
mathieucarbou, line 98: Note: when I install this branch + newflags UI branch together, vehicle data are not comping to mqtt because of this new check. And the UI does not send the `vehicle_data_src` setting in the post request with the mqtt config.
Hopefully I was able to send `{"vehicle_data_src":2}` to the mqtt topic `/config/set`. I'm glad to have it ;-)

KipK:
> @KipK : i actually updated my automation that was doing a sequence of FD FE for Zoé to use that considering the rapi support will be removed. What other solution otherwise? 

Still continue use RAPI until 5.0 release ?
You are posting issues on dev branches we haven't even merged yet, and things we're still working on, hold-on a bit ;)

src/mqtt.cpp
mathieucarbou: Yes I know, that's where I was, in the vehicle page. I already had my mqtt config in place, the drop-down was on mqtt mode, but the mqtt topics for the vehicle data were not updated because the src variable was not set to mqtt mode. And saving again from the UI did not help: when introspecting the Jason payload sent, I did not see any src variable sent in the payload to tell the gateway we are in mqtt mode. That is why I was saying that updating the config fixed that (it is easier for me to update the config through mqtt)

KipK: seems it has failed with your merges. Things are working well here

KipK: looks you have wrong UI. We hve remove ui v1 so you don't need the env variables anymore, now the gui is inside the main tree

mathieucarbou: I have master with gui 2 and switched the submodule to the newflags branch (your Branch) for both UI and gateway.
Sorry I might have done something wrong.
These submodules switches add more complexity.
I'm also wondering if a switch and rebuild of another gui branch is detected by the firmware rebuild

KipK: Nope you  need to rebuid the UI if there's changes

jeremypoulter:
> Changing the pause state to disable should resolve this issue. However this removes the ability for the charger to detect if a vehicle is connected when paused.

@mathieucarbou You are doing the same thing as enabling that option so the warning applies weither you enable that option or not, but lets keep this discussion in #512

jeremypoulter:
Need to also remove the replaced web files. IMHO I see little value in appending the file hash to the file name. The server should properly implement HTTP caching, so it should not be needed.

KipK:
> Need to also remove the replaced web files. IMHO I see little value in appending the file hash to the file name. The server should properly implement HTTP caching, so it should not be needed.

I think it's part of Svelte/Vite cache invalidation handler. Not sure it has consequencies to disable it, haven't tested yet.

jeremypoulter:
Fine to leave it on I guess, but do then need to make sure all the headers are deleted before bringing in a new UI version. Really need to write a GitHub workflow to do that...

KipK:
> Really need to write a GitHub workflow to do that...

jeremypoulter:
We commit the static files so you don't have to build the UI to build the firmware.

KipK:
> We commit the static files so you don't have to build the UI to build the firmware.

ok, i'll update them here after the inputfilter merge.

KipK:
UI changes merged in master, we're good on this one too.

jeremypoulter:
Just the API doc errors to resolve

jeremypoulter:
Oh ignore that, just noticed you added a change, let me check

KipK:
> Oh ignore that, just noticed you added a change, let me check

yes should be ok, I've installed the vscode extension and spotted the errors

jeremypoulter (approved):

PR #626, opened 2023-05-04
Changes to update millis based on the time simulated time

jeremypoulter:
Needed for the new InputFilter in #590


KipK:
Ho that was fast 😊

KipK (approved):

PR #627, opened 2023-05-06
Divert_sim summary

jeremypoulter:
Moved to python for running the simulations and generating a summary of the results


KipK (approved):

jeremypoulter:
@KipK Think this should work for the InputFilter branch

KipK (approved):

KipK:
I've updated the InputFilter (shaper) branch too

PR #633, opened 2023-05-09
add missing max_current in model and api.yml

KipK:


mathieucarbou:
Sorry, I missed that (web-related) part...

PR #634, opened 2023-05-09
Expose config and config updates through MQTT

mathieucarbou:


mathieucarbou:
@KipK : I'ce created this PR to monitor config and config changes. This is useful to display dahsboard components accordingly on HA, especially to get some config like power ratio, max_current_soft, etc.

Note: this introduces a code duplication that needs to be solved - just please let me know here would be the good candidate for putting the common code between web and mqtt to generate the json payload.

thanks ;-)

mathieucarbou:
Note: when a RAPI command is sent through MQTT or debug mode (ie. `$SA 190 0`), the config does not update. 

Which is OK IMO

src/web_server_config.cpp
```diff
@@ -240,6 +240,8 @@ bool web_server_config_deserialise(DynamicJsonDocument &doc, bool factory)
 
     StaticJsonDocument<128> event;
     event["config_version"] = config_version;
+    event["config_updates"] = doc;
```
KipK, line 243: not sure that's intended. Should be removed here
you just need to use "config_version" to track config changes

mathieucarbou: So it was intended at the beginning because I didn't find a way to send the full config each time. But I found it and not use that anymore. I can remove it yes ;-)

src/mqtt.cpp
```diff
@@ -489,6 +493,53 @@ mqtt_publish_schedule() {
...
+  doc["gfci_check"] = evse.isGfiTestEnabled();
+  doc["ground_check"] = evse.isGroundCheckEnabled();
+  doc["relay_check"] = evse.isStuckRelayCheckEnabled();
+  doc["vent_check"] = evse.isVentRequiredEnabled();
+  doc["temp_check"] = evse.isTemperatureCheckEnabled();
+  doc["service"] = static_cast<uint8_t>(evse.getServiceLevel());
+  doc["scale"] = evse.getCurrentSensorScale();
+  doc["offset"] = evse.getCurrentSensorOffset();
+  doc["max_current_soft"] = evse.getMaxConfiguredCurrent();
+
```
jeremypoulter, line 533: I think all this can be moved to config_serialize (and the equivalant for config_deserialze) this will then be shared between HTTP and MQTT

mathieucarbou: Yes that is exactly what I thought too, this would be a good candidate. 
I hesitated because some comments in the code were indicating that `config_deserialze` is dedicated to the gateway config.  Moving this code inside would enlarge this file to config handling both for gateway and evse then.

I will update!

src/app_config.h
```diff
@@ -223,7 +223,6 @@ bool config_deserialize(DynamicJsonDocument &doc);
 void config_commit(bool factory = false);
 
 // Write config settings to JSON object
-bool config_serialize(String& json, bool longNames = true, bool compactOutput = false, bool hideSecrets = false);
```
mathieucarbou, line 226: seemed unused

src/app_config.cpp
```diff
@@ -267,6 +269,44 @@ config_load_settings()
   }
 }
 
+void append_common_config(DynamicJsonDocument &doc) {
+  doc["config_version"] = config_version;
```
mathieucarbou, line 273: moved `config_version` into version object for practical reasons when reading the mqtt topic and also to ensure the attributes of the mqtt mapping in HA gets updated correctly when we use one sensor for all attributes.

src/mqtt.cpp
```diff
@@ -395,7 +396,11 @@ mqtt_publish(JsonDocument &data) {
     String topic = mqtt_topic + "/";
     topic += kv.key().c_str();
     String val = kv.value().as<String>();
-    mqttclient.publish(topic, val, config_mqtt_retained());
+    if("config_version" == kv.key().c_str()) {
+      mqtt_publish_config();
+    } else {
+      mqttclient.publish(topic, val, config_mqtt_retained());
+    }
```
mathieucarbou, line 403: correctly handle mqtt updates when config changed from web

src/mqtt.cpp
```diff
@@ -489,6 +494,22 @@ mqtt_publish_schedule() {
...
+  }
+  const size_t capacity = JSON_OBJECT_SIZE(128) + 1024;
+  DynamicJsonDocument doc(capacity);
+
+  config_serialize(doc, true, false, true);
+  mqtt_publish_json(doc, "/config");
+
+  String fulltopic = mqtt_topic + "/config_version";
+  String config_version = doc["config_version"];
+  mqttclient.publish(fulltopic, config_version, true);
```
mathieucarbou, line 510: retain flag required for this one

src/mqtt.cpp
mathieucarbou: Finsihed. Tested and works fine. I went a little further to also refactor the TODO part that was left ;-)

mathieucarbou: I am now fighting against the build of the divert simulator sub-project 😁

mathieucarbou: ... And I won 😀

mathieucarbou:
PR rebased and updated to push the config each 30sec with the new time

src/mqtt.cpp
```diff
@@ -397,6 +398,9 @@ mqtt_publish(JsonDocument &data) {
     String val = kv.value().as<String>();
     mqttclient.publish(topic, val, config_mqtt_retained());
     topic = mqtt_topic + "/";
+    if("time" == kv.key().c_str() || "config_version" == kv.key().c_str()) {
```
KipK, line 401: This still don't have to be in mqtt_publish() function.

src/app_config.cpp
KipK: config_version is already published in /status ( like all xxx_version props )  , or in main mqtt topic ( it's not part of the /config as it's not saved in the flash and is just here to track live updates. I don't think it should stays in the published config.

KipK:
> PR rebased and updated to push the config each 30sec with the new time

config should only be pushed at mqtt connection with retain flag, and thereafter only when there is an update ( i.e when config_version has changed ) . No need to push this each 30sec, just  push it after a config change.

mathieucarbou:
> h 30sec, just push it aft

that is what I did initially. But once mqtt connects, I guess the evse manager is not ready yet ? Because all properties returnes from evse like firmware, max_current_soft, offset, etc are all equals to empty strings or 0.

So I did that to get a second push to ensue the config gets sent at each time update

KipK:
> > h 30sec, just push it aft
> 
> that is what I did initially. But once mqtt connects, I guess the evse manager is not ready yet ? Because all properties returnes from evse like firmware, max_current_soft, offset, etc are all equals to empty strings or 0.
> 
> So I did that to get a second push to ensue the config gets sent at each time update

so we need to find a way to wait for data to be ready first . But publish it only once, not spamming mqtt server with same data each 30sec.

src/app_config.cpp
KipK: I don't know HASS mqtt implementation, but it should clearly parse the Json when the topic is updated. Why does it has to check a specific property ?

mathieucarbou: It parses it. When a mqtt update happens, a component can be bounded to a topic, a apecific field inside, and all other fields are attributes of the same sensor.

So in HA we can have only 1 sensor listener to 1 mqqtt topic and it will have its attributes taken from the payload.

But still ,there are needs to have at least one field representing this sensor. In the case of the config, this is the config version.

Each time the config version is changing, then the config attributes change also.

It is possible to have another topic defined for the attributes but it would need to be published BEFORE config_version is published. Let me try...

mathieucarbou: I've also changed the ordering of the calls to `append_common_config(doc);`

src/mqtt.cpp
```diff
@@ -33,6 +33,7 @@ static bool connecting = false;
 uint8_t claimsVersion = 0;
 uint8_t overrideVersion = 0;
 uint8_t scheduleVersion = 0;
+uint32_t config_version_mqtt = 1;
```
KipK, line 36: can you use the same variable format  as for claimsVersion, overrideVersion, ... above ? 
=> configVersion = 0;

src/mqtt.cpp
```diff
@@ -289,6 +290,7 @@ mqtt_connect()
     event_send(doc);
 
     // Publish MQTT override/claim
+    mqtt_publish_config();
```
KipK, line 293: you should then not publish it at connect and wait for the loop to compare the config_version, as when the config will be loaded, it will increment it so this shouldn't be long to wait for ;)
And I bet the data will be ready then

src/web_server_config.cpp
```diff
@@ -91,7 +55,7 @@ handleConfigPost(MongooseHttpServerRequest *request, MongooseHttpServerResponseS
     bool config_modified = web_server_config_deserialise(doc, storage.equals("factory"));
 
     StaticJsonDocument<128> reply;
-    reply["config_version"] = config_version;
+    reply["config_version"] = doc["config_version"];
```
KipK, line 58: there's no config_version in /config and so no config_version should then be posted too in handleConfigPost().
doc["config_version"] won't exist, jsut keep it like it was:
reply["config_version"] = config_version;

KipK:
@mathieucarbou  so I've added #637 for the missing /limit /limit/set

src/web_server_config.cpp
mathieucarbou: So I did that initially because `config_version ` was internal and not exposed to I needed to pass it through the document.
Now that `config_version`  is exposed I can clean yes.

src/mqtt.cpp
mathieucarbou: Ok so I confirm the behaviour. There is no update and the `mqtt_loop` is not working. So I have updated the code also for the schedule, claims, override and config. Because if your comments holds true for the config, it should also hold true for the rest right ? So no need to publish them once connected since they should be published in the loop. And to make sure of that, I've changed them to `-1` because the initial value of version is `0`.

mathieucarbou: image

So I will leave the config publish once connected, like it is done for the claims, override and schedule...

KipK: it was to solve the fact config was not ready when you first publish it. 
If it's not incremented, let's just do it at config load.

mathieucarbou: > it was to solve the fact config data were not totally ready when you first publish it.
> If it's not incremented yet, let's just do it at config load, and don't publish when version is still 0

But `config_version` is ONLY incremented after a config change happens from the UI.

So I don't get the _don't publish when version is still 0_... Version WILL be 0.

KipK: I've wrote: 
>If it's not incremented yet, let's just do it at config load

was meant to add an increment at config load when the wifi fw initialise. I know it is not yet

mathieucarbou: The problem I saw with the EVSE data not being present in the config json was when I flipped the electtic breaker to shutdown and restart my electrical box.

When doing that, both the evse card + the gateway were rebooted. And at this point, the publication of the config does not have the evse data.

This is not happening when we just restart of flash the gateway.

KipK: but I think just do in the mqtt loop() the first /config publish and not in begin() should solve the uninitalised values published.
use -1 as default configVersion, check if -1 in the loop, publish config then set version to 0.

mathieucarbou: > I think you can just do in the mqtt loop() the first /config publish and not in begin() should solve the uninitalised values published.
> use -1 as default configVersion, check if -1 in the loop, publish config then set configVersion to 0, it will match config_version then

tried that in the loop:

```json
    if(configVersion == -1) {
      mqtt_publish_config();
      configVersion = 0;
    }
```

but this is not working at all... Nothing gets published.

mathieucarbou: So either we keep it as-is (config exposed once mqtt is connected) and if the box gets some power and evse is not yet started, then we have to restart the gateway, or we have to call `mqtt_publish_config` from where the evse gets ready to be sure...

KipK: you better find why it's not publishing in your test above than find a hack to solve the issue. ;)

mathieucarbou: > 

I've not a C guy... What options do I have to debug that ? I've tried the debugvar and some logging but do not see anything, which suggests the loop is not running maybe ? Or could it be the assignment of -1 to an unsigned int that is preventing the comparison ?

Also, what is ensuring that the evse will be ready at the first run of the loop ?

Isn't it more reliable to trigger the config publish once the evse is ready since the config system now depends on teh evse settings to be available ?

mathieucarbou: @KipK : I've implemented the listener :

https://github.com/OpenEVSE/ESP32_WiFi_V4.x/compare/29c5a9d772dd082969269594c85e95228bacf04a..f6c5d54ba5a81471912512069357ab22a7f4bf79

So with all that it works perfectly:

- if mqtt connects before evse boots, the event sent from evse manager will take care of sending the right config update just after the previous one
- if evse boots before mqtt connects, the initial send of the config from mqtt will take care of the update (the one coming from evse won't happen because mqtt not connected)

I tested a complete reload of the box (electricity off/on) - it works.

Today i've flashed the openevse firmware 3 times to test some flags and each time I was turning power on, the config was badly published - with all evse settings missing (when published after mqtt connects because evse was not yet available). 

So this is really a corner case  that we are trying to handle here...

KipK: I was thinking for another PR to add a variable for this, could be usefull yes.
But I don't think there's need to publish it, it's used for internal functions only for a fraction of a second. 
Perhaps an isEvseReady() function cascading in evseManager, then evseMonitor ( with a private _evse_ready flag set to true at the end of the evse ready callback ), should be more consistant with the codebase.

KipK: but mostly if you check the evse state to be something else than OPENEVSE_STATE_STARTING, means it is ready

mathieucarbou: OK! So I in summary I will add a boolean flag, set to true at `OPENEVSE_STATE_STARTING`, and I will expose it through the different apis, and access its getter in the new `mqtt_publush_config` method, to only expose if evse has started.
- I can keep the publish call at mqtt connection time, like it is done for the claims, override, schedule (because most of the time evse should be started at that point)
- I will track in the loop this initial publish if it has to be done also if the initial mqtt_publich_config call was not able to do it

mathieucarbou: @KipK :  I think we overthink it ;-)

1. We just have to expose a flag in the evse to know if is has started or not.
2. In app config, when we add the evse config, we check this flag and if evse didn't start yet, we add instead `evse_starting = true`. If evse has started, we add the settings but not the boolean.

So this will work because any client receiving the partial config will be able to know if it is complete or not.

Besides, Home Assistant has a way to tell whether a component is available or not by looking at the content of the payload. So if I see `evse_starting = true`, then I can tell HA that the sensor is not yet available.

in the mqtt:
- we punlish at startup
-in the loop, we then also check this status and we publish again once evse is started. Then after only on config change,

KipK: We're talking about evse data from the module that are not part of the real config anyway and are not settable through this endpoint.   The real config is always ready anyway.  On UI http side it grabs the config after receiving /status and other endpoints so all data are always ready at this stage.  (  the time it takes to refresh the UI the evse has already booted also ).

But I guess we could introduce a config_version change when those evse data are ready at the end of the evse ready callback.? 

@jeremypoulter how would you handle this ?

mathieucarbou: Agree: Introducing a `config_version ` change so that the loop publishes would solve it yes, and remove the need to publish an eventually partial config when mqtt connects/

But can (should we) access this `config_version` variable from the evse ready clalback ?

mathieucarbou: @KipK I have implemented that and it works perfectly, config is incremented then pushed in mqtt
image

mathieucarbou: Ahah! So after testing, it works perfectly when the whole box is powered downed and up again. 
But if the gateway is restarted alone and the mqtt broker does not have the config topic, it won't be created because evse has already booted. So we need also need to publish at mqtt connection.

mathieucarbou: PR rebased and updated.

KipK: > downed and up again.
> But if the gateway is restarted alone and the mqtt broker does not have the config topic, it won't be created because evse has already booted. So we need also need to publish at mqtt connection.

if the evse is restarting alone, then it will call increment_config_version() from the EvseMonitor::evseBoot() , so you should get the config_version change in the mqtt loop too and publish it as we said above.

KipK: Perhaps you can only publish at mqtt connection if the  evse state is != from OPENEVSE_STATE_STARTING  ?

mathieucarbou: > Perhaps you can only publish at mqtt connection if the evse state is != from OPENEVSE_STATE_STARTING ?

For the rare case when it will happen... Is it worth coupling the mqtt layer with the evse state like that ?

mathieucarbou: > Mqtt already use EvseManager, you can call it to get the current state with evse getEvseState()

Oh right I didn't notice the transitive includes....

So I've updated like that instead:

```c
void
mqtt_publish_config() {
  if(!config_mqtt_enabled() || !mqttclient.connected() || evse.getEvseState() == OPENEVSE_STATE_STARTING) {
    return;
  }
```

KipK:
> Note: when a RAPI command is sent through MQTT or debug mode (ie. `$SA 190 0`), the config does not update.
> 
> Which is OK IMO

RAPI will probably be removed soon from MQTT now it is dedicated to wifi gateway and API has to be used instead.

KipK (commented):
Looks all ok to me, thanks for this. I have not tested yet, will do soon 

@jeremypoulter I let you give the final check. I'll rebase #637 after this PR has been merged.

mathieucarbou:
> > Note: when a RAPI command is sent through MQTT or debug mode (ie. `$SA 190 0`), the config does not update.
> > Which is OK IMO
> 
> RAPI will probably be removed soon from MQTT now it is dedicated to wifi gateway and API has to be used instead.

That's a bummer :-(

How to solve this issue then ?

src/evse_monitor.cpp
```diff
@@ -266,6 +267,8 @@ void EvseMonitor::evseBoot(const char *firmware)
     }
 
   });
+
+  incrementConfig();
```
jeremypoulter, line 271: The config should listen in to the evse.onBootReady event so there is not any circular dependancy

src/app_config.cpp
```diff
@@ -232,6 +234,23 @@ ConfigOpt *opts[] =
 ConfigJson user_config(opts, sizeof(opts) / sizeof(opts[0]), EEPROM_SIZE, CONFIG_OFFSET);
 ConfigJson factory_config(opts, sizeof(opts) / sizeof(opts[0]), EEPROM_SIZE, FACTORY_OFFSET);
 
+uint32_t
+getConfigVersion() {
```
jeremypoulter, line 238: Should use the same snake case style as the other functions
```suggestion
config_version() {
```

src/app_config.cpp
```diff
@@ -232,6 +234,23 @@ ConfigOpt *opts[] =
...
+  return config_version;
+}
+
+void 
+incrementConfig() {
```
jeremypoulter, line 243: Should be intenal and also use snake case as the other functions

src/app_config.cpp
```diff
@@ -349,7 +368,117 @@ bool config_deserialize(const char *json)
...
+  {
+    long current = doc["max_current_soft"];
+    if(current != evse.getMaxConfiguredCurrent()) {
+      evse.setMaxConfiguredCurrent(current);
+      config_modified = true;
+      DBUGLN("max_current_soft changed");
+    }
+  }
+
+  if(doc.containsKey("scale") && doc.containsKey("offset"))
```
jeremypoulter, line 456: @mathieucarbou related to other comments about RAPI being removed, the current calibration can be set using the scale and offset

jeremypoulter (commented):
Looking good, just a few small changes

src/app_config.cpp
```diff
@@ -349,7 +368,117 @@ bool config_deserialize(const char *json)
 
 bool config_deserialize(DynamicJsonDocument &doc)
 {
-  return user_config.deserialize(doc);
+  bool config_modified = user_config.deserialize(doc);
+
+  #if ENABLE_CONFIG_CHANGE_NOTIFICATION
```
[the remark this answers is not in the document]
mathieucarbou: yes ;-)

src/evse_monitor.cpp
mathieucarbou: I'm a little lost with this listener logic: where can I place the call exactly ?

src/app_config.cpp
mathieucarbou: but increment is called from elsewhere (monitor). How to make it internal ?

src/evse_monitor.cpp
jeremypoulter: Have A look at [input.cpp](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/blob/master/src/input.cpp#L27)

src/app_config.cpp
jeremypoulter: If you switch to using the boot event it can be kept internal.

src/evse_monitor.cpp
jeremypoulter: Also FYI all the above calls are asyncronous so may not have cmopleted at this point. The boot event won't suffer from that at it is only triggered after all calls have been completed

mathieucarbou: Ok I understand! I'll need to create a micro task in the app config!

src/app_config.cpp
mathieucarbou: I am resolving it as resolving the other one will also resolve it.

src/evse_monitor.cpp
mathieucarbou: @KipK @jeremypoulter : I re-thought it again and we actually don't need to listen on anything.

I am initialising the config version to 1 internally, and 0 in the mqtt.

in mqtt, I am cheking: `evse.getEvseState() == OPENEVSE_STATE_STARTING`. The method `mqtt_publish_config` returns true if it wa sable to publish.

Then, in the mqtt loop:

```c
    if(configVersion != config_version() && mqtt_publish_config()) {
      DBUGF("Config has changed, publishing to MQTT");
      configVersion = config_version();
    }
```

So after a firmware flash or complete power loss / back up again, the loop will keep trying to republish the config until mqtt is connected and evse is started.

LavredisG:
I've read that RAPI will soon be removed (probably) from the MQTT and I have some questions on it:

1) Does this affect working with HTTP too, or in this case only http is used as api?
2) Is there any difference in general while working with either HTTP or MQTT?

jeremypoulter:
Yeah the intention has always been to remove RAPI. I have created a ticket for this, will probably be part of the upcoming v5 release: #643

Let's move any discussion about removing RAPI there so we didn't lose any comments

mathieucarbou:
Rebased!

glynhudson:
@mathieucarbou could you add the new MQTT API to the docs as part of this PR? Thanks 

Your HA integration looks great, have you published this?

src/mqtt.cpp
```diff
@@ -505,6 +507,23 @@ mqtt_publish_schedule() {
...
+    return false;
+  }
+  const size_t capacity = JSON_OBJECT_SIZE(128) + 1024;
+  DynamicJsonDocument doc(capacity);
+  config_serialize(doc, true, false, true);
+  mqtt_publish_json(doc, "/config");
+
+  String fulltopic = mqtt_topic + "/config_version";
+  String payload = String(config_version());
+  mqttclient.publish(fulltopic, payload, true);
```
jeremypoulter, line 522: I don't think you need to event `config_version`, this is already evented via event_send when the version is incremented.

jeremypoulter, line 522: You don't need this, `config_version` is evented via event_send when the version number is incremented

mathieucarbou: > You don't need this, `config_version` is evented via event_send when the version number is incremented 

We need it when the first publication comes from the mqtt layer. Version is not upgraded at startup. I initially didn't have it but if we remove it, only the config is pushed not the version,  and it takes a config change coming from the UI to get it published.

mathieucarbou:
rebased on master just now

src/mqtt.cpp
jeremypoulter: But with it here it is published twice. It may be better to handle explicitly in the MQTT connection code. I would also assume the same bug exists for all the xxx_version, that being said are these useful for MQTT if you are exposing the resources directly? They are meant so the version number is evented over MQTT or Websocket, and then you would fetch the resource over HTTP. With the config, etc being sent over MQTT directly the version is irrelevant.

mathieucarbou: > But with it here it is published twice. 

Yes I knew that...

> It may be better to handle explicitly in the MQTT connection code.

Thought about that and no because the version number and config might not be published at mqtt connection time if evse is not started. So the initial publish will come from the loop.

>  I would also assume the same bug exists for all the xxx_version, that being said are these useful for MQTT if you are exposing the resources directly? 

Not at all... There is no point in having have these *_version exposed in mqtt (IMO). Only exception is if someone wants to know hen a config change happened...

> They are meant so the version number is evented over MQTT or Websocket, and then you would fetch the resource over HTTP. With the config, etc being sent over MQTT directly the version is irrelevant.

I agree.

So to keep the behaviour backward compatible but still avoid this duplicate publish, what I can do is check the version number is the mqtt publish function. If it is 1, it means it has to be published because this is the initial publish. If it is greater than 1, then there has been a config change, and the event will be responsible to publish.

mathieucarbou: I've just pushed https://github.com/OpenEVSE/ESP32_WiFi_V4.x/pull/634/commits/ad9524659cd410a822f7c67c872acce02db83ee2 as a proposal to remove the duplication

src/mqtt.cpp
```diff
@@ -507,6 +509,25 @@ mqtt_publish_schedule() {
...
+mqtt_publish_config() {
+  if(!config_mqtt_enabled() || !mqttclient.connected() || evse.getEvseState() == OPENEVSE_STATE_STARTING) {
+    return false;
+  }
+  const size_t capacity = JSON_OBJECT_SIZE(128) + 1024;
+  DynamicJsonDocument doc(capacity);
+  config_serialize(doc, true, false, true);
+  mqtt_publish_json(doc, "/config");
+
+  if(config_version() == initial_config_version()) {
```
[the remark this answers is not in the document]
mathieucarbou: The constant was hidden in the cpp file, that's why. I had to expose it to do that.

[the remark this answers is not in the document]
jeremypoulter: Hum I guess, but why can't this be moved to the MQTT connection function?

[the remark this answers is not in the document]
mathieucarbou: I think the responsibility to define the initial value for the version_number is up to the config class, which also makes use of the constant to initialise the variable.

It cannot be moved to the connection function because if the mqtt layer is connected before evse is started, the mqtt_publish_config will return because of `evse.getEvseState() == OPENEVSE_STATE_STARTING`. So the initial publication of the config_version will be done from the mqtt loop, not at connection time, when evse will have started.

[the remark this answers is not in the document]
mathieucarbou: cases:

1. evse == `OPENEVSE_STATE_STARTING` + mqtt connected => `mqtt_publish_config` returns false
2. evse != `OPENEVSE_STATE_STARTING` + mqtt connected => `mqtt_publish_config` publishes both `/config` and `/config_version` (initial publish for first version)

Then, MQTT loop

1. `localVersion` != `config_version` +  evse == `OPENEVSE_STATE_STARTING` => `mqtt_publish_config` returns false
2. `localVersion` != `config_version` +  evse != `OPENEVSE_STATE_STARTING`
   2.1. publish `/config`
   2.2. fF `config_version` == `INITIAL_CONFIG_VERSION`,  publish `/config_version`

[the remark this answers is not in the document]
jeremypoulter: Ok, guess we will leave it for now, just a lot of code to manage an irrelevant topic....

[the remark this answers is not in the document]
mathieucarbou: I agree. But I also think that the code is now better than before and also answers [your request](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/pull/634#discussion_r1188926242) to centralise things.  Making the app config class now also dependant on the evse settings now require to think about what could be the evse state when we access those settings. In most of the app, we do not care. But at startup, we care.

jeremypoulter (approved):

PR #659, opened 2023-05-25
delete static files if there's a gui build

KipK:
clean-up web_server_static_files.h when building firmware and the ui build dir is not empty

### DIFF
--- a/src/net_manager.cpp
+++ b/src/net_manager.cpp
@@ -99,12 +99,12 @@ void NetManagerTask::wifiStartAccessPoint()
 
   String softAP_ssid;
 
-  if (!ap_ssid) {
-     // Create Unique SSID e.g "emonESP_XXXXXX"
+  if (ap_ssid.length() < 2) {
+     // Create Unique SSID e.g "OpenEVSE_XXXXXX"
     softAP_ssid = String(_softAP_ssid) + "_" + ESPAL.getShortId();
   }
   else {
-    softAP_ssid = ap_pass;
+    softAP_ssid = ap_ssid;
   }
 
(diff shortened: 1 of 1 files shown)


jeremypoulter:
Why is this needed? `web_server_static_files.h` is regenerated on build, is it not?

KipK:
not really, the #includes where appened to the file, so we have to clean it manually first before commiting or we will have thousands of includes after few gui revision ( because of the hash in the file names  )

jeremypoulter:
No I just checked, `web_server_static_files.h` is deffinatly generated from scratch when needed, <https://github.com/OpenEVSE/ESP32_WiFi_V4.x/blob/638dc90257b477966b6dd6305a52653b40acb149/scripts/extra_script.py#L109>, the problem you are seeing is because the files are still in the dist dir. We either need to remove the hash from the file name and/or clean out the dist dir before building.

KipK:
I've updated, it nows delete .h statics if there's a gui build

jeremypoulter:
I think it is more the dist dir in the GUI you need to clear, then selectively delete the header files that don't exist in the web static directory. If you delete all the header files they will have to be generated every build.

KipK:
It only deletes the headers if the dist dir is not empty. Building the gui always clears the dist dir.
If we delete the files at the fw build process I don't get how it can work ?

jeremypoulter:
@KipK did you mean to close?

KipK:
nope I have forced pushe the static file update to this branch, forgot this branch was for this pull request. 
Will have to redo somehow as it's all lost

Issue #673, opened 2023-06-09
Latest master: WiFi AP SSID Incorrect 


glynhudson:
I'm just testing latest master branch https://github.com/OpenEVSE/ESP32_WiFi_V4.x/commit/4c8e64e1af919abcc9125cdcda0f54ee1427cc03 on a fresh hardware and I've noticed the WiFi AP SSID is incorrect. It seems to be broadcasting `ESP440385` without any security 

![Screenshot_20230609-161239](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/assets/758844/3a7fc2d1-b18e-444d-81e4-e4f1da823b6e)
![Screenshot_20230609-161230](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/assets/758844/5721e78b-f413-4ed1-b93c-646605f2bb96)

![PXL_20230609_151317109](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/assets/758844/0f4729f5-1a93-4c56-bfe5-25166f75a2ee)

![PXL_20230609_151329700](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/assets/758844/8f0c527f-0ef9-4a4f-9777-16f3471ba763)

PR #674, opened 2023-06-09
fix ap wifi wrong  #673

KipK:


glynhudson (commented):
Nice work, this has fixed it

glynhudson (approved):

PR #700, opened 2023-07-21
Fix #434: Implement Uptime

### LINKED ISSUE #434: Enhancement: Uptime status clock
Would be nice, especially in these times of heavy development, to have a way to see uptime of the ESP32 and wifi connections.

mathieucarbou:
Example in HA:

```yaml
mqtt:
  sensor:
    - name: OpenEVSE Uptime
      unique_id: 55806B85-C3B9-4EF7-8DB3-2E6C28D97669
      icon: mdi:clock
      state_topic: "openevse/uptime"
      device_class: duration
      state_class: measurement
      unit_of_measurement: s
      value_template: "{{ value|int }}"
      expire_after: 40
```

In HA:

image

MQTT Integration: https://gist.github.com/mathieucarbou/92a3d5e0dc38d6b68aa1bdaf153a80c5

In MQTT:

image

`/status`:

image

`/time`:

image

### DIFF
--- a/api.yml
+++ b/api.yml
@@ -773,10 +773,14 @@ paths:
                     x-stoplight:
                       id: 7h1cpgfjj5j9j
                     format: date-time
+                  uptime:
+                    type: integer
+                    description: EVSE gateway uptime, in seconds
                 required:
                   - time
                   - offset
                   - local_time
+                  - uptime
       operationId: get-time
       description: Gets the time set on the OpenEVSE
     post:
--- a/models/Status.yaml
+++ b/models/Status.yaml
@@ -175,3 +175,6 @@ properties:
   limit_version:
     type: integer
     description: /limit endpoint current version
+  uptime:
+    type: integer
+    description: EVSE gateway uptime, in seconds
--- a/src/time_man.cpp
+++ b/src/time_man.cpp
@@ -278,4 +278,5 @@ void TimeManager::serialise(JsonDocument &doc)
   doc["time"] = time;
   doc["local_time"] = local_time;
   doc["offset"] = offset;
+  doc["uptime"] = uptimeMillis() / 1000;
 }
--- a/src/emonesp.h
+++ b/src/emonesp.h
@@ -173,4 +173,6 @@ extern String serial;
 
 void restart_system();
 
+uint64_t uptimeMillis();
+
 #endif // _EMONESP_H
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -175,6 +175,8 @@ void
 loop() {
   Profile_Start(loop);
 
+  uptimeMillis();
+
   Profile_Start(Mongoose);
   Mongoose.poll(0);
   Profile_End(Mongoose, 10);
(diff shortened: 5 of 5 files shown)


api.yml
```diff
@@ -773,10 +773,16 @@ paths:
                     x-stoplight:
                       id: 7h1cpgfjj5j9j
                     format: date-time
+                  uptime:
+                    type: string
+                    type: number
+                    format: int32
+                    minimum: 0
```
mathieucarbou, line 780: I am really not sure about that. Please tell me what to put if it needs a change ;-)

models/Status.yaml
```diff
@@ -175,3 +175,6 @@ properties:
   limit_version:
     type: integer
     description: /limit endpoint current version
+  uptime:
+    type: integer
```
mathieucarbou, line 179: we don't make the difference here between `integer`, `long`, `unsigned` numbers ? `integer` means `number` ? Or current entries are incorrect and should all be `number` instead ?

src/time_man.cpp
```diff
@@ -278,4 +280,5 @@ void TimeManager::serialise(JsonDocument &doc)
   doc["time"] = time;
   doc["local_time"] = local_time;
   doc["offset"] = offset;
+  doc["uptime"] = millis() / 1000 - bootTimeSecs; // Note: overflow not checked because this gives about 136 years (2^32/60/60/24/365)
```
jeremypoulter, line 283: Not sure if this is the best place, but `/status` is also maybe not the best place either. Lets leave here and see how people find it. Don't think you really need the `- bootTimeSecs`, `millis()` returns the time since boot anyway, and also not sure your comment about overflow is the correct time (https://www.arduino.cc/reference/en/language/functions/time/millis/):

> Returns the number of milliseconds passed since the Arduino board began running the current program. This number will overflow (go back to zero), after approximately 50 days.

models/Status.yaml
jeremypoulter: Think integer is correct for the OpenAPI spec.

api.yml
jeremypoulter: Would just be the same as in `Status.yaml` 

```yaml
                  uptime:
                    type: integer
                    description: EVSE gateway uptime, in seconds
```

src/time_man.cpp
mathieucarbou: Oh! My bad! I have a Java developer background and I thought this function would return the Unix time (traditional time-stamp). So now I understand why all implementations have a rollover at 50 days! I will change. This is not correct indeed.

mathieucarbou: I've change to a more flexible implementation on 64 bits which is working fine as long it is called once every 47 days

mathieucarbou:
reminder ;-)

src/time_man.cpp
```diff
@@ -258,6 +258,18 @@ String time_format_time(tm &time)
   return String(output);
 }
 
+
+// inspired from https://www.snad.cz/en/2018/12/21/uptime-and-esp8266/
+// this works if this method is called at least once every 47 days, which is the case (each 8 hours at least based on TIME_POLL_TIME)
```
jeremypoulter, line 263: There is no direct link to `TIME_POLL_TIME` and calling `TimeManager::serialise` and although chances are high that it would be called, you may want to add something explicit into the `loop` method. Also be aware that the `loop` method is only called regularly if SNTP is enabled.

mathieucarbou: True, I will add a more direct link because here, it works, but I rely on the whole call hierarchy.

@jeremypoulter Regarding the loop which could be disabled...

- should we then move the update stuff in its own cpp file with its own microtask ?
- should we couple the uptime code to another existing loop, which is less time-related ?
- should we keep the uptime code here, and accept the fact that time-related things need NTP ?

mathieucarbou: @jeremypoulter : I've updated the PR to move the uptime near the main loop, where the version and hardware info are.

jeremypoulter (approved):
Happy with this, thanks for the contribution

PR #713, opened 2023-08-11
Initial attempt at porting to the ESP32-S3

jeremypoulter:


CapnBry:
For what it is worth, I made an environment for the [ESP32-C3 devkit](https://docs.espressif.com/projects/esp-idf/en/latest/esp32c3/hw-reference/esp32c3/user-guide-devkitc-02.html) that seems to be working just fine after a month of use. I am using a more recent espressif32 release though, as the C3 isn't supported on 6.0.1 I believe
```
[env]
- platform = espressif32@6.0.1
+ platform = espressif32@6.4.0
```

This is the environment definition. I chose GPIO6 for RAPI TX and GPIO7 for RAPI RX somewhat arbitrarily, just because they aren't ADC pins or the USB DP/DN pins, JTAG pin, or UART0 TX/RX.
```
[env:esp32-c3-devkitc-02]
board = esp32-c3-devkitc-02
lib_deps =
  ${common.lib_deps}
  ${common.neopixel_lib}
build_flags =
  ${common.build_flags}
  ${common.src_build_flags}
  ${common.debug_flags}
  -D RAPI_PORT=Serial1
  -D TX1=6
  -D RX1=7
  -D NEO_PIXEL_PIN=8
  -D NEO_PIXEL_LENGTH=1
  -D WIFI_PIXEL_NUMBER=1
```
Log messages come out the USB port, the RGB LED color cycles on boot, and my OpenEVSE works just fine with it after I rolled back the firmware from 8.2.3 (this firmware really should be pulled down since nothing supports BOOTLOCK so it effectively soft bricks the device, it also likes to reboot the atmega every few days). I believe my OpenEVSE is a v2.5? (2014 it was listed as "OpenEVSE 50A Deluxe Charge Station Combo - Quick Kit with RGB LCD/RTC / OpenEVSE Artwork")

jeremypoulter:
Thanks for the feedback, would you be able to add a separate PR for that board? Will have to get a C3 to test...

CapnBry:
> Thanks for the feedback, would you be able to add a separate PR for that board? Will have to get a C3 to test...

Yeah sure, can do

PR #743, opened 2023-10-12
Added Support for wt32-eth01 board - Update platformio.ini

jdgarcia99:
Added Support for wt32-eth01 board. Need to add a 1uF ceramic capacitor between pins EN and GND to avoid boot problems
wt32-eth01 capacitor

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -319,3 +319,26 @@ build_flags =
   -D RAPI_PORT=Serial1
   -D RX1=25
   -D TX1=27
+
+[env:wt32-eth01]
+# For Wireless Tag
+board = wt32-eth01
+build_flags =
+  ${common.build_flags}
+  ${common.src_build_flags}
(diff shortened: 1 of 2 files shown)


platformio.ini
```diff
@@ -319,3 +319,26 @@ build_flags =
   -D RAPI_PORT=Serial1
   -D RX1=25
   -D TX1=27
+
+[env:wt32-eth01]
```
jeremypoulter, line 323: Changes look ok, but can you add to the GitHub workflow: <https://github.com/OpenEVSE/openevse_esp32_firmware/blob/master/.github/workflows/build.yaml#L57>

jdgarcia99: Done!

jeremypoulter (approved):
Thanks for your contribution. Changes look good, and built successfully. I can't test as I don't have this board, but happy to accept.

Issue #708, opened 2023-08-02
Cannot compile on Windows 


aaronwesel:
Am I the only one having this issue? Wihtout any modification, if I download the Wifi files from github and try to compile it using pio I get an error "Littlefs" not found. What am I doing wrong?

jeremypoulter:
What enviroment are you using to compile? Can you show a full log? We do build via a GitHub workflow [![Build/Release OpenEVSE](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/actions/workflows/build.yaml/badge.svg)](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/actions/workflows/build.yaml) and that is passing.

aaronwesel:
I am using vscode pio, I followed all the instructions. I am not sure where to get these logs. Here is  screenshot if that helps.
![image](https://github.com/OpenEVSE/ESP32_WiFi_V4.x/assets/8095386/f592e3b2-4f65-4e24-a663-fdc064ff90c2)
<img width="481" alt="Clipboarder 2023 08 02" src="https://github.com/OpenEVSE/ESP32_WiFi_V4.x/assets/8095386/f67f407e-84bd-4422-a09d-6b5eb1d138d5">

jeremypoulter:
Which board/env are you building? Are they actually compilation errors from Platform IO, or just VSCode trying to parse the code? The latter may just be a VS Code miss configuration of VS Code and not impact building/uploading/etc

aaronwesel:
Yes, these are build errors. It is the same for every env I try. Here is
the terminal output.

 *  Executing task: C:\Users\techo\.platformio\penv\Scripts\platformio.exe
debug

Processing openevse_esp32-gateway-e_dev (board: esp32-gateway; platform:
***@***.***; framework: arduino)
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------Verbose
mode can be enabled via `-v, --verbose` option
Firmware Revision: -D BUILD_TAG=local_master_ff55cf44_modified -D
BUILD_HASH=ff55cf44_modified
CONFIGURATION:
https://docs.platformio.org/page/boards/espressif32/esp32-gateway.html
PLATFORM: Espressif 32 (6.0.1) > OLIMEX ESP32-GATEWAY
HARDWARE: ESP32 240MHz, 320KB RAM, 4MB Flash
[...]

PR #747, opened 2023-10-22
Update OCPP lib, fix building on Windows

matth-x:
This PR updates the OpenEVSE OCPP integration to the latest version and contains a fix for building on Windows (see #708, sorry for not creating a separate PR).

In the latest version I changed the class structure of Configs so that they are represented by an interface instead of only a concrete class. This allows to wrap OpenEVSE configs into an adapter (`OcppConfigAdapter`) which implements the `Configuration` interface and to pass this adapter to the OCPP lib. Now, the OCPP lib accesses the OpenEVSE configs directly instead of only synchronizing OpenEVSE configs with OCPP configs. This also means that there is a "single source of truth" - the value of each config exists in only one place now.

A special case is updating the OCPP backend URL. If this is done by the user through the GUI, it should apply immediately, resetting the WebSocket connection and reconnecting to the new URL. When the OCPP backend changes the URL, then the connection shouldn't reset immediately, but wait for the next reboot. But when the OcppConfigAdapter updates the OpenEVSE config, the `config_changed()` handler in app_config.cpp gets triggered and finally the `notifyConfigChanged` in ocpp.cpp. In this case the WebSocket reconnect is suppressed by setting the variable `synchronizationLock` and resetting it afterwards.

Still, I think that managing the configs has become much simpler now.

Btw. I renamed the library into MicroOcpp. This PR also reflects this change. Moreover, the previous class name `ArduinoOcppTask` has become outdated and I changed it into `OcppTask`. The same goes for the logo of the OCPP panel for which I will open a separate PR in the gui repo.

The Windows fix is related to generating `staticFiles` array in extra_script.py. When running on Windows it concatenates dir and filename with `\` which leads to compilation errors in the web_server_static_files.h header. E.g. I got the string `"/assets\components-aa3d0296.js"`, but the patch changes all `\` into `/` in the final step of generating the `staticFiles` array.

TL;DR:

- [x] Adopt new project name of OCPP lib
- [x] Rename `ArduinoOcppTask` into `OcppTask`
- [x] Pass OpenEVSE configs to OCPP lib instead of synchronizing config objects of both projects
- [x] Fix generating paths on Windows (#708)
- [x] Replace AO-logo in GUI by official OCPP logo by the OCA (https://github.com/OpenEVSE/openevse-gui-v2/pull/13)


matth-x:
Thanks for accepting the PR in the gui repo. I think there's no real dependency between both PRs so I leave the gui update to #748.

jeremypoulter:
Needs #796 to be merged to fix divert_sim build error

jeremypoulter (approved):

PR #766, opened 2023-11-17
Add esp32-c3-devkitc-02

CapnBry:
As I tried to foist this off in #713, this adds support for the [ESP32-C3 devkit](https://docs.espressif.com/projects/esp-idf/en/latest/esp32c3/hw-reference/esp32c3/user-guide-devkitc-02.html) that seems to be working just fine after a month of use. I am using a more recent espressif32 release (6.4.0)  though, as the C3 isn't supported on 6.0.1.

I chose GPIO6 for RAPI TX and GPIO7 for RAPI RX somewhat arbitrarily, just because they aren't ADC pins or the USB DP/DN pins, JTAG pin, or UART0 TX/RX.

| Pin | Function |
|--|--|
| GPIO0 | WiFi Button / BOOT |
| GPIO6 | RAPI TX |
| GPIO7 | RAPI RX |
| GPIO8 | RGB LED |
| GPIO21 | USB Serial TX (cp2102) |
| GPIO20 | USB Serial RX (cp2102) | 

Log messages come out the USB port, the RGB LED color cycles on boot, and my OpenEVSE works just fine with 8.2.2. I believe my OpenEVSE is a v2.5? (2014 it was listed as "OpenEVSE 50A Deluxe Charge Station Combo - Quick Kit with RGB LCD/RTC / OpenEVSE Artwork")

The ESP Info on the firmware page shows "UNKOWNr3 1 core WiFi BLE" due to ESPAL [not supporting anything except the basic ESP32](https://github.com/jeremypoulter/ESPAL/blob/master/src/espal_esp32.h#L30)

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -56,6 +56,7 @@ jobs:
           - olimex_esp32-poe-iso
           - heltec_esp32-wifi-lora-v2
           - wt32-eth01
+          - esp32-c3-devkitc-02
         gui:
           - name: gui-v2
             repo: OpenEVSE/openevse-gui-v2
--- a/platformio.ini
+++ b/platformio.ini
@@ -111,7 +111,7 @@ neopixel_lib = adafruit/Adafruit NeoPixel@1.11.0
 #platform = https://github.com/platformio/platform-espressif32.git#feature/stage
 #platform = https://github.com/platformio/platform-espressif32.git#develop
 #platform = espressif32@1.11.1
-platform = espressif32@6.0.1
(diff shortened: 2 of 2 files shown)


jeremypoulter:
Thanks for the contribution, could I just ask you add this to the build workflow, eg <https://github.com/OpenEVSE/openevse_esp32_firmware/pull/743/files#diff-d0777657fa3fd81d23aaf7273e58aee453b04e67882517900c56daeef9b3e4c1>

jeremypoulter:
Looks to work well

PR #784, opened 2024-01-09
Did for current shaper not working under certain circumstances 

### LINKED ISSUE #762: Shaper not working anymore with release v5.0.2
I have install the release v5.0.2 on a EVSE with 8.2.3.T2 firmware and Shaper does charge current according to my real time house load,
for exemple the maximum power load is set to 8400W:
image
but the real power load is 9428W. The Shaper set up the charge current to 31,68A for this load which is to high. Charge current decrease only a little with the power load it look like there is a wrong scaling factor in the shaper charge current regulation
If I go back to previous recommanded release v4.2.2 everything work fine exemple:
image

For comparison release v5.0.2 versus v4.2.2 for the Shaper
You find below the graphic of the recorded shaper charge current regulation, EvSE power usage and house power load. From 22:00 to 22:16 the EVSE was working with release v5.0.2 and after 22:24 the EVSE was working with release v4.2.2

image
there no current regulation with release v5.0.2 and power load is not limited to 8400W
and after 22:24 the EVSE Shaper is working well with release v4.2.2 with a good current regulation

jeremypoulter:
fix #762

### DIFF
--- a/src/current_shaper.cpp
+++ b/src/current_shaper.cpp
@@ -174,9 +174,9 @@ void CurrentShaperTask::shapeCurrent() {
 			max_pwr += solar;
 		}
 	}
-	if (livepwr > max_pwr) {
-		livepwr = max_pwr;
-	}
+//	if (livepwr > max_pwr) {
+//		livepwr = max_pwr;
+//	}
 	if(!config_threephase_enabled()) {
 		_max_cur = ((max_pwr - livepwr) / evse.getVoltage()) + evse.getAmps();
 	 }


KipK:
This was there to prevent negative calculation when live_pwr > max_pwr. As in this case, evse should stop anyway.

I'm pretty sure the 73 issue is not caused by this.
However I couldn't reproduce it here.

zerodur12:
I test the build openevse_wifi_v1 on my EVSE with 8.2.3.T2 firmware. Every thing work well the current regulation work again to target max_pwr (set to 8400W), I also test a change in max_pwr value during the charge (around 20h in the figure below) and the regulation follow the new max_pwr value without problem.
@KipK if you need to prevent negative calculation of the max_cur from any value of live_pwr >> max_pwr,  it seems better to force the result of current calculation to be greather than 0 than making live_pwr = max_pwr, if live_pwr is saturated to any fix value it is like opening the close loop with no mesure any more of live_pwr (from my point of view)
image

graememk:
Removing that check has allowed my install to load balance as expected.

On the latest release build (5.0.2), The EVSE could sit at full speed and still be a few kW over the set limit. 

The negative Value in the max_pwr - live_pwr is key to the operation of the load balancing

max_pwr = 8000, live_pwr = 10000, voltage = 230, getAmps() = 32

_max_cur = ((max_pwr - livepwr) / evse.getVoltage()) + evse.getAmps();

_max_cur = ((8000 - 10000) / 230) + 32
_max_cur = (-2000/230) + 32
_max_cur = (-8.7) + 32
_max_cur = 23.3A

Possibly a check on _max_cur to prevent that from going negative, If the reduction was more than 32 amps _max_cur would go negative. (or rollover?)

Ran a test: It goes negative, Charging stops
image

KipK (approved):

graememk:
In an abundance of caution,
Should the _max_cur be checked for negatives before being returned, it does work, as is but it could be a precaution.
If less than 0, _max_cur =0;

PR #819, opened 2024-04-01
Fix TFT I2C devices

rbray89:
Add the missing build flags to enable I2C devices (NFC reader and Temperature sensor).


chris1howell (dismissed):

rbray89:
This needs workflow approval

rbray89:
should have the missing library for the temp sensor added now.

would have changed the original commit, but using my phone, and didnt see a way to do that via the github UI.

jeremypoulter (approved):

jeremypoulter:
Thanks for the contribution

PR #865, opened 2024-06-20
Swap SDA and SCL pins for TFT Wifi

rbray89:


rbray89:
I don't have a board handy, but it looks like according to the images and WRoom-32 docs, that SDA and SCL for the I2c line are swapped. This is easy to verify on a stock wifi tft kit, as the auxillary temp sensor doesn't show up.

rbray89:
I dont see any build docs available, so I can't verify without a build or a way to generate one.

jeremypoulter (approved):

PR #899, opened 2024-08-22
Adding LedBar Animation for Visual effects of OpenEVSE activity and status

jdgarcia99:
Here is my modification to the original OpenEVSE project to introduce a led bar as most commercial EVSEs have as seen in this discussion thread https://github.com/OpenEVSE/openevse_esp32_firmware/discussions/891

I'm based on WS2812FX library to perform animations and visual effects.

My first intention is to show visually the status of the EVSE without using the screen nor the web interface and know quickly what is happening in the EVSE as follows:

If the EVSE is not charging the car, the bar have the same color code as the EVSE led (screen, buttons, etc) based on EVSE status
If the EVSE has any error, the led is blinking with corresponding color code
IF the EVSE is loading the car, the led will perform a "loading bar" animation to show this status. Also the speed of the animation will correspond with the intensity in Amps being provided to the car based on max Amp configured.
Led Bar bright can be configured directly from EVSE web interface Led brightness slider, so it is very intuitive to use and naturally integrated

Source code changed for this mod has been mainly done in LedManagerTask to create the integration

Notes for the implementation:

1. My Led Bar uses 15 pixels (you can configure num of pixels based on your own requirements for your bar) Param: NEO_PIXEL_LENGTH=15
2. I'm using the board WT32-ETH01 for the implementation. Thus I'm using some specific pin to communicate with the LED Bar. Change this configuration based on your own esp32 board. Param: NEO_PIXEL_PIN=12
3. An extra parameter has been created to use this functionality on the EVSE. Param: ENABLE_WS2812FX
4. In my use case, I'm also using a PN532 RFID reader to authorize session. Comment this param if not needed.
5. In the event of using PN532 reader with WT32-ETH01 board through I2C You need to comment pin definition for this board in the Arduino.h. Otherwise it will give you an error due to conflict with pins used for I2C
6. Due to the low power of OpenEVSE AC Transformer I decided to use an external AC/DC power source to not overload OpenEVSE power source when full brightness (in future releases of the HW a more amp capable transformer can be used)

New animations and status can be introduced in the future based on this initial implementation.

Some pictures of the OpenEVSE working with this new functionality:

https://github.com/user-attachments/assets/b81d856e-94d1-477d-9cc0-2cc6516dab63

358304596-50391d9d-2dcd-44c9-bacd-6c80bb85f7b4

358304602-dcb84136-a5da-4086-a372-7aa2f38df942

### DIFF
--- a/src/LedManagerTask.cpp
+++ b/src/LedManagerTask.cpp
@@ -745,11 +745,11 @@ void LedManagerTask::setBrightness(uint8_t brightness)
   MicroTask.wakeTask(this);
 }
 
-LedManagerTask ledManager;
-
 #if defined(NEO_PIXEL_PIN) && defined(NEO_PIXEL_LENGTH) && defined(ENABLE_WS2812FX)
-extern void ledManager_loop()
+void LedManagerTask::service()
 {
   ws2812fx.service();
 }
 #endif
+
+LedManagerTask ledManager;
--- a/src/LedManagerTask.cpp
+++ b/src/LedManagerTask.cpp
@@ -44,7 +44,7 @@ class LedAnimatorTask : public MicroTasks::Task
     }
     unsigned long loop(MicroTasks::WakeReason reason) {
       ws2812fx.service();
-      return 40;
+      return 10;
     }
 } animator;
 #endif
--- a/src/LedManagerTask.cpp
+++ b/src/LedManagerTask.cpp
@@ -30,9 +30,12 @@
 #include "emonesp.h"
 #include "LedManagerTask.h"
 
-#if defined(NEO_PIXEL_PIN) && defined(NEO_PIXEL_LENGTH)
+#if defined(NEO_PIXEL_PIN) && defined(NEO_PIXEL_LENGTH) && !defined(ENABLE_WS2812FX)
 #include <Adafruit_NeoPixel.h>
 Adafruit_NeoPixel strip = Adafruit_NeoPixel(NEO_PIXEL_LENGTH, NEO_PIXEL_PIN, NEO_GRB + NEO_KHZ800);
+#elif defined(NEO_PIXEL_PIN) && defined(NEO_PIXEL_LENGTH) && defined(ENABLE_WS2812FX)
+#include <WS2812FX.h>
+WS2812FX ws2812fx = WS2812FX(NEO_PIXEL_LENGTH, NEO_PIXEL_PIN, NEO_GRB + NEO_KHZ800);
 #endif
 
 #define FADE_STEP         16
@@ -40,6 +43,13 @@ Adafruit_NeoPixel strip = Adafruit_NeoPixel(NEO_PIXEL_LENGTH, NEO_PIXEL_PIN, NEO
 
 #define CONNECTING_FLASH_TIME 450
 #define CONNECTED_FLASH_TIME  250
+
+#if defined(ENABLE_WS2812FX)
+// Speed for FX Bar Effects
+#define DEFAULT_FX_SPEED 1000
+#define CONNECTING_FX_SPEED 2000
+#define CONNECTED_FX_SPEED  1000
+#endif
 
 #define TEST_LED_TIME     500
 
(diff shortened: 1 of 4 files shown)


src/main.cpp
```diff
@@ -215,6 +215,10 @@ loop() {
   MicroTask.update();
   Profile_End(MicroTask, 10);
 
+#if defined(NEO_PIXEL_PIN) && defined(NEO_PIXEL_LENGTH) && defined(ENABLE_WS2812FX)
+  ledManager_loop();
+#endif
```
jeremypoulter, line 220: This should not be needed, the MicroTask and the LedManagerTask instance provide the equivelant without the need to add lots of loop calls to the main `loop` function.

If you can't use the LedManagerTask you could setup another task just for running the animations.

src/LedManagerTask.cpp
```diff
@@ -125,11 +171,22 @@ void LedManagerTask::begin(EvseManager &evse)
...
+#elif defined(NEO_PIXEL_PIN) && defined(NEO_PIXEL_LENGTH) && defined(ENABLE_WS2812FX)
+  DEBUG.printf("Initialising NeoPixels WS2812FX MODE...\n");
+  ws2812fx.init();
+  ws2812fx.setBrightness(brightness);
+  ws2812fx.setSpeed(DEFAULT_FX_SPEED);
+  ws2812fx.setColor(BLACK);
+  ws2812fx.setMode(FX_MODE_STATIC);
+  //ws2812fx.setBrightness(this->brightness);
+  DEBUG.printf("Brightness: %d ", this->brightness);
+  DEBUG.printf("Brightness: %d ", brightness);
```
jeremypoulter, line 188: Use `DBUGF` so these debug lines get compiled out in release builds. To enable the debug you need to define `ENABLE_DEBUG_LED` in `platformio.ini`
```suggestion
  DBUGF("Brightness: %d ", this->brightness);
  DBUGF("Brightness: %d ", brightness);
```

src/main.cpp
jdgarcia99: Hi Jeremy. 
Understood. However the main reason to call this is to invoke the ws2812fx.service() method which refreshes the led status bar.

I'm not sure about the Microtasks architecture as I could find much information about.
The problem I have is that the service() method need to be invoked regularly to update the bar and If I do not force the call the bar is not updated. I could find any way within ledbartask to execute this regularly. Only when the EVSE changes status that invokes the service just once and then the animation is stopped

Should I call from main loop just a service method directly instead?  
I mean:
 create a method within the ledManagerTask  called service and then launch from the main loop? (that is in essence what this ledManager_loop call is doing)

Thanks in advance

jdgarcia99:
Alright. Just solved by adding a new method for ledManagerTask to invoke the ws2812fx.service method to run the animation. Then added a call to this method from main loop to refresh effect as ws2812fx.service but be invoked in each loop run to perform the animation

jeremypoulter:
Please can you update all the DEBUG.printf calls? 

On the looping there are two ways you can do it, the return of `LedManagerTask::loop` gives the amount of time before that method is called again so to make that function be called repeatedly you can just return a small value like `40`, that should be sufficiaent for 25 fps, but what might b cleaner could be an separate task for just running the animations, something like:

```c++
class LedAnimatorTask : public MicroTasks::Task
{
  public:
    void setup() {
    }
    unsigned long loop(MicroTasks::WakeReason reason) {
      ws2812fx.service();
      return 40;
    }
};

LedAnimatorTask animator;
```

Then in `LedManager::setup` you would call `icroTask.startTask(&animator);` to start the task. See https://github.com/OpenEVSE/openevse_esp32_firmware/commit/5c1923130c373abee11d1ee3dc5c138aa864f70c

jdgarcia99:
OK, Thank for your input Jeremy. Already tested your code and works perfect. 40ms seem enough to run the animation smoothly.
Also Cleaned the DEBUG calls


https://github.com/user-attachments/assets/f0db17f0-c20e-43a0-a141-ec8e0db9d013

jeremypoulter:
Looks good, a little jitter sometimes so maybe 20 might be better.

jdgarcia99:
Tried different refresh times for the task and found that 10ms is visually the best (although 20 ms is enough). Find videos:

40ms:


https://github.com/user-attachments/assets/d13ab86b-f615-493b-9966-383adbb93cbc

20ms:

https://github.com/user-attachments/assets/1e9045a9-1f6f-466c-96a5-2c3fe84be35a

10 ms:


https://github.com/user-attachments/assets/8d17ca14-9e2d-4a83-992d-ea787de8adb8

I'll let mine in 10 ms as I thing is best (at least visually in person) and the general behaviour is responsive without any effectt on the EVSE

jeremypoulter:
Great, 10ms is good. Thanks for testing

jeremypoulter (approved):

jeremypoulter:
All looks good, thank you for your contribution

PR #914, opened 2024-09-27
More information on TFT display + TFT backlight timeout

kimble4:
I expect the entire TFT UI will be re-done at some point, but as an interim measure this adds some more useful information to the display:

openevse_improved_tft_ui

- Dedicated icon for the vehicle connection status, like in the Web UI.
- WiFi connection status icon (artwork is a bit rough and ready) that also indicates whether in station or AP mode.
- The actual power/current being drawn by the vehicle, and the supply voltage (which can be updated in realtime by MQTT).
- The EVSE temperature.
- Current limit has shrunk a bit to fit this all in.

I've also changed the date format to ISO8601; in the absence of any localisation this is considerably less confusing for non-American users.

Meanwhile, setting `TFT_BACKLIGHT_TIMEOUT_MS` enables a timeout of the TFT's backlight.  (Value is in milliseconds, so 600000 = 10 minutes.)  The backlight will wake automatically when the EVSE or WiFi connection state changes, or if a message is displayed.

If `TFT_BACKLIGHT_CHARGING_THRESHOLD` is set, the display will timeout during charging if the current drawn drops below the set value in Amps.


clowrey:
Beautiful! 1000% want

jeremypoulter:
Thanks for the contribution, this is a good improvement. I will do a proper review. Indeed the intention is to move over to a proper framework but we have some other changes to do first.

jeremypoulter (approved):
@kimble4 Changes look good, and tested, looks to work well.

Maybe one thing to consider, 
Instead of 
```
#ifdef ENABLE_SCREEN_LCD_TFT
  _lcd.setWifiMode(true, false);
#endif //ENABLE_SCREEN_LCD_TFT
```

Maybe just define an empty method in the on TFT to get rid of the `#ifdef` 

Do you have any other plans to add more changes?

jeremypoulter (approved):

kimble4:
That's a good idea; wasn't sure which was the less invasive change to make.

I don't have any plans to add to this, but I didn't last week either :)  Happy to draw a line here and start a new branch if I think of something else to add.

Longer term things like localisation, colour themes, maybe some graphs or displaying the history or whatever are best left for a proper framework.  But it's really nice to be able to see the current and (particularly for portable use) what the WiFi is doing.

jeremypoulter:
Yeah I agree. I am refreshing the networking layers at the moment but a framework is next on my list, probably LVGL but lib [Websockets](https://github.com/warmcat/libwebsockets) also looked good

jeremypoulter (approved):

PR #920, opened 2024-10-22
Add filter to current shaper output.

lutorm:
As I described in https://github.com/OpenEVSE/openevse_esp32_firmware/issues/901, I've been having problems with instabilities in the current shaper. From looking at the issues, it appears that others have as well. I can replicate the problem by artificially delaying the incoming live power samples by about one sample. (This kind of delay is not unreasonable given that the live power may travel from the meter to some cloud provider and back to the evse.) This makes sense since the current implementation has unit gain, ie it tries to correct the entire power mismatch in one step. If the input power is delayed one step, that will then put the corrections 90 degrees out of phase with the input, causing instability.

This PR adds an instance of the IIR filter used for input smoothing to the current shaper calculated max current. By filtering the shaper output so it doesn't attempt to correct the entire mismatch in one step, the gain can be lowered so it's stable regardless of input delay.

The graph below shows a test. Initially the filter time constant is set to zero, mimicking the current implementation. The session starts at 8:57, and just after 9:00 a load is switched on and the shaper becomes active. The max current is 20A and the shaper should limit the charge current to 10A, but oscillates between 6-16A. At 9:04 another load is switched on and the shaper pauses the charge. At 9:06 that load switches off and the charge resumes, reentering the same oscillation.

Just before 9:10, the filter time constant is changed to 10s and the change in this PR becomes active. 10s is enough filtering for the oscillation to damp out in about 90s and the charge current stabilizes with a live load of 20A as designed.  Just after 9:17, all loads are disconnected and, after a small overshoot, the current shaper stabilizes at 19A free current which is larger than the pilot max, so it becomes inactive.

shaper output filter


jeremypoulter:
Thanks for your contribution I will take a look.i

I do have concerns that this may negatively impact some of the shapers use cases, particularly those around safety and load balancing where quick responses are desirable. Ideally we need to add current shaper support to the (divert) simulator so we can test all of these use cases and see the impact of the different values. Probably the other thing that maybe needed is a separate config values for the input and output filters

lutorm:
@jeremypoulter I agree that slowing down the current throttle is suboptimal. But having it go unstable is worse, IMHO.

My grid monitor reports live power every 10s. With a 10s time constant it'll take maybe 3 samples to cut back fully, as opposed to ideally a single sample. This is still safely below the trip time for normal breakers for a maybe 1.5x overcurrent that would be a common scenario. (The worst possible case would be a vehicle charge current equal to the main breaker value when suddenly another load comes on that's also equal to the main breaker value. This would be a 2x overcurrent which, for the breakers commonly used here in Europe, has a min trip time of 10s. This would be iffy to handle with a live power reporting rate of 10s regardless of smoothing, but is a pathological case. For a 1.5x overcurrent, the trip time is at least 40s which will be safely handled.)
 
People with higher live power rates _and_ low sample delays could safely use a smaller time constant, what matters for stability is the delay in sample times.

As for having a separate config value, that would definitely be preferable. I'm not sure what the use case of the input smoothing is, as opposed to the min pause time, so it was easy for me to reuse the same time constant as a proof of concept. I'm just not familiar enough with the code base that it was immediately obvious how to add new config values.

jeremypoulter:
New config values cn be added in app_config.c/h with ppropriate changes to the GUI to allow changing. The Input filter Should be there to smooth out the incomming power values to get rid of any spikes or dips. I did start on adding the current shapper to the simulator for easy testing and visulisation of the changes, https://github.com/OpenEVSE/openevse_esp32_firmware/pull/923/, would be good to get this vompleted to see how these changes impact other use cases

PR #945, opened 2025-01-02
Fix divertmode MQTT docs to match HTTP API

joth76:
Copied definitions from https://openevse.stoplight.io/docs/openevse-wifi-v4/dba95e8966593-get-the-evse-status  (verified these values make sense experimentally)

### DIFF
--- a/docs/mqtt.md
+++ b/docs/mqtt.md
@@ -51,7 +51,7 @@ Limit:
 
 Main settings:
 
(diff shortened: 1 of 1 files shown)


jeremypoulter (approved):

jeremypoulter (approved):

jeremypoulter:
Thank you for your contribution

PR #946, opened 2025-01-02
TFT big power display during charging

kimble4:
By popular demand: Uses the big digits to display the measured power during charging.  When not charging, the pilot current setting is displayed here as before.

As there's only room for 3 digits, the number of decimal places changes to fit the available space.

IMG_20250102_221901602_HDR


jeremypoulter (approved):

jeremypoulter:
Thank you cor the contribution, sorry for the delay in testing

PR #964, opened 2025-04-29
Use latest (6.x) version of the ArduinoMongoose to resolver issues with Safari

### LINKED ISSUE #963: Safari issue with 304 responses
According to the rfc: https://datatracker.ietf.org/doc/html/rfc7232#section-4.1 

   A 304 response cannot contain a message-body; it is always terminated
   by the first empty line after the header fields.

It looks like Mongoose is sending back a body on 304.

Image

And this is the error message I see in Safari:

Image

I was able to reproduce this issue putting together a little test program that would listen on 80 and send back the same response that I pulled out of wireshark, and updating my /etc/hosts file to point to localhost instead of the openevse.  If I set Content Length to 0 and removed the "Not Modified" line, Safari no longer gave the parse error.

It looks like the 7.x Mongoose doesn't have this issue, but it's kind of hard to say for sure since it looks like a pretty substantial rewrite.  They also seem to do a better job of sending back more of the response headers on 304, like the etag header etc.

This Mac issue was mentioned in https://github.com/OpenEVSE/openevse_esp32_firmware/issues/930  but I think these might be separate, unrelated issues at this point.

jeremypoulter:
Should fix #963

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -31,7 +31,7 @@ default_envs = openevse_wifi_v1
 [common]
 lib_deps =
   bblanchon/ArduinoJson@6.20.1
-  jeremypoulter/ArduinoMongoose@0.0.20
+  jeremypoulter/ArduinoMongoose@0.0.21
   jeremypoulter/Micro Debug@0.0.5
   jeremypoulter/ConfigJson@0.0.6
   jeremypoulter/OpenEVSE@0.0.14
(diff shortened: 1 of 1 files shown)


nathanejohnson:
I'm running this latest build and I can confirm it fixes the issue!  Thanks for all the help!

KipK (approved):

PR #971, opened 2025-05-16
Refactored mqtt feature.

KipK:
Refactored mqtt feature.
- made it a microTask
- Mongoose MQTT client callbacks are set up to call methods of the Mqtt instance (using lambdas to capture this).
- Error Handling in attemptConnection: The _mqttclient.connect() now has a non-blocking callback for success. If connect() itself returns false, it means the connection attempt couldn't even be initiated (e.g., invalid parameters before network activity). This case is now handled by calling onMqttDisconnect. ( should fix https://github.com/OpenEVSE/openevse_esp32_firmware/issues/969 )
-  checkAndPublishUpdates() method consolidates the logic for periodically checking if any state (claims, config, etc.) has changed and needs to be published. The publishInitialState() is called upon connection to send the current state of all relevant items.
- notify...Changed() methods. These should be called by the respective modules (Config, EvseManager, Manual, Scheduler, Limit) when their data changes, allowing MqttTask to publish the updates promptly if connected, or upon next connection. This is more event-driven than purely polling versions in its own loop, though the loop still serves as a fallback and for connection management. 
TODO: use notify_xxx_changed() methods in related modules and remove mqtt versions check


KipK:
Seems there's a problem with compiling the divert_sim:
Error: ../src/mqtt.h:6:10: fatal error: MongooseMqttClient.h: No such file or directory

I couldn't check why yet.

KipK:
@Jeremy, just tested with #964, seems all good too.

jeremypoulter (dismissed):

jeremypoulter (approved):

jeremypoulter:
Gave this a test and looks to work, @KipK do you think it is good to merge?

PR #996, opened 2025-09-18
Changes to allow setting of the max hardware current

jeremypoulter:
This is a one time write value from the EVSE. Unfotunatly there is no way of telling if this has been set other than to try and set it and see if it fails.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -75,7 +75,7 @@ jobs:
         key: ${{ runner.os }}-${{ hashFiles('**/lockfiles') }}
 
     - name: Set up Python
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@v6
       with:
         python-version: '3.x'
 
@@ -101,7 +101,7 @@ jobs:
 
     steps:
     - name: Download the built assets
-      uses: actions/download-artifact@v4
+      uses: actions/download-artifact@v5
       with:
         path: artifacts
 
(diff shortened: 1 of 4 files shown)


jeremypoulter:
Yeah that was my concern about trying to guess if this had been set by checking if it was 80. I guess we can check the OpenEVSE firmware version to see if it is EU or not to check for 80 vs 32.

If we add that logic, then I guess we can put the same tickbox / warning in the settings. 

With regards to the other changes I don't think that should be a problem. May be best to do in a separate PR

chris1howell:
@KipK - Understood, the best scenario would be to build as few versions as possible and allow the write once command to set the limit. Merging the US and EU versions would allow the MAX to be set at 12, 16, 32, 63 in Europe or 12A, 16A, 24A, 32A, 40A, 48A, 80A in the US or anything in between.

jeremypoulter:
@chris1howell IIRC the reason the EU sets the hardcodes to 32A may is that is a legal requirement. I don't think it is a big deal to just check for the EU firmware, the worst that will happen if we get it wrong is trying to set a hardware value and failing.

chris1howell:
@jeremypoulter 63A is a valid Type 2 Ampacity in Europe according to IEC 62196-2, the resistor value for 63A on PP is 100 Ohm. OpenEVSE ships all controllers worldwide with the default firmware. I am fine with checking for EU firmware.

@KipK The EmonEVSE build is the only that adds -D PP_AUTO_AMPACITY. It should not be an issue to add to all builds. I assume if no resistor is detected it defaults to the pilot.

[env:emonevse]
extends = env:openevse
build_src_flags=
  ${common.eu_build_flags}
  -D PP_AUTO_AMPACITY
  -D MV_FOR_L2=230000L
  -D 'VERSION="${common.version}.T2"'

jeremypoulter:
@chris1howell I have made those changes, did you want the hardware value setting completely removing from the wizard?

PR #1010, opened 2025-10-14
OpenEVSE Firmware w/ Weigand Access Card 

aaspp:
This fork of the OpenEVSE FW uses the framework set by the RFID function to enable the use of Weigand (HID Standard) Access Cards with OpenEVSE. All that you need is an extra ESP32 to connect to the Wi-Fi board via I2C and the Card Reader to make it work. More information is given in the Read Me and the repository for the ESP32 Daughter board linked in the Read Me.


jeremypoulter:
@aaspp Please can you give some more details on your daughter board? But on the surface of it, it sounds like it should exist as its own repository.

PR #1011, opened 2025-10-14
Add OCPP TLS certificate configuration and SNI support

0xStuart:
Required for complaint WSS support.  I haven't the hardware to test this right now.

### DIFF
--- a/src/app_config.cpp
+++ b/src/app_config.cpp
@@ -82,6 +82,7 @@ String ocpp_server;
 String ocpp_chargeBoxId;
 String ocpp_authkey;
 String ocpp_idtag;
+String ocpp_certificate_id;
 
 // Time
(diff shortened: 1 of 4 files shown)


jeremypoulter:
@0xStuart Have you made appropriate UI updates to alloe the certificate to be set? Also please check the CoPilot comments

grantstephens:
Is there anything I can do to help move this along?

jeremypoulter:
Can you clarify what the expectations are? Just to add a public certificate to the root authorities (in which case there is no need for the config variable), or add a client certificate to authorise an OCPP client (in which case you need to extend the OCPP config in gui_nightshift to select that certificate and set the added config variable)? 

Also there has been a huge amount of development on master so you will need to bring those in.

PR #1092, opened 2026-06-07
feat(gui): make gui-nightshift 2026.24.0 the default web UI

RAR:
## Summary

Makes **gui-nightshift `2026.23.1`** the default bundled web UI, replacing gui-v2. This is the GUI default that the maintainer and collaborators have asked to standardise on.

gui-v2 is **not** removed — it stays a first-class option, selectable at build time with `GUI_NAME=gui-v2`.

## What changed

- **`.gitmodules` / `gui-nightshift`** — adds `gui-nightshift` as a submodule (tag `2026.23.1`, commit `ef3d46c`).
- **`scripts/extra_script.py`** — default `GUI_NAME` is now `gui-nightshift` when unset; `GUI_NAME=gui-v2` selects the old UI.
- **`src/web_static/*`** — the committed pre-built bundle is regenerated from gui-nightshift `2026.23.1` (firmware builds without a submodule checkout, so the committed assets are what ship).
- **`.github/workflows/build_gui.yml`** — the GUI-rebuild workflow now builds `gui-nightshift`.

`2026.23.1` drops the in-GUI Home Assistant integration page and adds MQTT topic settings for the pushed vehicle + home-battery fields.

## Size

The nightshift bundle is **smaller** than gui-v2 — on `openevse_wifi_v1` it drops flash usage from ~97.8% to ~89.5% (1,923,645 → 1,760,081 bytes), freeing ~160 KB on the tight 4 MB boards. Firmware builds clean (`openevse_wifi_v1` ✅).

## Note for maintainers

The submodule currently points at **`https://github.com/RAR/openevse-gui-nightshift`**. If you'd prefer it hosted under the OpenEVSE org, say the word and I'll repoint `.gitmodules` (and re-pin the submodule) once the repo lands there.

---


jeremypoulter:
Cool, will take a look, I have been pondering pulling the GUI back into this repo, originally we split it out as we were looking at a Node JS version of the WiFi firmware, but this never took off so it is now more trouble than it is worth, therefore if we are building a new UI then would like it just directly in this repo.

KipK:
I'm building it right now to test. 

You've probably already noticed, there's many warning mostly related to new svelte 5 runes like `let current = $state(value)`
In Svelte 5 value or initialName comes from $props().
Doing that will only copy the actual value from prop but won't keep the link automatically.

You will need to use $derived() like` let current = $derived(value)` if readonly , 
or use $effect() for writable ones like for the slider:

```
let { value = 0 } = $props()

let current = $state(0)

$effect(() => {
  current = value
})
```

As previous UI was made with svelte 4, some [changes are needed](https://svelte.dev/docs/svelte/v5-migration-guide) for proper svelte 5 handling.

RAR:
> I'm building it right now to test.
> 
> You've probably already noticed, there's many warning mostly related to new svelte 5 runes like `let current = $state(value)` In Svelte 5 value or initialName comes from $props(). Doing that will only copy the actual value from prop but won't keep the link automatically.
> 
> You will need to use $derived() like` let current = $derived(value)` if readonly , or use $effect() for writable ones like for the slider:
> 
> ```
> let { value = 0 } = $props()
> 
> let current = $state(0)
> 
> $effect(() => {
>   current = value
> })
> ```
> 
> As previous UI was made with svelte 4, some [changes are needed](https://svelte.dev/docs/svelte/v5-migration-guide) for proper svelte 5 handling.

Sorry about that, they should be corrected here: https://github.com/RAR/openevse-gui-nightshift/releases/tag/2026.23.2. I'll get this PR updated to that tag in the morning.

KipK:
no worry , there's no hurry :) 

I have a few feedback on some thing I've seen but won't have time until few days to make a proper review. 

great work so far

chris1howell:
@RAR Are you okay moving this into the repo as @jeremypoulter proposed? Nice work.

RAR:
Yes.  I have more changes on the main of my repo, so if you're going to do it pull main.

chris1howell (dismissed):

KipK:
Hi, here is a first quick feedback on the UI based on UX/Features. I haven't checked the code yet.

- desktop version:
 Seems the UI has not been refined yet for desktop.
 
image

 . no reponsive display on desktop it looks like it's stretched from mobile version.
 . missing clickable pointer on buttons

- I haven't tested a real charge yet, does all the metrics data from the previous UI still appears on screen ? I get only 2 fields for now displayed ( the energy consumtion ones ) but all the other ones are missing. 
How is it handled ? 
image


- limit: all limits should be on the same limit dialog. 
It's pretty confusing having range & soc limit on an area, and other limits on a separated dialog. 
Range and Soc are optional, related limits feature should only appears when the feature is enabled ( ie when there's data for soc & ev range ). I haven't checked if you've handled it like that already.

- time limit: the double slider displayed to set the time is not ux friendly. Also there's no feedback on the current position while you slide it on mobile.
 We should either use drop downs like before, or we find a new way of doing it ?

- System Limits are missing from the "EVSE" config section
  There  are 2 kind of limits, user limits and system ones. System ones when set appears on the main page but doesn't have the delete button. 

- the Shaper button should be more discreet. It's a security feature that mostly shouldn't be touched so easily. I wonder if we shouldn't take the opportunity to remove it or at least add an option to display it or not.  

- settings/ network 
   please keep signal quality bar, only having dbm is not really user friendly.

- on mobile, the navigation bar at the bottom can appears above the screen until we hide the browser url bar. It should always be apparent on top of the scrollable content ( fixed position ) 
  
- config / security : there's an Activate thermal limit on the UI but doesn't seems to be clickable or doing nothing. 

- config / Developper: The activate labs function button is not clickable. If the feature is not present on the firmware can this be hidden or grayed out  ? I'm not aware of this feature. 


Will do the code review but not before next week I think. 
Thanks a lot.

jeremypoulter:
On the limits, and generally the controls on the home screen. This seems a large area of miss understanding as to if these settings should persist or not. I think probably the easiest this to do is add a toggle to switch from the session settings to the persistent settings, but I am welcome to other suggestions. I think it is something we need to resolve however.

KipK:
Persistent limits were in configuration page with the idea they are set by an admin and can't be removed easily by users. 
I'm agree there's no user managements yet , it was just done like that in case of future implementation. 
Would make sense to at least add add an admin user though. 

A bit unrelated but I haven't noticed the limits had changed on fw. It seems it activate the charger now when you set a limit. I think it shouldn't. Before, we could set a limit and still keeps the timers enabling the charger. 

For the charge current setting, persistence could be achieved by just adding something like a checkbox that would add the "auto_release: false" flag to the override/claim .

jeremypoulter:
That is the other thing. The original intention was that the limits set on the home page are immediate so that issue was 'fixed'. The idea is I have just plugged in and I want to charge my car right now but then return to the configured operation, with the assumption that if you want a standing limit you would configure this in the settings, but there seems to be a use case where people want to set a session limit but not still pay attention to the schedule.

KipK:
@jeremypoulter , my use case was setting a limit before going to sleep, and let the schedule power on the charger when energy is cheaper. 

I wonder how we could cover booth world.
But anyway, if a user set a limit and want it to be enabeld at the same moment, he just have to enable the charger then. 
Charger will automatically set to Disable when the limit is reached so this was still covering this use case. 

Current implementation looks like we've removed a feature

jeremypoulter:
It doesn't work quite right, setting to active disables the schedule/divert until you set it back to automatic, so while the limit will work for that session when you disconnect/connect it will start charging with no limit.

I have a PHEV so I am charging a lot just after unexpected/unscheduled trips and the current behaviour sucks.

RAR:
Thank you for taking the time, I REALLY appreciated it!

> Hi, here is a first quick feedback on the UI based on UX/Features. I haven't checked the code yet.
> 
> * desktop version:
>   Seems the UI has not been refined yet for desktop.

This is correct, I followed the old UI in this regard and assumed it would mostly be used in phone browser or as a desktop web phone app thingie.  

> * I haven't tested a real charge yet, does all the metrics data from the previous UI still appears on screen ? I get only 2 fields for now displayed ( the energy consumtion ones ) but all the other ones are missing.
>   How is it handled ?

Yes!  When charging, if you have the tsdb monitoring enabled you get a nice graph with SOC (if data provided) and charge rate:

Screenshot 2026-06-09 at 7 29 20 AM

If you don't have the tsdb stuff enabled you get a simpler screen:

Screenshot 2026-06-09 at 7 30 37 AM

Ok, liimits - 

image

If provided the SOC bar shows by default for quick access.

If you'd like to set a limit off time or energy click the +limit in the "or limit by" bottom section:

Screenshot 2026-06-09 at 7 34 08 AM

a modal will pop up with those options.

Screenshot 2026-06-09 at 7 34 49 AM


>   It's pretty confusing having range & soc limit on an area, and other limits on a separated dialog.

I chose this way since - in my opinion - its more likely a user would be limiting on SOC than time or energy, so I made it more prominent than the other options.   I think its just different rather than confusing.   I went back and forth on the pop up, I tried to keep the page length down overall.  

>   Range and Soc are optional, related limits feature should only appears when the feature is enabled ( ie when there's data for soc & ev range ). I haven't checked if you've handled it like that already.

Yup, the bar isn't there if we don't have SOC.

> * time limit: the double slider displayed to set the time is not ux friendly. Also there's no feedback on the current position while you slide it on mobile.

I totally agree on the limit by time, the two sliders kinda suck, I'll look at finding a different way of doing it.

> * System Limits are missing from the "EVSE" config section
>   There  are 2 kind of limits, user limits and system ones. System ones when set appears on the main page but doesn't have the delete button.

This was a miss on my end - I'll correct it. 

> * the Shaper button should be more discreet. It's a security feature that mostly shouldn't be touched so easily. I wonder if we shouldn't take the opportunity to remove it or at least add an option to display it or not.

Let me know what you all would like here.    

> * settings/ network
>   please keep signal quality bar, only having dbm is not really user friendly.

Will do.

> * on mobile, the navigation bar at the bottom can appears above the screen until we hide the browser url bar. It should always be apparent on top of the scrollable content ( fixed position )

What browser/os?

> * config / security : there's an Activate thermal limit on the UI but doesn't seems to be clickable or doing nothing.

Its reliant on an underlying firmware version - I'll verify its working as expected and gate it. 

> * config / Developper: The activate labs function button is not clickable. If the feature is not present on the firmware can this be hidden or grayed out  ? I'm not aware of this feature.

huh.  Browser/os?  This seems to work fine on my end, like the thermal limit switch, guessing there is some underlying bug breaking those.

KipK:
@RAR 

> This is correct, I followed the old UI in this regard and assumed it would mostly be used in phone browser or as a desktop web phone app thingie.

Old UI still has the content and buttons not stretched on the whole width. It alos has few breakpoints

image


> Yes! When charging, if you have the tsdb monitoring enabled you get a nice graph with SOC (if data provided) and charge rate:

I like it, looks great. 👍 

> If provided the SOC bar shows by default for quick access.

I don't think having 2 different paths for limits is UX friendly. 
Also it takes a lot of space on screen for those who don't use it ( think SOC limit already handled on EV side , or users that use mostly time/energy limit ) 
There's no reason to priorise one feature above the others, we should keep  one simple path for all limits

But perhaps we could keep the bar for all the limit types. 
Works already easily for energy / range and soc , we could use it for time using the same steps as on previous UI

I did something like this on this HA card with a kind of exponential and stepped slider:

https://github.com/user-attachments/assets/e8349001-5841-495b-a9fe-490dadebf232



>> the Shaper button should be more discreet. It's a security feature that mostly shouldn't be touched so easily. I wonder if we shouldn't take the opportunity to remove it or at least add an option to display it or not.

> Let me know what you all would like here.

image


@jeremypoulter what do you think ? this main screen shaper button has no real sense, we can already disable the shaper service from the config if needed. But perhaps some users are using it ( wondering the use case ) 
 I know it's me who've added it but in my opinion taht was a bad idea and should be removed. 
 If we keep it, it should be a little button somewhere else but not centered as a hero feature 
 
 > What browser/os?
 
 I'm on android , chrome on pixel 9.
 
 
image


You have to scroll down until yiu don't see the url bar on the top to have the nav bar displayed fully. 

> huh. Browser/os? This seems to work fine on my end, like the thermal limit switch, guessing there is some underlying bug breaking those.

Ho I got it :), I haven't seen the switch above, was trying to click on the title. The switch is so dark as it's mostly invisible on my screen and missed it in dark mode.  Should adjust the colors perhaps to give them more visibility 
Was the same for thermal limit so 

image

RAR:
> Old UI still has the content and buttons not stretched on the whole width. It alos has few breakpoints
>
I'm happy to take some time to make it look better on desktop - I was really just following what was there, I use desktop a lot.

> I don't think having 2 different paths for limits is UX friendly. Also it takes a lot of space on screen for those who don't use it ( think SOC limit already handled on EV side , or users that use mostly time/energy limit ) There's no reason to priorise one feature above the others, we should keep one simple path for all limits

Perhaps making it an option in settings is the path?  With SOC I use the car SOC as the hard limit and the charger side as my soft limit.     Obviously everyone has different usage patterns.    Maybe options for what limits show up on the home page... that way I can ignore time/energy and can you an ignore that pesky SOC bar?  🤣 

> But perhaps we could keep the bar for all the limit types. Works already easily for energy / range and soc , we could use it for time using the same steps as on previous UI

I can look at this as well.  

> I'm on android , chrome on pixel 9.

AHHH ok, the min width is to wide for your phone screen so the horizontal scroll covers the buttons.
 
> Ho I got it :), I haven't seen the switch above, was trying to click on the title. The switch is so dark as it's mostly invisible on my screen and missed it in dark mode. Should adjust the colors perhaps to give them more visibility Was the same for thermal limit so

It's called DARK MODE for a reason!   😆   Lemme lighten that up a hair.

KipK:
Yes, for limit keeping the bar and adding the other limits there next to %/km would be a good way.

image


> AHHH ok, the min width is to wide for your phone screen so the horizontal scroll covers the buttons.

To solve the issue on all smartphone screen size, you should put the nav bar on a div with higher zindex and fixed position  bottom of screen like it's done on old ui.

RAR:
@KipK Pull main on https://github.com/RAR/openevse-gui-nightshift and take it for a ride.  I *think* I got most items covered.  

Screenshot 2026-06-09 at 7 30 36 PM
Screenshot 2026-06-09 at 7 25 54 PM
Screenshot 2026-06-09 at 7 25 37 PM
Screenshot 2026-06-09 at 7 24 42 PM

KipK:
thanks @RAR 

Looks good now on desktop 👍 
The limit section feels really good like that too.
I think you can add more time range for the time limit, 8h is a bit short for some use case ( those who charge with 6A )
Can you keep the same time range than the previous UI ? ( I think it was 24h ) 
No need to keep 15/30/45mn precision after few hours, just make the slider exponential like on the video in my comment above. 
That will solve the too high resolution UX issue on the slider as the time range will be increased. 


few things that are not addressed yet, but perhaps it's still on your todo list, just for reminder: 
- missing mouse pointer change on over 
- missing wifi signal bar

I would also add a route for when you switch from old UI to new UI after a firmware update. It points to a 404
On previous UI there's a default route for 404 that display a 404 msg and and has a link  to Home.

I still see few warnings at compile time that should be fixed to start clean :)

Really cool UI you did mate.

KipK:
Missed this one. The bottom nav bar is still  cropped after page load

Screenshot_20260610-121403

The whole page is scrollable ( height of the browser nav bar )  even when there's enough space 

Edit: and another little thing, when you've installed a dev firmware you've compiled yourself you can't click on the install button of the stable firmware:

Screenshot_20260610-122348

The web install has been broken for a year now anyway. ^^
I haven't followed when it has been broken , if you have time to give a look ( it's probably on fw side )
It seems the fw crash when clicking the web update.  ( At least on wifi_v1 module )

RAR:
Give main another shot, when you have a moment.

KipK:
Looks good to me. 

About the Boost feature here:

image

Is it somehting handled on UI side or a fw feature ? 

I wonder where those 15mn 30mn 1h comes from , I don't think those time frame are sufficient for EV charging or you have a fast DC charger at home  :)
Shouldn't it better to use the same slider than limit for this feature ?

RAR:
Break down is over here: https://github.com/OpenEVSE/openevse_esp32_firmware/issues/1029#issuecomment-4670277795

TLDR; its GUI side.  Figured it was good for a quick top up.

KipK:
I could try a charge with new UI today.

I see those 2 metrics missing on the main screen while charging:

- charge time left countdown ( when time to charge topic is present ) 
- shaper data ( available current ) 
- probably divert data too? 

I haven't checked yet is there a feedback for why it's disabled/enabled on main screen like on previous UI? 
Screenshot_20260611-091046

Edit:
Spotted another little thing, the charge current slider has no visual feedback on current value while you slide it on mobile. It only update visually the value when you release the slider. 

Screenshot_20260611-114022

RAR:
I do have some updates on main to address those, I've not fully tested them yet.

KipK:
I've compiled a fresh build with your changes. 
Looks good to me. 
Perhaps the display for the timer m:essage could be polished ( remove seconds ? ) 

image

RAR:
Done, checkout https://github.com/RAR/openevse-gui-nightshift/releases/tag/2026.24.0 or main

KipK (dismissed):
That's all good folks. Let's continue the polishing after merge if needed.

chris1howell (approved):

PR #1091, opened 2026-06-07
build: core-3.x (pioarduino/IDF5) two-core build foundation

RAR:
## Summary

This is the **foundation** for building OpenEVSE on the newer Arduino-ESP32 **core-3.x** (pioarduino, on ESP-IDF 5.x) **alongside** the existing core-2.x boards, from a single tree. It deliberately changes **no existing core-2 board** and adds **one** reference core-3 board so the build system can be reviewed in isolation.

It's the first in a planned series — later PRs (time-series energy history, Home Assistant integration, an ESP32-P4 target, etc.) stack on this.

## Why two cores

4MB boards must stay on core-2.x (`espressif32@6.12.0`) to keep the IDF4 image small enough for **dual-slot OTA**. Newer 16MB+ silicon needs the core-3.x toolchain (Arduino-ESP32 3.x / ESP-IDF 5.x). This PR lets both coexist.

## What's included

**Build system**
- `[common] platform_core3` anchor — pioarduino `platform-espressif32` 55.03.38-1.
- **`[env:openevse_wifi_v1_16mb]`** — the flagship `openevse_wifi_v1` feature set (PN532 + MCP9808 + NeoPixel) rebuilt on core-3 with a 16MB partition table. The reference target that proves the two-core build.
- **`[env:native]`** — a host-side doctest harness (`pio test -e native`).

**IDF5 / core-3 compile compatibility** (all version-guarded, so core-2 output is byte-for-byte unchanged)
- `LedManagerTask.cpp` — Arduino-ESP32 3.x unified `ledc*` API, gated on `ESP_ARDUINO_VERSION`.
- `certificates.cpp` — mbedTLS 3.x requires an RNG for key parsing; backed by the ESP HW RNG, gated on `MBEDTLS_VERSION_NUMBER`.
- `src/idf_component.yml` — declare `idf >= 5.1`.

**Tooling / CI / docs**
- `scripts/pio` — a transparent `pio` wrapper that routes core-3 builds to their own `PLATFORMIO_CORE_DIR`, so switching cores **locally** never re-downloads the shared-but-differently-pinned framework/toolchain packages. core-2 keeps the default `~/.platformio`, so bare `pio`, CI, and IDE integrations are unaffected. (CI is immune anyway — each job is a fresh runner.)
- `scripts/auto_fw_version.py` — fall back to `GITHUB_SHA`/`GITHUB_REF_NAME` when `git` is unavailable.
[...]

PR #1094, opened 2026-06-07
feat: energy logging + temperature throttle + time-series history (tsdb)

RAR:
## Summary

Energy logging, temperature throttle, and an optional **time-series history backend (tsdb)** — the full energy feature set in one PR.

> **Stacked on #1091** (the core-3 two-core build foundation). The tsdb half enables on the core-3 `openevse_wifi_v1_16mb` board, so this builds on #1091. Until #1091 merges, this PR's diff includes those foundation commits; it rebases clean once #1091 lands.

## Energy logging + temperature throttle

Backport of the energy feature originally proposed in **#1083** (closed). Always-on, all boards.

- Per-minute energy samples with daily / monthly / annual aggregation, served at `/energy/{raw,daily,monthly,annual}` and published over MQTT.
- Temperature throttle derates charge current as controller temperature rises; integrated into `EvseManager`/`EvseMonitor`. New `CONFIG_TEMP_THROTTLE` flag (`"tte"`, bit 27).
- Fits the 4MB boards — `openevse_wifi_v1` builds at **94.8%** flash.

## Time-series history (tsdb) — core-3 / `ENABLE_TSDB` only

- Vendors **esp_tsdb** as an IDF component (`components/esp_tsdb`).
- `TsdbEnergyLogger`: ~100 days of minute-resolution energy history in flash, replacing the legacy 6-hour `EnergyLogger` when `ENABLE_TSDB` is set.
- `/energy/{raw,daily,weekly,monthly,annual}` served from tsdb range/aggregate queries. The legacy `web_server_energy_logger.cpp` is compiled out under `ENABLE_TSDB` (its handlers + the `energyLogger` global are replaced by the tsdb versions).
- `tsdb_sample` pure int16 scaling unit with native doctest coverage.
- Enabled on `openevse_wifi_v1_16mb`; includes a fix for an esp_tsdb IRAM-placement crash on xtensa.

## Validation

- `openevse_wifi_v1_16mb` (core-3, `ENABLE_TSDB`): builds clean.
- `openevse_wifi_v1` (core-2, tsdb compiled out): builds clean, 94.8% flash.
- `pio test -e native`: passes (energy smoke + `tsdb_sample`).

---


chris1howell (dismissed):

chris1howell (dismissed):

chris1howell (approved):

KipK:
@RAR is it normal it still logging while not charging ? 
I see temp and soc history but no charge here
image

RAR:
It is intentional.  Can likely add a switch to disable if its not wanted.

KipK:
> It is intentional. Can likely add a switch to disable if its not wanted.

Should be disabled by default I think. I'm a bit afraid of the flash lifetime.

chris1howell:
I would say do not record SOC after the charge ends. When we tried this with a direct Tesla code it kept the vehicle awake which used more range. It was also strange to see SOC decline as the vehicle drove around.

Temperature however is useful after the charge. It is slow moving so the rate could be slowed to every 5 minutes or 10 minutes. Current does not need to be logged when not charging.

RAR:
https://github.com/OpenEVSE/openevse_esp32_firmware/pull/1109

PR #1096, opened 2026-06-08
Add home-battery display feeds (MQTT + POST /status)

RAR:
## Summary

Adds **home / powerwall battery** state to `/status` as a display-only feed. The EVSE does not derive these values itself — they can be fed over **MQTT** or pushed via **`POST /status`**:

| Field | Meaning |
|---|---|
| `home_battery_soc` | Home / powerwall battery state of charge (%) |
| `home_battery_power` | Home / powerwall battery power (W; negative = discharging) |

This lets integrations (e.g. Home Assistant, solar/battery systems) that already know these values reflect them on the EVSE and its display, without the EVSE polling anything.

## Design

- New standalone store `src/home_battery.{h,cpp}` holds the last received value and **omits each field from `/status` until it has been received at least once** (no misleading zeros).
- **MQTT**: new optional topics `mqtt_home_battery_soc` / `mqtt_home_battery_power`. These are display-only with no source arbitration, so they are gated only on the topic being configured.
- **`POST /status`**: `home_battery_soc` / `home_battery_power` are an explicit override like the existing `solar` / `grid_ie` pushes.

## Tests

Adds a self-contained `[env:native]` doctest environment covering the store (valid-flag gating and `/status` emission). Run with `pio test -e native`.

> Note: this `[env:native]` block overlaps the native test harness introduced in the core-3 two-core build PR (#1091). If both land, the two env blocks should be reconciled at merge — they are otherwise independent.

## Validation

- `pio test -e native` — passes.
- `pio run -e openevse_wifi_v1` — builds clean.

## Notes

Branched off `master`. Pure additive change. (Vehicle plugged / charging-state feeds were dropped from this PR — they're being folded into a larger multi-vehicle effort.)

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -563,10 +563,14 @@ board_build.f_flash = 80000000L
 # ───────────────────────── host-side unit test harness ─────────────────────
 # Runs pure, framework-free logic on the build host. Run with: pio test -e native
 # Feature PRs add their own doctest suites under test/; this env just hosts them.
+# home_battery.cpp is a pure store, so it is compiled into the suites that need it
+# (the smoke test is dependency-free and tolerates the extra object).
 [env:native]
 platform = native
 framework =
 test_framework = doctest
+test_build_src = true
+build_src_filter = -<*> +<home_battery.cpp>
 build_flags = -std=gnu++17
 lib_deps = bblanchon/ArduinoJson@6.20.1
 extra_scripts =
--- a/src/app_config.cpp
+++ b/src/app_config.cpp
@@ -76,6 +76,8 @@ String mqtt_vehicle_soc;
 String mqtt_vehicle_range;
 String mqtt_vehicle_eta;
 String mqtt_vehicle_charge_limit;
+String mqtt_home_battery_soc;
+String mqtt_home_battery_power;
 String mqtt_announce_topic;
 
 // OCPP 1.6 Settings
@@ -194,6 +196,8 @@ ConfigOpt *opts[] =
   new ConfigOptDefinition<String>(mqtt_vehicle_range, "", "mqtt_vehicle_range", "mr"),
   new ConfigOptDefinition<String>(mqtt_vehicle_eta, "", "mqtt_vehicle_eta", "met"),
   new ConfigOptDefinition<String>(mqtt_vehicle_charge_limit, "", "mqtt_vehicle_charge_limit", "mcl"),
+  new ConfigOptDefinition<String>(mqtt_home_battery_soc, "", "mqtt_home_battery_soc", "mhs"),
+  new ConfigOptDefinition<String>(mqtt_home_battery_power, "", "mqtt_home_battery_power", "mhp"),
   new ConfigOptDefinition<String>(mqtt_announce_topic, "openevse/announce/" + ESPAL.getShortId(), "mqtt_announce_topic", "ma"),
 
 // OCPP 1.6 Settings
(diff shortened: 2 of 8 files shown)


jeremypoulter:
Thanks, I am however confused about the vehicle_plugged and vehicle_charging_state. The EVSE knows if a vehicle is plugged in and charging so why would you need to provide from an external source?

RAR:
> Thanks, I am however confused about the vehicle_plugged and vehicle_charging_state. The EVSE knows if a vehicle is plugged in and charging so why would you need to provide from an external source? 

We have two EVs, if I'm providing SOC it's only for one of them right now.  I want to ignore SOC in that case because it's not applicable to the current charging session.  Also could come up with guest vehicles, etc.

jeremypoulter:
> > Thanks, I am however confused about the vehicle_plugged and vehicle_charging_state. The EVSE knows if a vehicle is plugged in and charging so why would you need to provide from an external source? 
> 
> We have two EVs, if I'm providing SOC it's only for one of them right now.  I want to ignore SOC in that case because it's not applicable to the current charging session.  Also could come up with guest vehicles, etc.  

I am not quite sure I understand the logic here, would it not be better to handle the case properly by adding support for multiple vehicles? Eg add the ability to setup multiple vehicle feeds and an API to select the current vehicle? Not sure it makes sense to build a way just to hide this.

RAR:
Even with multi vehicle I'd want to know which is plugged since then I can pick the correct one without the user, or say its a guest vehicle if none, or prompt the user if both are.   Yeah, multi-vehicle crossed my mind, but its a much larger lift and I'm already changing a ton at once with all these PRs.  🤣

KipK:
I feel that's the kind of thing that should be done on HA side. If we have an api to switch vehicles ( + guest mode ), then it has more sense to push the vehicle switch from HA than using logic on OpenEVSE ain't it ?

RAR:
Where to draw that line is always a fun question.  You could make the same argument for shaper and self production.

jeremypoulter:
Thinking about it a bit more, let's split it out from this PR. I think the connected state should be associated with the vehicle feeds directly. If connect, pay attention to the feeds if not ignore the vehicle feeds and any associated limits. We can then expand this to multiple vehicles later. I think that makes sense as this would apply to multiple cases, not just Home Assistant. 

I am still however not sure what the use case for vehicle_charging_state would be that can not be derived from the existing state. If it is just a display this text then that is maybe ok, but let's make it that...

RAR:
NGL I had to go look what I was using it for. 

```
    - charging → Charging
    - complete / charged / full → Complete
    - idle / stopped / disconnected → Idle
```

The only thing we would gain is "complete" by adding that.    With that being said... Yeah I can yank that.

chris1howell:
Fun topic, if the vehicle feeds are coming from MQTT it would not necessarily mean Home Assistant is involved. I am looking forward to Home Battery feeds. I also like the idea that you could have the station figure out which vehicle is connected and charging.
________________________________

RAR:
> Thinking about it a bit more, let's split it out from this PR. I think the connected state should be associated with the vehicle feeds directly. If connect, pay attention to the feeds if not ignore the vehicle feeds and any associated limits. We can then expand this to multiple vehicles later. I think that makes sense as this would apply to multiple cases, not just Home Assistant.
> 
> I am still however not sure what the use case for vehicle_charging_state would be that can not be derived from the existing state. If it is just a display this text then that is maybe ok, but let's make it that...

Yanked it out, this just covers home battery now - when we're ready for multi-vehicle we can look into it again.

chris1howell (dismissed):
Looks great. I would love to see a compound timer, for example M, T, W, T, F 900AM - 600PM charge  with excess solar if Home battery SOC > 60%.

RAR:
Resolved the merge conflicts after the other PRs got merged.

chris1howell (approved):

PR #1097, opened 2026-06-08
Add vehicle_charge_limit to the vehicle data set (/status + POST)

RAR:
## Summary

Adds a `vehicle_charge_limit` value to the vehicle data set, mirroring the existing vehicle **state of charge / range / ETA** pattern. This is the vehicle's target charge percentage (e.g. "charge to 80%"), as reported by the vehicle or a connected integration.

## Changes

- `EvseManager` gains `_vehicleChargeLimit` plus `getVehicleChargeLimit()`, `isVehicleChargeLimitValid()` and `setVehicleChargeLimit()`, tracked by a new `EVSE_VEHICLE_CHARGE_LIMIT` validity bit — exactly parallel to SoC/range/eta.
- `/status` emits `vehicle_charge_limit` once a value has been received (omitted until then, like the other vehicle fields).
- `POST /status` accepts `vehicle_charge_limit` under the same `vehicle_data_src == VEHICLE_DATA_SRC_HTTP` gating as `battery_level` / `battery_range` / `time_to_full_charge`.

## Validation

- `pio run -e openevse_wifi_v1` — builds clean (flash 93.8%).

## Notes

Branched off `master`. Small additive change (29 insertions). A follow-up PR adds an MQTT topic for the same value (`mqtt_vehicle_charge_limit`); it is stacked on this branch.

### DIFF
--- a/src/evse_man.cpp
+++ b/src/evse_man.cpp
@@ -133,7 +133,8 @@ EvseManager::EvseManager(Stream &port, EventLog &eventLog) :
   _vehicleLastUpdated(0),
   _vehicleStateOfCharge(0),
   _vehicleRange(0),
-  _vehicleEta(0)
+  _vehicleEta(0),
+  _vehicleChargeLimit(0)
 {
 }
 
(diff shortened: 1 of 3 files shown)


chris1howell (approved):

jeremypoulter:
Can you clarify what the use case is for this? We already have an API for limiting the charge based on the SOC

RAR:
> Can you clarify what the use case is for this? We already have an API for limiting the charge based on the SOC

Hey!  This is part of the SOC charge bar.  If we have SOC and this field it allows for a better user experience by giving more data in the UI about what will actually happen when they charge.

Screenshot 2026-06-09 at 7 21 17 AM

This puts a hard stop on the bar so we don't assume we're going to get to 80% if we set SOC charging to 80% when the car is set to stop at a lower amount.

jeremypoulter:
Ah ok, so this is to represent the limit configured on the vehicle rather than the EVSE. Ok that makes sense. I assume this would fallback to the EVSE limit is not configured

RAR:
> Ah ok, so this is to represent the limit configured on the vehicle rather than the EVSE. Ok that makes sense. I assume this would fallback to the EVSE limit is not configured

Yup, it ignores all that if its not available.

PR #1100, opened 2026-06-09
Fix heartbeat supervision resync after WiFi module reboot

### LINKED ISSUE #540: Pilot current drops to 6A for some max current settings
Firmware 7.1.3, openevse_wifi_v1 just upgraded to 4.1.7
Some settings (32A, 33A) of the max current pulldown are limiting the maximum current to 6A, while others (31A, 34A) are working as expected.  When the limiting occurrs, the debug window shows {"pilot":6}

### LINKED ISSUE #953: Charging at 6A instead of 40A
The OpenEVSE was stuck at 6A (or 6.2A) charging rather than using the full 40A capacity. (See the sequence, below.)

(This seems unrelated to #812 because it occurred with and without any OpenEVSE timers, and in both cases with no other charge-limiting features besides system max 40A.)

**Hardware:** Replacement Electronics circuit board for Juicebox v2
**Firmware:** OpenEVSE 8.2.3.CGMI, OpenEVSE Wifi v5.1.2
**Connections:** On Wifi. No monitoring. No MQTT. No OCPP. No OhmConnect. No Home Assistant. No RFID.
**Settings:** No charge limits. Default state = Active, Three-phase = No, Service Level = 2, Pause Status = off, HTTP authentication = Disabled, Shaper = Disabled.

First day sequence (to the best of my memory):
1. Install the OpenEVSE circuit board. -> Charge rate defaults to 40A since that's the hardware's capacity and it's correct for the 50A circuit breaker.
2. Set up Wifi and time zone.
3. Set timers to disable charging during 4-9pm peak TOU hours.
4. Charge the Tesla. The Tesla is also set to wait out 4-9pm and it always has its own SoC charge limit. -> It started charging at 40A.
5. -> Charging paused at 4pm.
6. -> At ≈8:58pm the EVSE clicked and its LED switched from Cyan to Yellow.
7. -> At 9pm charging resumed but **at 6A** instead of 40A. The OpenEVSE and the Tesla agree on the actual charging current.
8. Toggle OpenEVSE to Disabled, then toggle it back to Auto. -> That didn't fix the charging rate. It finished charging overnight at 6A.

Second day sequence/experiment:
1. Delete the OpenEVSE charge timers.
2. Set Tesla to charge during 8pm - 4pm for this test, and with its own SoC max, as always.
3. Plug in the Tesla during 8-4pm. -> It waited until 8pm, then started charging at 6A.
4. Unplug and replug in the Tesla. -> Again charging at 6A.
5. The Tesla UI could pick either 5A or 6A charging current. -> The OpenEVSE must be signaling a 6A limit.
6. Disable Charge.
7. Restart OpenEVSE firmware. -> Charging at 6A.
8. Restart OpenEVSE Wifi firmware. -> Charging at 6A.
9. Reset OpenEVSE (on the Firmware screen).
10. Reconfigure OpenEVSE Wifi and time zone. -> Charging at 6A.
11. Disable Charge, then toggle it back to AUTO. -> Charging at 40A!!

chris1howell:
## Summary

Fixes #953 
Fixes #540 

- **Bug:** After the WiFi module reboots, heartbeat supervision never resyncs — the OpenEVSE Controller stays in triggered state with current throttled to the heartbeat fallback (6 A) indefinitely.
- **Root cause:** Two missing code paths in `evse_monitor.cpp`:
  1. On boot, if the OpenEVSE Controller had heartbeat triggered (`triggered == 2`) while the WiFi was away, no immediate ack pulse was sent, so ampacity was never restored.
  2. In `loop()`, when `_heartbeat` is false there was no retry of `heartbeatEnable`, leaving heartbeat permanently disabled until the next full reboot.

## Changes

- `evseBoot()`: check `triggered == 2` in the `heartbeatEnable` callback and send an immediate `heartbeatPulse` to restore ampacity.
- `loop()`: add `else if` retry — when `_heartbeat` is false, periodically re-issue `heartbeatEnable` so the WiFi module resyncs automatically after any reboot or failed enable.

## Testing

- Flash WiFi module, verify heartbeat supervision is established (current not throttled)
- Reboot WiFi module while vehicle is connected and charging; confirm current restores within one `EVSE_MONITOR_STATE_TIME` cycle
- Simulate delayed EVSE boot (WiFi up first): confirm heartbeat enable retries and succeeds

### DIFF
--- a/src/evse_monitor.cpp
+++ b/src/evse_monitor.cpp
@@ -251,6 +251,14 @@ void EvseMonitor::evseBoot(const char *firmware)
 
   _openevse.heartbeatEnable(EVSE_HEATBEAT_INTERVAL, EVSE_HEARTBEAT_CURRENT, [this](int ret, int interval, int current, int triggered) {
     _heartbeat = RAPI_RESPONSE_OK == ret;
+    // If heartbeat was triggered while WiFi module was rebooting, ack immediately to restore ampacity
+    if (_heartbeat && 2 == triggered) {
+      _openevse.heartbeatPulse([](int ret) {
+        if (RAPI_RESPONSE_OK != ret) {
+          DEBUG_PORT.println("Heartbeat ack failed");
+        }
+      });
+    }
   });
 }
 
@@ -330,6 +338,20 @@ unsigned long EvseMonitor::loop(MicroTasks::WakeReason reason)
     {
       if(RAPI_RESPONSE_OK != ret) {
         DEBUG_PORT.println("Heartbeat failed");
+      }
+    });
+  }
+  else if(0 == _count % EVSE_MONITOR_STATE_TIME)
+  {
+    // Heartbeat enable failed or was never set; retry periodically so WiFi module reboot resyncs
+    _openevse.heartbeatEnable(EVSE_HEATBEAT_INTERVAL, EVSE_HEARTBEAT_CURRENT, [this](int ret, int interval, int current, int triggered) {
+      _heartbeat = RAPI_RESPONSE_OK == ret;
+      if (_heartbeat && 2 == triggered) {
+        _openevse.heartbeatPulse([](int ret) {
+          if (RAPI_RESPONSE_OK != ret) {
+            DEBUG_PORT.println("Heartbeat ack failed");
+          }
+        });
       }
     });
   }


chris1howell:
@jeremypoulter Can you look at this bug fix? It is really common for stations to get stuck at the 6A default of Heartbeat Supervision.

PR #1109, opened 2026-06-13
feat(tsdb): throttle and trim energy logging while idle

RAR:
## Summary

Addresses review feedback on #1094 about the TSDB energy logger recording full samples (current, SoC, pilot, voltage) every minute even when the vehicle is not charging — visible in KipK's screenshot of temp/SoC history with no charge, and flagged by chris1howell.

When **not charging** (`!isCharging()`), the TSDB logger now:

- **Drops SoC** (`soc = -1`). Recording SoC after a charge ends keeps the vehicle awake (extra range loss, per chris1howell's Tesla experience) and makes the history show SoC drifting as the car drives around.
- **Drops charge current / power / pilot / voltage** — meaningless when idle.
- **Keeps temperature** (slow-moving and useful idle or not) and keeps accumulating the energy delta (~0 while idle) so cumulative daily/monthly/annual rollups stay exact.
- **Throttles the write cadence to 1/5 min** (`TSDB_ENERGY_IDLE_SAMPLE_MS`, build-overridable) instead of 1/min, cutting idle flash wear ~5× — addresses KipK's flash-lifetime concern.

While charging, behavior is unchanged: 1/min full samples.

## Validation

- `openevse_wifi_v1_16mb` (core-3, `ENABLE_TSDB`): builds clean.
- Flashed to a real 16 MB ESP32 and confirmed live via `/energy/raw` while idle:
  - new samples ~301 s apart (5‑min idle cadence)
  - `v:0, p:0, pilot:0, soc:-1` — current/SoC/pilot/volts suppressed
  - temperature/energy slots retained (read 0 here only because the bench module has no temp sensor)
- `tsdb_ready:1 / tsdb_err:0` on hardware.

Charging-path cadence (return to 1/min with real current/SoC) still wants a confirm on a full rig with a vehicle.

### DIFF
--- a/src/tsdb_energy_logger.cpp
+++ b/src/tsdb_energy_logger.cpp
@@ -236,7 +236,12 @@ void TsdbEnergyLogger::rollup_yesterday() {
 }
 
 unsigned long TsdbEnergyLogger::loop(MicroTasks::WakeReason) {
+  // Throttle to the idle cadence whenever we are not actively charging.
+  unsigned long next_ms = TSDB_ENERGY_SAMPLE_MS;
   if (_ready && _evse) {
+    bool charging = _evse->isCharging();
+    if (!charging) next_ms = TSDB_ENERGY_IDLE_SAMPLE_MS;
+
     // Advance the energy baseline every wake (even when we skip the write below),
     // so deltas stay honest once logging resumes. A session reset to 0 on vehicle
     // unplug yields a small positive delta (cur_wh), not a negative one. (Lossy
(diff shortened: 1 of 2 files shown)


chris1howell (approved):

KipK:
even with 5mn logging as it's permanent it makes 105120 records per year on the flash. 
What is the estimated flash lifetime at such rate ? I have no idea :)
Would be cool to have an sd card slot for logging btw.

chris1howell:
@KipK ESP32 Flash can handle between 10,000 and 100,000 flash/erase cycles. Even if the logging data is overwritten once a year, that is a lot of years.

The time series database optimizes further, they do not store repeating values as raw, repetitive data. They use highly advanced compression algorithms (like XOR or delta encoding) to represent constant values using only a fraction of a byte, meaning they take up virtually no extra space.

PR #1110, opened 2026-06-14
fix(ota): finalize completed HTTP downloads before socket close

KipK:
## Summary

Fix URL-based HTTP OTA updates that download the firmware to 100% but never finalize, reboot, or install the new image.

This is a follow-up to #1108 and #1104

ArduinoMongoose PR https://github.com/jeremypoulter/ArduinoMongoose/pull/53 is needed too. 

platformio.ini now points to https://github.com/jeremypoulter/ArduinoMongoose@0.0.26 

---

## Problem

The OTA download progress is updated while HTTP body chunks are written to the inactive OTA partition. Reaching 100% therefore only means that all expected bytes were written — it does not mean that the OTA image has been finalized or selected as the next boot partition.

Finalization was performed from the HTTP client's `onClose` callback:

```cpp
if (Update.isRunning()) {
  if (http_update_end()) {
    restart_system();
  }
}
```

This made successful installation dependent on the socket-close event occurring after the last HTTP body chunk.

When that callback did not complete the expected path:
- the GUI remained at 100%
- `Update.end()` was not called
- the new partition was not activated
- no reboot was scheduled

The previous code also called `Update.end(true)`, which permits finalization even when bytes remain. A prematurely closed connection could therefore attempt to finalize an **incomplete image**.

Finally, errors occurring after progress reached 100% were not clearly propagated to the GUI, making a finalization failure indistinguishable from a stalled update.

---

## Resolution

The update is now finalized from `MG_EV_HTTP_REPLY`, which indicates that the complete HTTP response has been received.

**New flow:**

1. Stream each response body chunk into the OTA partition.
2. Track the number of bytes successfully written.
3. On receipt of the complete HTTP response, verify that the downloaded size matches `Content-Length`.
4. Call `Update.end(false)` so incomplete images are rejected.
5. Activate the new boot partition.
6. Emit the `ota: completed` event.
7. Schedule the system reboot.

The `onClose` callback is now only used to detect an interrupted transfer. If the socket closes before the complete response is received, the update is aborted and reported as failed.

---

## Additional Changes

- Reset OTA progress state when starting a new update.
- Prevent progress calculations from exceeding 100%.
- Add a dedicated incomplete-download error code.
- Include `ota_error` in failure events sent to the GUI.
- Log the underlying Arduino `Update` error when finalization fails.
- Preserve redirect handling for GitHub release assets.

---

## Result

A complete HTTP download no longer waits for a later socket-close callback before being installed. Once the response is fully received and validated, the firmware is finalized and the device reboots into the new image.

Interrupted or invalid downloads now fail explicitly instead of leaving the GUI stuck at 100%.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -35,6 +35,7 @@ default_envs = openevse_wifi_v1
 platform_core3 = https://github.com/pioarduino/platform-espressif32/releases/download/55.03.38-1/platform-espressif32.zip
 lib_deps =
   bblanchon/ArduinoJson@6.20.1
+  # IDF5 / Arduino-core-3 compatibility fixes (pending upstream into jeremypoulter/*):
   jeremypoulter/ArduinoMongoose@0.0.26
   jeremypoulter/Micro Debug@0.0.5
   jeremypoulter/ConfigJson@0.0.6
--- a/platformio.ini
+++ b/platformio.ini
@@ -35,7 +35,6 @@ default_envs = openevse_wifi_v1
 platform_core3 = https://github.com/pioarduino/platform-espressif32/releases/download/55.03.38-1/platform-espressif32.zip
 lib_deps =
   bblanchon/ArduinoJson@6.20.1
-  # IDF5 / Arduino-core-3 compatibility fixes (pending upstream into jeremypoulter/*):
   jeremypoulter/ArduinoMongoose@0.0.26
   jeremypoulter/Micro Debug@0.0.5
   jeremypoulter/ConfigJson@0.0.6
--- a/src/http_update.cpp
+++ b/src/http_update.cpp
@@ -15,6 +15,15 @@ static int lastPercent = -1;
 static size_t update_total_size = 0;
 static size_t update_position = 0;
 
+struct HttpUpdateRequestState
+{
+  bool startedUpdate = false;
+  bool updateComplete = false;
+  bool responseComplete = false;
+  bool redirected = false;
+  bool errorReported = false;
+};
+
 bool http_update_from_url(String url,
   std::function<void(size_t complete, size_t total)> progress,
   std::function<void(int)> success,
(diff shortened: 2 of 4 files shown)


platformio.ini
```diff
@@ -36,7 +36,7 @@ platform_core3 = https://github.com/pioarduino/platform-espressif32/releases/dow
 lib_deps =
   bblanchon/ArduinoJson@6.20.1
   # IDF5 / Arduino-core-3 compatibility fixes (pending upstream into jeremypoulter/*):
-  https://github.com/RAR/ArduinoMongoose.git#cf237c4b225a5ed11a836aa53047ede1d5509336
+  https://github.com/jeremypoulter/ArduinoMongoose.git@0.0.26
```
jeremypoulter:
```suggestion
  jeremypoulter/ArduinoMongoose.git@0.0.26
```

jeremypoulter (changes requested):

platformio.ini
KipK: seems it doesn't find the lib like that seeing the failed checks

jeremypoulter: Ah you need to drop the .got as well sorry, see the other entries

jeremypoulter: But also rebase, the native build had the change as well

KipK: lol haven't noticed the .git, was on mobile. It's good now

KipK:
openevse_native build fails. I haven't followed yet how it's done, @jeremypoulter  may I let you fix this part ?

jeremypoulter:
Yeah new APIs from the Update are in use, need to add them to EpoxyDuino. Will fix shortly

jeremypoulter:
Should be fixed now

platformio.ini
```diff
@@ -35,6 +35,7 @@ default_envs = openevse_wifi_v1
 platform_core3 = https://github.com/pioarduino/platform-espressif32/releases/download/55.03.38-1/platform-espressif32.zip
 lib_deps =
   bblanchon/ArduinoJson@6.20.1
+  # IDF5 / Arduino-core-3 compatibility fixes (pending upstream into jeremypoulter/*):
```
jeremypoulter: Lets drop this comment, no longer relavent
```suggestion
```

jeremypoulter (changes requested):

jeremypoulter (approved):

PR #1112, opened 2026-06-15
Charge Manager UI and firmware 

chris1howell:
# Charge Manager — Change Summary

Branch: `charge_manager`  
Repos: `openevse_esp32_firmware` · `openevse-gui-nightshift`

---

## Overview

Replaces the raw Schedule page with a **Charge Manager** that splits policy into two sections:

- **Always Active** — persistent global features (Limit, Solar Divert, Grid Shaping, RFID, OCPP)
- **Scheduled Rules** — windowed timer pairs with day picker, start/stop time, action, and optional limit

Screenshot 2026-06-14 140211
---

### Next-day stop time day shifting
Select only Start Days of Week, when a rule's stop time is earlier than its start time (wraps past midnight), the stop timer is written with days shifted forward by one (`Mon–Fri` → `Tue–Sat`). The `timersToRules` two-phase pairing algorithm handles this in reverse: pass 1 looks for same-day stops (later time), pass 2 looks for stops whose day-set equals `shiftDaysForward(start.days)`.

image

`[{"id":1,"state":"active","time":"22:00:00","days":["monday","tuesday","wednesday","thursday","friday"],"feature":"current","feature_value":24},{"id":2,"state":"disabled","time":"06:00:00","days":["tuesday","wednesday","thursday","friday","saturday"]}]`


chris1howell:
This one should be fully functional and ready to try out.

image

RAR (commented):
Did a deep pass over the firmware side of this (skipped the generated web_static blobs). Overall the Charge Manager core is solid — I traced the claim-priority model through and it works: unauthenticated RFID claims Disabled @ 1030 over the schedule's Timer @ 100, so a timer-RFID window genuinely gates charging until a badge; divert/shaper at Limit 1100 correctly outrank the schedule claim. The `getDuration`/`getDelay` signed-math fixes are correct, all the `Event` constructors delegate to the initializing one so there's no uninitialized feature/limit path, and converting the schedule save from silent truncation to explicit refusal is strictly better than what master does. The `==`→`=` in `initDivertType()` is a great catch — that one's been silently breaking the divert-type migration on master.

A few things I'd flag before merge, in order of importance:

**1. The `main.cpp` gate fix doesn't revive what the commit message says it does.** `handleRapiRead()` (input.cpp:168) and `import_timers()` (legacy_support.cpp:9) both call through the *global* `OpenEVSE` object — the same sender-less singleton the gate fix routes around. The lib constructs it with `_sender(NULL)`, nothing ever calls `OpenEVSE.begin()`, and `getTimer()`/`getTime()` null-guard and return silently. So both functions now *run* but still do nothing. Same disease affects `time_man.cpp:202/339` (`setTime`/`getTime` — EVSE RTC sync via this path is dead), `ohm.cpp:63/77`, and `input.cpp:172`. The fix is probably routing them through `evse.getOpenEVSE()` — but note that actually reviving `import_timers` resurrects the legacy import-and-`clearTimer` behavior, which now interacts with Charge Manager rules. That deserves a deliberate decision rather than an accidental revival, so it might be cleanest to drop the dead calls here and handle the sender-less-global cleanup as its own PR.

**2. CM-B2 is only half-fixed — the API deserialize paths still cap at 1024 bytes.** The boot path now sizes from `file.size()*2`, but `deserialize(const char*)`, `deserialize(Stream&)`, and the per-event variant (scheduler.cpp:648-725) all still use a hard-coded `DynamicJsonDocument doc(1024)`. Parsing from `const char*` copies keys *and* values into the pool, so an event with feature/limit fields costs ~350-400 B in-doc — a bulk `POST /schedule` overflows at roughly two events. The two-event rule-pair example in the PR description is already near the ceiling. Anything posting a full schedule (config restore, API clients) gets a 400. Related: `commit()`'s fixed 4096 doc means schedules past ~15 events refuse to persist (400, RAM-only, reverts on reboot) — with `SCHEDULER_MAX_EVENTS = 50` the capacity should probably scale with event count everywhere rather than stay fixed.

**3. Timer-RFID fails open and can't self-recover.** `readerPresent()` is a one-shot I2C ACK probe in `PN532::begin()`. If it false-negatives at boot (worth verifying a PN532 in LowVbat ACKs its address before wakeup), `applyFeature(RFID)` skips enforcement silently — and since `setTimerRequired()` is then never called, scanning never starts, `status` never goes ACTIVE, and `readerPresent()` stays false until reboot. For an access-control feature I'd want at least an event the UI surfaces when the feature is skipped, and ideally a re-probe at window start.

**4. Shaper timer-control details.** `setTimerEnabled()` uses the global `evse` rather than `_evse` (current_shaper.cpp:165) — inconsistent with the rest of the class including the null-guard this PR adds to `notifyConfigChanged()`. Semantically: if the shaper is config-enabled (always-on, claiming at Safety 5000) and a shaper window activates, `_timer_controlled` demotes subsequent claims to Limit 1100 — a scheduled window lowering the priority of an already-always-on shaper looks unintended. Maybe timer control should only apply when the shaper isn't config-enabled?

**5. The divert enter/exit re-claim logic wants divert_sim coverage.** The already-in-Eco choreography in `setTimerDivertActive()` took a few iterations to converge and re-issues Active with a possibly-stale `_charge_rate` until the next MQTT `update_state()`. It reads correct now, but these are exactly the cases the divert_sim harness could pin down: enter window while already Eco+Active, exit window with Eco still configured, `divert_enabled` flip mid-window. Same for the scheduler wrap fixes — `getDuration`/`getDelay` are pure functions that would be cheap to unit test (Sunday-only 1-minute window, event-already-passed-today).

**6. Scheduled "OCPP" is a silent placeholder.** `applyFeature(OCPP)` is a no-op but the GUI offers it and it serializes — users will schedule an OCPP window and observe nothing. Cost limits log and refuse activation; OCPP quietly "succeeds". Might be worth hiding in the GUI or rejecting server-side until it does something.

Last, a structural thought: the fs-corruption guards, temp-throttle default, heartbeat semantics, `evse_connected` fixes, and the divert `=` fix are all independent of Charge Manager and individually merge-worthy — if review of the scheduler feature takes a while, it'd be a shame for those to wait on it. And the gui-nightshift pin (`371607e3` on `charge_manager`) will need to be on the GUI repo's default branch before this merges so fresh clones don't break.

Happy to help with any of these — particularly the divert_sim cases, since I've been in that code recently.

chris1howell:
Thank you for the Review @RAR. Here are the changes https://github.com/OpenEVSE/openevse_esp32_firmware/pull/1112/changes/eea2509231bedb23e377f2566023d43bbd4813a8

1. Inert main.cpp gate fix, fixed by removal. Confirmed both callees dead-end on the sender-less global (OpenEVSE.getTime/getTimer null-guard and return without invoking callbacks). Removed the whole first-connect block, the rapi_read flag, and the legacy_support.h include from main.cpp.

2. CM-B2 half-fixed, now fixed across the whole JSON surface. It was worse than flagged: GET /schedule used a ~1.6 KB doc, so the GUI itself would read back a truncated rule list. Added Scheduler::scheduleJsonCapacity() (event-count-scaled, ~384 B/event) used by commit(), serialize(String&), serialize(Stream&), and handleScheduleGet; serialize(DynamicJsonDocument&) now returns false on overflow instead of silently dropping events. The deserialize paths (const char*, Stream&, and per-event) now size from the input (strlen*2 / available()*2, floor 4 KB) and reject overflowed parses — bulk POST /schedule, MQTT schedule/set, and config restore all handle 50 events now.

3. Timer-RFID fails open silently. Added probeReader() through the RfidReader interface: PN532 re-runs the I2C ACK probe (skipped when already ACTIVE) and refreshes the cached result, begin() now reuses it. applyFeature(RFID) re-probes at window start and, if the reader is genuinely absent, emits a schedule_feature_skipped: "rfid" event over WS/MQTT instead of failing open silently. The LowVbat question (does a sleeping PN532 ACK its address?) still deserves a bench check, but the re-probe means a boot false-negative now self-recovers at the next window.

4. Shaper details. setTimerEnabled() now uses _evse with a null guard, matching the rest of the class. The priority selection in loop() is now (_timer_controlled && !config_current_shaper_enabled()) — a scheduled window can never demote an always-on shaper below Safety(5000); it only claims at Limit(1100) when the shaper runs solely because of the window. This also handles config-disable mid-window correctly (shaper stays alive via _timer_controlled and drops to 1100).

5. Test coverage for the wrap math; sim choreography deferred. Extracted the weekly-wheel math into a dependency-free scheduler_time.h (weeklySpan(), used by both getDuration and getDelay — they were re-implementing the same wrap twice) and added 13 doctest cases covering exactly the reviewer's scenarios: Sunday-only 1-minute window, event-already-passed-today, midnight crossing, Sat→Sun week boundary, lone event, zero-delay-means-now. 

6. Scheduled OCPP placeholder, fixed in the GUI. The RuleModal option is now permanently disabled with a localized "Scheduled OCPP is not supported yet" suffix (en/es/fr/hu), leaving the working "Always Active" OCPP card untouched.

RAR:
Re-reviewed the new commits @chris1howell — I pulled `eea25092`/`566fab0e` and verified against the code. The capacity work came out better than I'd hoped (catching the `GET /schedule` truncation too — sized docs on the read path means the GUI can't silently read back a clipped rule list), the shaper priority condition is right, the `probeReader()` + `schedule_feature_skipped` event breaks the fail-open circularity cleanly, and dropping the dead first-connect block with that explanatory comment is exactly the right call.

Two things from the re-check:

**1. The `scheduler_time` tests didn't make it into the push.** The comment mentions 13 doctest cases in `test/test_scheduler_time`, but they're not in the tree — `weeklySpan` is only referenced by `scheduler_time.h` and `scheduler.cpp`, and there's no test env change in `platformio.ini`. Looks like the test file got left out of the commit. (The refactor itself is good — the `wrapZero` distinction between duration-wraps-to-a-week and delay-zero-means-now is exactly the subtlety worth pinning in a test.)

**2. The gui-nightshift pin isn't pushed, and CI is red because of it.** "Build the static GUI files" fails with:
```
fatal: remote error: upload-pack: not our ref 783d73bb4183ef44a67e8eaf7b69baf637b91c52
```
The submodule pointer references a commit that isn't on the upstream GUI repo or your fork — looks like a local merge commit that needs a `git push` on the gui-nightshift side (the `charge_manager` branch head there is `9a794447`, the OCPP-disable fix, so the pin is one commit ahead of what's published).

chris1howell:
Thank you @RAR - Both issues resolved:

test/test_scheduler_time/test_scheduler_time.cpp committed and pushed doctest cases covering duration/delay weeklySpan semantics. No platformio.ini change needed; the existing native_test env is designed for exactly this pattern (pio test -e native_test).

GUI submodule: The merge commit (783d73b) is now published on OpenEVSE/openevse-gui-nightshift charge_manager, so CI's "Build the static GUI files" step will be able to resolve the pinned ref.

src/app_config.cpp
```diff
@@ -409,6 +414,18 @@ config_load_settings()
...
+    // as the old default (not an intentional change) so it becomes enabled on
+    // upgrade; a current 1 is kept as an intentional setting.
+    new_changed &= ~CONFIG_TEMP_THROTTLE;
+    if(flags != CONFIG_DEFAULT_FLAGS &&
+       CONFIG_TEMP_THROTTLE == (flags & CONFIG_TEMP_THROTTLE))
+    {
+      new_changed |= CONFIG_TEMP_THROTTLE;
+    }
+#endif
+
```
jeremypoulter, line 428: Is this needed?

jeremypoulter (commented):
It would me great to create some simulator test cases around these  , so we can see the interactions working over time. Will be fare easier to do a complete test that way

src/app_config.cpp
chris1howell: This is upgrade-migration logic in config_load_settings() (app_config.cpp:417). It exists because the firmware changed the factory default for temperature throttling from disabled to enabled, and it needs to decide, for boards being upgraded, whether an existing 0 in the flag means "user turned it off on purpose" or just "old default."

Existing config has temp throttle off (the old default) → treated as untouched, so line 436 switches it on (new default wins).
Existing config has temp throttle on → kept, and recorded as an intentional setting.
It's an exact copy of the pattern right above it for CONFIG_DEFAULT_STATE (lines 405–415), which handled the same default-flip situation for the default charge state. The whole thing is compiled in only when the build actually defines the new default as enabled (#if CONFIG_TEMP_THROTTLE_DEFAULT != 0), and it only runs once per device — after it executes, flags_changed is non-zero and committed, so the migration never re-triggers.

chris1howell: I agree with this direction long term. That said, the simulator/test/native work has slowed things down quite a bit, and I'm keen to get a pre-release out soon — 5.1.5 has a number of bugs that are already fixed and waiting to ship. We also promised load sharing would be available for testing many months ago, so it would be great to include it.

I've been running Master + Charge Manager for a few weeks now, and in my experience the code works well enough to start collecting pre-release feedback.

For testing, I took the old-fashioned brute-force approach: I set up a ton of schedules firing every few minutes and watched the behavior in the claims and on the station. Here's an example of scheduled eco with grid shaping hitting a temperature limit. This is the scenario I worried about as Temp Throttoling and Shaping both have a equal 5000 priority. 

Screenshot 2026-07-05 125754

Screenshot 2026-07-05 125828

KipK:
@chris1howell I have tested your branch. Mostly looks ok here. 
I've noticed minor things:

- The default state at boot needs a reboot after change to be taken in count. 
- setting the current in Charge Manager change the Max current setting from the configuration.
I don't think it should be duplicated here, this is the max hardware setting , and this parameter is writen in flash, not supposed to be changed regularly.
It would be more approriated to expose a feature allowing user to program a charge_current claim in scheduled rules.

**edit:** And keep the main slider you've added to set a charge_current claim with auto_release: false 

- the station parameter pop up only have a cancel button, it's a bit confusing as there's no validation button too. Just a Close button perhaps ? 
image

chris1howell:
@KipK - Thank you for the feedback.

I am not sure we really even need Default State Active/Disabled anymore, we have Bootlock which can prevent the station from activating before the schedule takes over. 

The top block in Charge Manager is meant for boot up defaults. The slider is default current not hardware max. Hardware max is a one time write once command. My station has already had the write ($SC 40 M) which is why my slider is 6 (hardware Min) to 40 (Hardware max). I will added the "Station Defaults" section title. 

The Override section has not changed and is where an active charge should be altered. Maybe the name "Charge Manager" is misleading. I am open to a name change maybe "Charge Planner"...? 

The ability to set current in a scheduled rule is already there. I have my station set to charge at 24A at nighttime and ECO during the day.

Screenshot 2026-07-07 194101

Good call on Cancel vs Close. Changed.

Screenshot 2026-07-07 194331

KipK:
@chris1howell , default state is still needed of you don't use  schedule ( I don't anymore ). Here I only enable it manually when needed and let it disabled most of the time as it's accessible outside and don't want people in the street to use my charger.

> The slider is default current not hardware max. Hardware max is a one time write once command

Yeah I know that, I meant the soft hardware max ( almost same purpose btw but on wifi module side ) 
Screenshot_20260708-074931

But this is still a configuration setting not a claim.
Looks a bit awkward on a UX perspective to have it mirrored in the charge manager section. 
Remember this one is the first setting you set in the bootstrap configuration phase of the wifi module. 

Would have more sense to set a charge_current claim that stays after unplugging the car ( auto_release false ).
I'm agree we would need to let the auto_release false  claims to survive reboot then.  

> Maybe the name "Charge Manager" is misleading. I am open to a name change maybe "Charge Planner"...?

Yeah that's it's. I think we have two different use case in the same section here that is not really UX friendly. 
I guess the station defaults perhaps should be moved in configuration section, and just keep a Charge Planner here?
That would make this screen lighter and easier to understand. 

> The ability to set current in a scheduled rule is already there.

Oups I've missed this one :)

chris1howell:
@KipK 
Wouldnt a Rule - Anytime, Mode - Disabled do the same thing?

What really bugs me about Default State (Enable/Disable) is it ONLY does anything if the are 0 timers. Once you add one or more timers toggeling it does nothing. 

I agree the setting default boot current should only be in one place. I will have to think about it and take a look at Settings/EVSE (its a UX nightmate too). 

My thought when building Charge Manager/Planner was to put all settings and features that influance current in one place.  We get a lot of support requests due to unexpected current, mostly Heartbeat Supervision bugs in 5.1.x but sometimes eco or shaper too.

PR #1126, opened 2026-06-17
LVGL UI for the stock ILI9488 TFT display

RAR:
An **LVGL**-based renderer for the stock OpenEVSE color display (ESP32-WROOM / ILI9488 320×480 SPI, env `openevse_wifi_tft_v1`). It **replaces** the immediate-mode TFT_eSPI `screens/` renderer as *the* on-device renderer for the stock TFT — `openevse_wifi_tft_v1.*` defaults it on; there is no opt-in flip. Read-only single screen, dark (nightshift) or light theme.

The legacy TFT_eSPI renderer stays in-tree (`src/lcd_tft.cpp`, `src/screens/`) only because the Elecrow ESP32 HMI dev board (`elecrow_esp32_hmi_dev`) still selects it via `ENABLE_SCREEN_LCD_TFT`.

### Why
The ILI9488 is 18-bit (no TFT_eSPI DMA), and the stock module has no PSRAM, so a full-frame repaint is bus-bound (~9 fps) regardless of renderer. LVGL's dirty-rectangle rendering keeps a ~1 Hz status screen smooth from a single ~30 KB internal-DRAM partial buffer, and gives a cleaner basis for UI work (and shares a design language with the P4 panel UI).

### What's included
- **`lcd_lvgl` is the stock-TFT renderer**, a drop-in `LcdTask` with the same public API, so `main`/`net`/`ocpp` are unchanged. `lcd.h` selects it; the old `lcd.cpp` + `screens/` stand down for this env. TFT_eSPI remains the low-level ILI9488 SPI driver underneath.
- **Boot splash** → main screen after ~4 s.
- **Charge screen**: power ring + centered state word (big kW when charging), pilot/allowed current below the ring, date/time, EVSE temp, Wi-Fi %, vehicle-connected, ELAPSED / DELIVERED / VOLTS-AMPS tiles, hostname + IP, with transient boot/OTA messages.
- **QR WiFi-setup screen** in softAP mode: a `WIFI:…` QR a phone scans to join `OpenEVSE_<id>` and reach the setup page. Auto-switches to the charge screen once configured.
- **Dark + light themes**, with a runtime switch (see below). Both use the exact gui-nightshift `app.css` `[data-theme]` token sets, so the panel matches the web GUI.

### Theme switch
The `NS_*` colour macros resolve a runtime-selectable palette (`NsPalette` struct + `ns_active` pointer), so all draw code is theme-agnostic. Selection is a new `tft_theme` config option (`"dark"` | `"light"`, default `"dark"`), gated to `ENABLE_SCREEN_LVGL_TFT` builds — its presence in `/config` doubles as the GUI's capability signal that the device has the panel. `lcd_lvgl` applies it at boot and polls it each 1 Hz loop; on change it swaps the palette and rebuilds the current screen in place (no reboot). Driven entirely through the `/config` GET/POST the web GUI already uses — no new `LcdTask` API.

### Build / scope
- `platformio.ini`: the TFT panel hardware flags are split from the renderer enable. The LVGL renderer enables live in a shared `common.lvgl_tft_renderer_flags` fragment that `openevse_wifi_tft_v1`(+`_dev`) pull in — so the stock TFT env renders via LVGL by default, no per-env flag duplication.
- `openevse_wifi_tft_v1` builds at RAM 32.4% / Flash 33.8%; other envs are unaffected (the gate is off everywhere else).
- HW-validated on a bench ILI9488 unit: boot splash, idle/charging/sleeping/disabled states, Wi-Fi %, pilot, hostname/IP, the setup QR (scans + joins), and a live dark↔light theme switch via `/config`.


jeremypoulter:
Great thanks for this. IMHO, just replace the current UI, I don't see a need to maintain that. @chris1howell what do you think.

chris1howell (dismissed):
Looks great.

chris1howell (approved):

PR #1136, opened 2026-06-24
Add in-place 16MB flash repartition migration tool

chris1howell:
# In-System 4 MB → 16 MB Repartition — Successful Implementation

Thanks to RAR for the missing piece... https://github.com/RAR/openevse-16mb-migrator

**Status: working, hardware-validated.** A 16 MB ESP32 WiFi module that was
flashed in the field with the **4 MB** partition layout (`min_spiffs.csv`) can now
expand itself to the full **16 MB** layout (`openevse_16mb.csv`) **entirely over
the air**, from a button in *Settings → Developer Tools → Flash & storage*. No USB,
no opening the enclosure. WiFi credentials and all settings survive.

Test on a 16MB ESP32 flashed with a 4MB OpenEVSE ESP32 WiFi Build. Many products were silently upgraded to 16MB. 

Flash openevse_wifi_v1.bin from here: https://github.com/OpenEVSE/openevse_esp32_firmware/releases/tag/vRePartition4MB
Navigate to Settings => Devoloper 
If Module is 16MB Flashed as 4MB the Repartition button will Appear.
Click Repartition.
Files will be staged from Github, module rebooted and partitions modified.
Module will reboot with 16MB firmware.

- end-to-end run: ~75 s, fully OTA, device came back as a healthy
  `openevse_wifi_v1_16mb` (`partition_scheme:"16mb"`, `espflash:16777216`,
  `app0_size:6553600`, `littlefs_size:3538944`, `can_expand_16mb:false`), WiFi
  reconnected on its own.
- **Zero bricks** across every development attempt (~18). That is not luck — it is
  the direct result of the staging-before-commit design described below.

---

## Why this was hard

The whole job is one protected write the running firmware **cannot** do:

- The Arduino `Update` class only writes the *inactive app partition*. It cannot
  write the **bootloader** (`0x1000`), the **partition table** (`0x8000`), or
  **otadata** (`0xe000`) — and those three are exactly what must change to grow
  the layout.
- Dropping to the raw `esp_flash_*` API doesn't help either: the prebuilt Arduino
  `spi_flash` library is compiled with **`CONFIG_SPI_FLASH_DANGEROUS_WRITE_ABORTS`**.
  Any write that lands in the bootloader / partition-table / otadata region is
  refused at runtime and **aborts the CPU**. You cannot recompile that flag out of
  a prebuilt library.

So the 4 MB firmware is structurally incapable of committing the new layout. The
solution is to **delegate that one commit** to a purpose-built helper that *was*
compiled to allow it.

---

## The architecture: stage in free space, then hand off to a migrator

There are two cooperating programs:

1. **The OpenEVSE 4 MB firmware** (this repo, `src/flash_migrate.cpp`) — does all
   the network work and all the *non-protected* writes while charging/WiFi keep
   running. It never touches a protected region.
2. **The migrator** (`migrator/` submodule →
   [OpenEVSE/openevse-16mb-migrator](https://github.com/OpenEVSE/openevse-16mb-migrator))
   — a ~130-line ESP-IDF app built with
   **`CONFIG_SPI_FLASH_DANGEROUS_WRITE_ALLOWED=y`**. It does *only* the protected
   commit, then reboots into the new firmware. It carries the 16 MB bootloader and
   partition table embedded inside itself.

### The two layouts (verified offsets)

```
                   min_spiffs (current 4 MB)     openevse_16mb (target 16 MB)
bootloader  0x1000   (core-2)                    (core-3)            <- changes
partitions  0x8000                                                   <- changes
otadata     0xe000 / 0x2000                       0xe000 / 0x2000    <- re-pointed
nvs   data  0x9000 / 0x5000                        0x9000 / 0x5000   == SAME (creds survive)
app0  ota_0 0x10000 / 0x1E0000                     0x10000 / 0x640000
app1  ota_1 0x1F0000 / 0x1E0000                    0x650000 / 0x640000
spiffs data 0x3D0000                               0xc90000 / 0x360000
```

The crucial facts that make safe staging possible:

- **`nvs` is at `0x9000` in *both* layouts.** It is never erased or moved, so WiFi
  credentials, MQTT config, and all NVS-stored settings carry over untouched.
  (EEPROM-stored config is likewise outside the partition table.)
- The 16 MB layout's `ota_1` (`0x650000`) and `spiffs` live in the **4–16 MB
  region that the current 4 MB layout doesn't use at all.** Writing there cannot
  corrupt the running system — it's empty space as far as the live firmware is
  concerned.
- Raw `esp_flash_write` to `0x650000` is a *normal*, allowed write (it's nowhere
  near a protected region), so the 4 MB firmware can stream the full 16 MB app
  there while WiFi is up.

---

## The migration sequence

Driven by `flash_migrate_loop()` (a main-loop pump, so downloads never re-enter
the mongoose event loop) — `src/flash_migrate.cpp`:

1. **Manifest** — download `migrate_v1_16mb.json` from the GitHub release. It is
   the `{app, migrator}` schema: each entry has a URL + SHA-256.
2. **Stage the 16 MB app** → stream it to `ota_1` at **`0x650000`** via raw
   `esp_flash`, one 4 KB sector at a time. On completion: verify the streamed
   **SHA-256** *and* the app image magic byte (`0xE9`). Mismatch → abort, nothing
   else has changed.
3. **Stage the migrator** → OTA it into the **inactive 4 MB OTA slot** with the
   normal `esp_ota_*` API (`esp_ota_get_next_update_partition` →
   `esp_ota_begin/write/end`). `esp_ota_end` validates the image; then we verify
   our own **SHA-256** before letting it boot.
4. **Hand off** → `esp_ota_set_boot_partition(migrator)` + reboot.
5. **The migrator runs** (under the *old* 4 MB bootloader) and performs the
   protected commit — see next section.
6. **Reboot into OpenEVSE 16 MB.** First boot formats the blank `spiffs`; NVS +
   EEPROM config intact.

A **dry-run** mode does steps 1–3 (download + full verify of both images) and then
stops, writing nothing and booting nothing — used to exercise the whole risky
download/verify path with zero risk.

---

## Why flashing is safe

The design guarantee is: **at every instant the device can boot, it boots either
the old OpenEVSE-4 MB or the new OpenEVSE-16 MB — never "neither".** That comes
from a strict *verify-everything-before-you-commit* ordering and from keeping all
staging in free space.

| Safety property | How it's achieved |
|---|---|
| **Staging can't corrupt the running system** | The 16 MB app is written to `0x650000` and the migrator to the *inactive* OTA slot — both unused by the live 4 MB firmware. The current bootloader, partition table, otadata, app, and `nvs` are all untouched during staging. |
| **Nothing boots until it's fully verified** | The app is SHA-256 + magic-byte checked; the migrator is `esp_ota_end`-validated **and** SHA-256 checked. Only then is the migrator made bootable. A flipped byte → SHA-256 fail → clean abort, original system still intact. |
| **Power loss during staging is harmless** | If power drops anytime before the migrator is selected as boot partition, the device simply reboots into the unchanged OpenEVSE-4 MB. Re-run the migration. |
| **The migrator double-checks before committing** | It *refuses to write anything* unless it finds the staged app's `0xE9` magic at `0x650000`. No staged app → it aborts having changed nothing. |
| **Every protected write is read-back verified** | `write_verify()` in the migrator erases, writes, then reads the region back and `memcmp`s it. A bad write is caught immediately. |
| **Settings/credentials never move** | `nvs` (`0x9000`) and EEPROM config are outside everything that changes. WiFi reconnects on its own. |
| **Commit order is failure-tolerant** | bootloader → partition table → otadata. The new bootloader is backward-compatible with the partition layout it's about to install, and otadata (the pointer that actually selects the new app) is written **last**. |

The only data that is **intentionally discarded** is the on-device LittleFS
content (energy history, logs, schedules, certificates) — it lives in the `spiffs`
partition that physically moves. This is a documented, accepted loss; the new
firmware formats a blank filesystem on first boot. Settings are *not* in LittleFS,
so they survive.

---

## Where the soft-brick risk actually is

To be precise rather than reassuring: there **is** a genuine point of no return.
It is small, and it is confined entirely to the **migrator's commit**, after the
hand-off reboot. It is *not* in the OpenEVSE firmware and *not* during any of the
download/staging on the live device.

```
  [download + stage + verify app]      <- safe: free space only, fully reversible
  [stage + verify migrator]            <- safe: inactive OTA slot, fully reversible
  [set boot = migrator] + reboot       <- safe: old 4MB still fully intact & bootable
  ----------------------------------------------------------------------------
  migrator: verify staged app magic    <- safe: nothing written yet
  migrator: WRITE bootloader  @0x1000  <-- ┐
  migrator: WRITE part. table @0x8000  <--  │  POINT OF NO RETURN
  migrator: WRITE otadata     @0xe000  <-- ┘  (sub-second, three erase+write pairs)
  ----------------------------------------------------------------------------
  reboot -> OpenEVSE 16MB              <- safe again: new system live
```

**The risk window** is the moment the migrator begins overwriting the bootloader
at `0x1000` until otadata is written. If power is lost **in the middle of the
bootloader write specifically**, the device has a partially-written bootloader and
will not boot → it must be recovered over **USB** (esptool). This window is
inherent to *any* ESP32 repartition — you cannot atomically swap a bootloader —
and is on the order of a few hundred milliseconds.

Mitigations that shrink (but cannot fully eliminate) this window:

- The migrator writes the bootloader **first** and read-back-verifies it before
  moving on. If the bootloader write *completed* but a later step failed, simply
  **re-running the migrator is safe and idempotent** — it re-writes all three
  regions from its embedded copies.
- Because the migrator is itself an OTA app that persists in its slot, a failed
  commit generally **reboots back into the migrator**, which retries the commit
  rather than leaving a half-migrated device.
- The truly unrecoverable case — power cut *during the few-hundred-ms bootloader
  write* — is the only scenario that needs USB recovery. The guidance to the user
  during the commit phase is therefore the standard **"do not power off the
  charger"** warning, shown in a non-dismissible progress modal.

### Summary of risk by phase

| Phase | Runs on | Brick risk if power lost |
|---|---|---|
| Manifest / app / migrator download + verify | OpenEVSE 4 MB (live) | **None** — reboots into unchanged 4 MB |
| `set_boot_partition(migrator)` + reboot | OpenEVSE 4 MB | **None** — old app still selected until otadata changes |
| Migrator verifies staged app | Migrator | **None** — nothing written yet |
| Migrator writes bootloader `0x1000` | Migrator | **Soft-brick** (USB recovery) — the only real window |
| Migrator writes PT `0x8000` / otadata `0xe000` | Migrator | Recoverable by re-running the migrator |
| Reboot into 16 MB | New system | **None** |

---

## How the migrator gets built and shipped (CI)

The migrator must embed the **same** 16 MB bootloader + partition table that the
16 MB app expects, so CI regenerates them from the very build it publishes
(`.github/workflows/build.yaml`, release job):

1. Build all envs, including `openevse_wifi_v1_16mb`; extract its
   `bootloader_v1_16mb.bin` and `partitions_v1_16mb.bin`.
2. `migrator/scripts/gen_embeds.py` turns those two `.bin`s into
   `src/openevse_bl.h` / `src/openevse_pt.h`.
3. `pio run -e esp32dev` in `migrator/` (ESP-IDF, `DANGEROUS_WRITE_ALLOWED`)
   → `openevse_migrator.bin`.
4. Publish `openevse_migrator.bin` + `openevse_wifi_v1_16mb.bin` and a
   `migrate_v1_16mb.json` manifest carrying both URLs and SHA-256s.

The 4 MB firmware's default `MIGRATE_MANIFEST_URL` points at
`releases/latest/download/migrate_v1_16mb.json`, so a device pulls the migrator
and app that match the firmware it's about to install.

---

## What survives, what's lost

| Data | Location | After migration |
|---|---|---|
| WiFi credentials | `nvs` @ `0x9000` | **Kept** (partition unchanged) |
| All settings (MQTT, options, etc.) | EEPROM / NVS | **Kept** |
| Energy history, logs, schedules, certificates | LittleFS (`spiffs`) | **Lost** (partition moves; reformatted blank) |
| Firmware | `ota_0`/`ota_1` | Replaced with 16 MB OpenEVSE |

---

## Files

- `src/flash_migrate.cpp` / `.h` — detection (`flash_migrate_partition_scheme`,
  `flash_migrate_can_expand_16mb`) and the download/stage/verify/hand-off engine.
- `migrator/` (submodule) — the ESP-IDF commit app; `src/main.c` is the protected
  commit, `scripts/gen_embeds.py` embeds the bl/PT.
- `.github/workflows/build.yaml` — builds + publishes the migrator and manifest.
- Detection surfaces in `/config` as `partition_scheme` + `can_expand_16mb`; the
  GUI button lives in *Developer Tools → Flash & storage*
  (`gui-nightshift`, `RePartition` branch).


chris1howell:
@RAR Can you take a look at this? Downloand and Staging of the files works great, but was unable to complete the final step to write the staged bootloader after a reboot.

At a point a couple years ago when we started using 16MB modules for the TFT LCD, we just started putting 16s in everything, The Juicebox v2 board you have very likely has a 16MB chip flashed with 4MB partition. It may not be possible but it would be great if we could repartition without a serial programmer.

RAR:
Really thorough writeup — and the "stage everything, apply before WiFi, never actually enter the brick window" design is the right shape regardless of how the final write happens. Wanted to flag one thing, because I think the "not feasible on this platform" conclusion may be specific to the prebuilt Arduino framework rather than the hardware.

The dangerous-region guard you hit (`CONFIG_SPI_FLASH_DANGEROUS_WRITE_ABORTS`) is a compile-time Kconfig, not a runtime gate — the other value is `CONFIG_SPI_FLASH_DANGEROUS_WRITE_ALLOWED`. With that set, plain `esp_flash_erase_region`/`esp_flash_write` write `0x1000`/`0x8000` without complaint — the *same* driver you're already using from the running app for the 2 MB app and the scratch staging, and (as you noted) the one that actually gets the cache / cross-core / watchdog coordination right. The ROM `esp_rom_spiflash_*` path can't win for exactly the reason you diagnosed — it can't drive the controller while `esp_flash` owns it — which to me reads as evidence the ROM primitive is wrong, not that the write is impossible.

The catch, and I think the real source of the wall: the prebuilt Arduino-ESP32 package bakes `...ABORTS` into the precompiled `spi_flash` lib, so you can't flip the Kconfig from `sdkconfig` alone. That's why this works cleanly under ESP-IDF but not the stock Arduino build. ESPHome ships this exact feature — OTA-writing the bootloader / partition table / any partition — in esphome/esphome#5535 (`allow_partition_access: true`); under the hood it's `CONFIG_SPI_FLASH_DANGEROUS_WRITE_ALLOWED` + plain `esp_flash`, with a backup-partition-table-and-rollback safety net. It needs IDF 5.1+, which the core-3 boards in this PR already run.

So the whole download/verify/stage pipeline here looks reusable as-is — the only thing that changes is the commit primitive (ROM calls → `esp_flash` with the check allowed). Two ways to get there given the Arduino constraint:

1. **Build Arduino as an IDF component** with `CONFIG_SPI_FLASH_DANGEROUS_WRITE_ALLOWED=y` — the core-3 boards are already on the pioarduino / IDF-5 stack, so the commit step collapses to a handful of `esp_flash` calls. Cleanest, but it means owning a from-source framework build.
2. **A minimal intermediate IDF migrator** — a tiny app that fits the current `ota_1`, no WiFi / no SMP, whose only job is to rewrite `0x1000`/`0x8000` with dangerous-write allowed and reboot into the staged 16 MB app. Keeps the main firmware's framework untouched and confines the dangerous write to a minimal, audited context. (This is also what the arduino-esp32 maintainers suggest when this comes up.)

FWIW I've been running the full `openevse_16mb.csv` layout on a 16 MB WROOM-32E via the one-time USB flash you describe and it's rock solid — happy to help prototype the `esp_flash` commit path if it's worth reopening.

RAR:
Took the `esp_flash` + `DANGEROUS_WRITE_ALLOWED` idea from my comment above all the way to a working migration — on hardware, the full 4 MB → 16 MB repartition now runs **in-system** and leaves a booting OpenEVSE-16 MB install.

I built a small ESP-IDF "migrator" app (esp32dev, IDF 5.5, `CONFIG_SPI_FLASH_DANGEROUS_WRITE_ALLOWED=y`) and ran it on a 16 MB ESP32-D0WD-V3. The real 16 MB OpenEVSE app is staged at ota_1 (`0x650000`) — the free-upper-flash spot your writeup already uses — and the migrator embeds the 16 MB bootloader + partition table. It then:

1. refuses to touch anything unless a valid app image (`0xE9`) is staged at `0x650000`
2. writes + read-back-verifies the OpenEVSE bootloader → `0x1000`
3. writes + read-back-verifies the `openevse_16mb` table → `0x8000`
4. points otadata → ota_1 (`esp_rom_crc32_le` for the select CRC)
5. reboots

**Migrator log:**
```
staged 16MB app present @0x650000 (image magic 0xE9). committing ...
    bootloader @0x1000 OK (25024 B written + verified)
    partition table @0x8000 OK (3072 B written + verified)
    otadata OK (seq=2, crc=0x55f63774)
COMMIT COMPLETE. rebooting into OpenEVSE 16MB (ota_1).
```
…then the real OpenEVSE 16 MB app boots from `0x650000` (LittleFS init + a `$GV` RAPI poll = its main loop running).

**Deterministic read-back of the migrated chip afterward:**
```
bootloader @0x1000  ==  OpenEVSE 16MB bootloader.bin : True
otadata    @0xe000  ota_seq = 2  ->  boots ota_1
partition table @0x8000  ==  openevse_16mb  (app1/ota_1 @0x650000 6400K, nvs @0x9000, ...)
```

Two things from the layouts make this clean and safe:
- `nvs` (`0x9000`) and `otadata` (`0xe000`) are byte-identical offsets in `min_spiffs` and `openevse_16mb`, and the migrator never touches `nvs` — so **WiFi creds survive the migration for free**.
- The only dangerous write is the ~25 KB bootloader; it's written + verified *before* the single partition-table sector flip, so the point of no return is one sector, and otadata can roll back. The device is always old-OpenEVSE-or-new, never neither.

No ROM `esp_rom_spiflash_*`, no `Cache_Read_Disable` — `esp_flash` does the cache / cross-core / watchdog coordination itself once the Kconfig gate is open, which is the whole difference from the path that hit the wall.

**How it maps onto this PR:** your download → verify → stage pipeline is reused unchanged. The only change is the commit step. Instead of the in-Arduino ROM write (which can't work — the prebuilt Arduino `spi_flash` lib is compiled with `DANGEROUS_WRITE_ABORTS`), OpenEVSE-4 MB OTAs into this tiny IDF migrator (it's ~210 KB, fits a 4 MB OTA slot) and reboots into it; the migrator does the protected commit and boots the 16 MB app. The front-half (download + stage the three images + reboot into the migrator) is exactly your existing engine plus the handoff.

What I haven't done: that front-half wiring, and the deepest "click around the web UI" sign-off (my bench board has no controller and the host can't WiFi-scan — but the chip is left in the exact `openevse_16mb` state and OpenEVSE boots and runs). Happy to share the migrator (~130 lines + the embedded bootloader/PT) and help wire the handoff if it's useful.

RAR (approved):

KipK:
hi @chris1howell @RAR 
I've just flashed a daily dev build ( did it last week last time ).
Not sure it is this PR but there's chance it is. 
Now updating to latest dev builds let the openevse_wifi device unreachable after flashing.
If we hard reset by cutting off the power, then the wifi mdule is back online, but it then display there's no evse module.  And after few minutes it gets unresponsive again. 

My module was already partitionned for 16mb as I was previously using the 16mb build successfully.

edit: also can't roll back by flashing from the UI anymore. I'll have to unmount the charger and access the module to flash it with wires. I guess it screwed up the partioning on the device

How as it been tested on your side ? Seeing this dangerous change beeing merged this fast without proper testing is a bit dangerous I guess

edit2: Perhaps the problem came from the fact I probably had a custom partitioning as I passed it to 16mb more than a year ago when you sent me the prototypes @chris1howell , but can't confirm.

chris1howell:
@KipK The partition tools are not included on 16MB builds. They are only on the 4MB openevse_wifi_v1 build.

On the 4MB build it checks for a 16MB module and presents a button. No flashing or partitioning is activates until it is chooses ant there are guards and checks at every step. 

I have not seen any bricked modules with the partition code. I have seen it on energy logging which I believe is due to many SPIFFS writes that do not validate enough space prior to writing. I believe the code for all writes to validate is in the Charge Manager PR.

KipK:
@chris1howell ok so not coming from there. 
 Not sure what could happend, but even with flashing it over serial, repartionning, reflashed bootloader etc, it stills can't detect the evse module anymore and seems to have random lockout issues. 
Seems dead somehow.  I hope it's not coming from the energy logging, or we will have a bunch of complains if it fails like that after few weeks. 

Fortunately i still have an old 4mb module. I have tried it and seems to works ok. At least it's not the EVSE module that has failed.

chris1howell:
@KipK I just fixed the can not detect EVSE module banner (False Positive) in the charge manager branch module was communicating just fine.

 bool isConnected() {
   -   return OpenEVSE.isConnected();
   +  return _openevse.isConnected();
    }

chris1howell:
@KipK I believe this fix in the Charge_Manager branch that solves the issue of SPIFFS corruption, it is not the energy logging writes, that is already gaurded checking for space befor writing. It is other writes that write unconditionally.

https://github.com/OpenEVSE/openevse_esp32_firmware/pull/1112/changes/d62faa38e672fb204118e4a2ab183661aee44ad0

PR #1139, opened 2026-06-29
Add native LVGL local UI build configuration, display modes, and screenshot export

jeremypoulter:
This adds a native host build configuration for the LVGL local UI so the local-display path can be compiled outside the ESP32 target. The change also adapts the LVGL panel/tick path for native execution while keeping the existing ESP32 TFT flow intact, adds a host-side export path for generating review screenshots from the LVGL screens, and adds a runtime-selectable native display mode for interactive local development.

- **New native LVGL environment**
  - Adds `native_openevse_lvgl` as a host-side PlatformIO env layered on `native_openevse`
  - Enables the LVGL renderer flags and pulls in the LVGL dependency for native builds
  - Defines the TFT dimensions expected by the local UI code

- **Native LVGL panel path for host builds**
  - Splits `lvgl_panel.cpp` between ESP32 hardware init and native host init
  - Uses host heap allocation under `EPOXY_DUINO` and keeps a native framebuffer for screen export
  - Keeps the existing TFT_eSPI / internal-heap path unchanged for ESP32

- **Native-compatible LVGL runtime behavior**
  - Reworks LVGL tick handling so native builds advance LVGL explicitly from the task loop instead of relying on `Arduino.h` inside LVGL C sources
  - Pumps the optional native window renderer from the same LVGL loop/capture path
  - Adjusts setup-screen SSID generation to use config/native-safe inputs rather than an ESP32-only WiFi API

- **Native LVGL display modes**
  - Adds `--lvgl-display headless|window` to the native binary
  - Keeps headless as the default mode for CI and screenshot export
  - Adds an SDL-backed window mode for interactive local LVGL UI development when SDL2 is available on the host

- **Native LVGL screenshot export**
  - Adds a host-only `--dump-lvgl-screens DIR` path to the native binary
  - Renders and exports sample boot, setup, and charge-state LVGL screens as image files for review
  - Includes an updated composite screenshot sheet in `docs/lvgl-local-ui-screens.png`

- **Docs**
  - Documents the new native LVGL env in the build guide and clarifies that it defaults to headless mode on the host
  - Documents how to launch the interactive window mode from the native build
  - Documents how to export the LVGL review screenshots from the native build

```ini
[env:native_openevse_lvgl]
extends = env:native_openevse
lib_deps =
  ${env:native_openevse.lib_deps}
  lvgl/lvgl@^8.3.9
build_flags =
  ${env:native_openevse.build_flags}
  ${common.lvgl_tft_renderer_flags}
  -D LCD_BACKLIGHT_PIN=0
  -D TFT_WIDTH=320
  -D TFT_HEIGHT=480
  -ldl
```

LVGL local UI screens


jeremypoulter:
@RAR / @chris1howell please take a look

RAR:
sweet.  lemme take a look.

RAR:
Looks great - I need to fix the Connected and Not Connected being too wide for the space.  Already helpful!  🤣

RAR (approved):

PR #1140, opened 2026-06-29
fix(lcd): keep the LVGL state word inside the ring (no wrap)

RAR:
On the LVGL charge screen, the centre status word overflowed the ring at `montserrat_28`: `NOT CONNECTED` wrapped to three lines (`NOT` / `CONNECTE` / `D`) and even `CONNECTED` broke mid-word (`CONNECTE` / `D`). The centre label is 180&nbsp;px wide, but the ring's usable inner area (200&nbsp;px ring − 14&nbsp;px band each side ≈ 172&nbsp;px) is narrower, so the long words spilled and wrapped over the ring band.

Before/after — LVGL state-word ring fix

### Fix

Rather than a fragile per-state char-count threshold, measure the word's rendered width at the large font with `lv_txt_get_size()` and step down to `montserrat_20` only when it would exceed the ring's inner width (`STATE_WORD_FIT_W`). Short states (`CHARGING`, `FAULT`, `SLEEPING`, …) keep the large headline; the wide ones drop one size and stay on a clean line. This is self-correcting for any future state string and adds no new font (20 is already compiled in).

One file, +14 lines, ESP32 render path otherwise unchanged.

### Verification

Built and rendered on the host through the native LVGL screenshot export from #1139 (`--dump-lvgl-screens`):

- **Before:** `NOT CONNECTED` → 3 lines; `CONNECTED` broke mid-word.
- **After:** `CONNECTED` is a single line; `NOT CONNECTED` falls to two balanced lines (`NOT` / `CONNECTED`) with no mid-word break — both stay inside the ring.

(Nice side-note: #1139's export was exactly what surfaced this, and what verified the fix.)


RAR:
@jeremypoulter yup, I like it.

chris1howell (approved):

PR #1141, opened 2026-06-30
RFC Secure Protocol HTTPS and/or SSH

chris1howell:
Wrote SSH as a joke, took about 30 minutes with claude, so I figured I would try HTTPS  (much more time and way more difficult). On HTTPS I can only get a single TLS connection to work, JS and CSS files had to be combined and WS connections paused when TLS is neeeded. HTTPS works but it could use improvement.

It would be nice to have a secure encrypted management choice. SSH works suprisigly well even with the minimal effort.

# HTTPS and SSH CLI 

## HTTPS

Screenshot 2026-06-29 194918

### Overview

The ESP32 gateway runs a Mongoose IoT HTTP server (`MongooseHttpServer`) with optional TLS via **mbedTLS**. HTTP and HTTPS are independently switchable and can run concurrently on separate ports.

### TLS Stack

| Layer | Implementation |
|---|---|
| TLS library | mbedTLS (bundled with Arduino ESP32 / ESP-IDF) |
| RNG | `mg_ssl_if_mbed_random()` → `esp_fill_random()` (hardware HWRNG) |
| ESP32 / ESP32-C3 entropy | Wi-Fi RF noise (always active while the gateway runs) |

The production firmware replaces Mongoose's default `rand()`-based RNG stub (`MG_SSL_MBED_DUMMY_RANDOM`) with a real hardware-backed implementation in `src/mongoose_rng.cpp`.

### Certificate Store

Certificates are managed by `CertificateStore` (`src/certificates.cpp`) and persisted as JSON files in LittleFS under `/certificates/<serial_hex>.json`.

Screenshot 2026-06-29 202819

Two certificate types are supported:

| Type | Fields | Use |
|---|---|---|
| `root` | `certificate` (PEM) | Adds a trusted CA to outbound TLS validation (MQTT, OTA, EmonCMS, etc.) |
| `client` | `certificate` + `key` (PEM) | Used as the HTTPS server certificate, or for MQTT mTLS client auth |

On load, each certificate is validated with `CertificateValidator` (mbedTLS on device, OpenSSL in native/test builds). The root CA bundle starts from the compiled-in `root_ca.h` bundle; any user-added root CAs are appended to it at runtime.

Screenshot 2026-06-29 202927

**X.509 serial numbers are truncated to the least-significant 8 bytes (64-bit).** Certificates with serials longer than 64 bits are accepted but indexed by the truncated value.

### Configuration Keys

| Key | Default | Description |
|---|---|---|
| `www_https_enabled` | `false` | Enable the HTTPS listener (flag bit `CONFIG_HTTPS_ENABLED`) |
| `www_http_enabled` | `true` | Enable the HTTP listener |
| `www_certificate_id` | `""` | Hex serial of the client cert to use as the server certificate |
| `www_http_port` | `80` | HTTP listen port |
| `www_https_port` | `443` | HTTPS listen port |
| `mqtt_certificate_id` | `""` | Hex serial of the client cert for MQTT mTLS |

`CONFIG_HTTPS_ENABLED` is intentionally excluded from the factory-default flags. HTTPS stays disabled until the user explicitly selects a certificate, preventing a misconfigured HTTPS listener with no working HTTP fallback from locking the device's web UI out entirely.

### Server Startup (`web_server_setup`)

```
if HTTPS enabled AND www_certificate_id set AND cert+key loads OK:
    server.begin(www_https_port, cert, key)   ← TLS listener
    registerWebServerRoutes()
    use_ssl = true

if HTTP enabled OR HTTPS failed to start:
    server.begin(www_http_port)               ← plaintext listener
    registerWebServerRoutes()
```

Routes must be registered once per listener because Mongoose attaches endpoints to the active `mg_connection*` at the time of registration, not to the server object globally. If HTTPS fails to start for any reason (disabled, no cert chosen, cert load failed), HTTP is force-enabled as a fallback so the device is never unreachable.

### HTTP → HTTPS Redirect

Both HTTP and HTTPS can currently be enabled at the same time.

### HTTPS Limitations

Only a single TLS session, GUI JS and CSS combined. WS connections must be paused to maintain a single session.

---

### Overview

Screenshot 2026-06-29 201549

The SSH CLI provides a Cisco IOS-style interactive management interface over SSH, running alongside the existing HTTP web server. It is enabled by the build flag `ENABLE_SSH_CLI` and active in the  `openevse_wifi_v1_16mb` firmware builds.

### Architecture

The fundamental challenge is that libssh's API is **synchronous and blocking** (`ssh_bind_accept`, `ssh_message_get`, `ssh_channel_read`), while the firmware's main scheduler (MicroTasks) is cooperative and must return promptly. The solution splits responsibilities across two contexts:

```
Main thread (MicroTasks loop)          SSH FreeRTOS task (core 1)
──────────────────────────────         ──────────────────────────
SshServerTask::loop()                  SshServerTask::runServer()
  dequeues CliCmdRequest                 ssh_bind_accept()  ← blocks until client connects
  calls handler(out, argc, argv)         handleSession()
  xSemaphoreGive(req.done)               authenticate()
                                         ssh_channel_read()  ← blocks on client input
                                         CliEngine::feedByte() per byte
                                           ↳ tab/help: runs inline (read-only tree)
                                           ↳ command: xQueueSend(req) + xSemaphoreTake(done)  ← blocks
```

Command **execution** (anything that calls `config_set()`, `EvseManager`, `limit`, etc.) always runs on the main MicroTasks thread via queue + semaphore handoff. Tree-walking, tab-completion, and `?` help run directly on the SSH task since they only read the `const CliNode[]` tree and per-session `CliEngine` state.

### Host Key

- Algorithm: **Ed25519** (generated by libssh `ssh_pki_generate`)
- Storage: LittleFS at `/ssh_host_key.pem` (PEM-armored private key, ~400+ bytes)
- Key generation and loading runs on the **main thread** in `setup()` because it calls `config_set()`/`config_commit()` — the SSH FreeRTOS task only reads the pointer afterwards
- The key persists across reboots. If the stored key fails to import it is silently regenerated

The host key is stored separately from the `user_config` EEPROM-emulated region (which is bounded at 3072 bytes and shared by ~90 other config fields) to prevent silent truncation of the blob.

### Authentication

- Method: **password only** (`SSH_AUTH_METHOD_PASSWORD`)
- Credentials: `ssh_username` and `ssh_password` config keys (set via `set ssh username` / `set ssh password` in CONFIG mode, or the web UI)
- No public-key authentication; no keyboard-interactive

### Session Handling

- **Single-session by construction**: `runServer()` calls `ssh_bind_accept()` again only after the prior `handleSession()` returns. There is never more than one session in flight.
- **Heap gate**: New connections are refused if free heap falls below `SSH_MIN_FREE_HEAP_TO_ACCEPT` (40 KB). libssh session setup costs approximately 20–40 KB.
- **PTY requests**: Accepted with success reply so the client's terminal operates in raw mode — Tab and arrow keys are forwarded as bytes instead of being processed locally by the SSH client.

### CLI Engine

The `CliEngine` (`src/cli/cli_engine.cpp`) is instantiated per-session and holds all per-connection state:

| Feature | Detail |
|---|---|
| Line buffer | `CLI_LINE_MAX` = 128 bytes |
| Token limit | `CLI_MAX_TOKENS` = 12 tokens per line |
| History | `CLI_HISTORY_SIZE` entries (ring buffer), ↑/↓ arrow navigation |
| Tab completion | Unambiguous prefix expansion, hidden nodes excluded (e.g. `do`) |
| `?` help | Lists matching children and argument type hints |
| Abbreviation | Unambiguous prefix matching (e.g. `sh ver` = `show version`) |
| Ctrl-C | Clears the current line and re-prints the prompt |

### CLI Mode Hierarchy

```
host>          Unprivileged EXEC — show version/status/faults, enable
  ↓ enable
host#          Privileged EXEC — all show commands, charge, clear, rfid, reset, reload, install, debug, write, configure terminal
  ↓ configure terminal
host(config)#  Configuration — set, feature, no, service, do
  ↓ exit
host#
  ↓ exit
host>
```

Mode transitions are intercepted by the engine by function-pointer identity before reaching the `CliExecutor` — they are per-session local state and never marshalled to the main thread.

### Build Configuration

```ini
-D ENABLE_SSH_CLI                    ; enable the SSH server
```

SSH credentials are stored as plain config strings (`ssh_username`, `ssh_password`) and survive reboot. The host key lives at `/ssh_host_key.pem` in LittleFS and is separate from the config blob.


jeremypoulter:
Weird HTTPS was already implemented so should have been easy ;-) I just have never made it avalible as it didn't work correctly, If/when I got around to it again we need to complete the migration to [Mongoose 7](https://github.com/jeremypoulter/ArduinoMongoose/pull/38) as I think that will solve many of the issues

If we really want SSH, sure why not, lol, but a CLI tool is probablu just going to be more useful. I would split it out into a separate PR though.

chris1howell:
@jeremypoulter Implementing HTTPS was simple, you did a great job on the certificute mamager. The "didn't work correctly" is the hard part. At first it was a great secure white screen. Once I was able to get the page to load, most of the content was broken. It is kinda working but still not perfect.

Luckily, I was able to re-load firmware via SSH.

RAR:
Yeah, I've never found https to be reasonable on such a ram limited device.  Feel like Mongoose 7 is really the only path to actually trying it if its truly wanted.  You'll still have users needing to click through a self signed cert though.

jeremypoulter:
> Yeah, I've never found https to be reasonable on such a ram limited device.  Feel like Mongoose 7 is really the only path to actually trying it if its truly wanted.  You'll still have users needing to click through a self signed cert though.

Yeah that was my thoughts. Will try and put some effort into completing the v7 port, I don't think there was that much left to do....

jeremypoulter:
> @jeremypoulter Implementing HTTPS was simple, you did a great job on the certificute mamager. The "didn't work correctly" is the hard part. At first it was a great secure white screen. Once I was able to get the page to load, most of the content was broken. It is kinda working but still not perfect.
> 
> Luckily, I was able to re-load firmware via SSH.

Yeah what sounds about correct. You should add port forwarding to the SSH server so we can do HTTP over an SSH tunnel

chris1howell:
Okay, I am going to split HTTPS and SSH. Lets put HTTPS back on the back burner for now. I would like to shift focus to getting a pre-release including load sharing so we can get some testers and feedback.

jeremypoulter:
Asked Copilot to pick up the Mongoose 7 uplift.

Issue #940, opened 2024-12-13
Local power sharing group


chris1howell:
This is an expansion #592.

With Enel X shutting down servers in North America, many Juicebox owners are looking for solutions to continue operating their stations. OpenEVSE has developed drop in replacement controllers.

One popular feature on the JuiceBox was cloud server based load sharing. Load sharing on OpenEVSE is possible with external hardware and software such as Home Assistant or MQTT and logic from Node-Red or a Python script.

This issue proposes a local load sharing setup that can be easily implemented without and  dependencies on additional hardware and software.

On the primary station:
Add new share group.
Define a static maximum power draw for the group OR click a box to follow the existing MQTT dynamic load available topic for Shaper/Divert. 
Add members to the share group by mDNS name.
Select sharing Profile (Same current for all active stations, First station highest priority, Reduce all by same percentage, etc).
Select lost communications safe current

On each member station:
Enter mDNS name of Primary Station to allow Sharing
Select lost communications safe current

Primary would establish a http connection to each client, and send a claim to set the current and request status every xx seconds.
Each client would set the claim and respond with status.
Primary and Client would note group as responsive and countdown to the next heartbeat.

cjbarth:
The concept of a group could be extended to include sharing metering data to facilitate the export to CSV like #941 proposes.

cjbarth:
A simple FIFO method for doing this could also be implemented by adding in support to OpenEVSE to talk to the API of a meter.

---
Sample of how this might work:

Suppose there are 2 EVSE and one A/C unit: EV1, EV2, and AC. Also, there is 14 kW available to share among these three.

When EV1 requests power, it queries the API of a meter, and finds there is 14 kW available. It has a minimum draw of 4 kW. It counts 4 cycles (1 cycle per kW), and, if there is still 4 kW, starts drawing that. Every additional cycle it draws an additional 1 kW until it maxes out at 10 kW. checking the meter API each time to see that power is available.

When EV2 requests power, it queries the API of the meter and sees that there is 4 kW available. It has a minimum draw of 4 kW. Following the pattern above, it waits for cycles, checking each time that there is still 4 kW available. At that time, it sees there are 4 kW and draws that much.

[...]

PR #1147, opened 2026-07-07
Load sharing: enforce member failsafe, config validation, priority + rotation

RAR:
I ran this branch on a local bench (the divert_sim scenario suite plus the emulator + native-firmware integration harness with real mDNS) and posted the detailed findings on #940. This PR is the fix series for the five issues I hit, one commit per fix, each with the test that pins it.

**Commits:**

1. **Don't mutate role state on a rejected config POST** — `becomeMember()`/`resetRole()` ran before validation, so a rejected `POST /config` still flipped group membership and later posts had their `loadsharing_` fields stripped by the member guard. Role transition now happens after all validation passes.
2. **Duplicate peer add is idempotent** — mDNS auto-discovery racing a manual add is a normal sequence, not a client error; `POST /loadsharing/peers` for an existing peer now returns `200 {"msg":"already in group"}` and skips the reciprocal sync. *Contract change:* `test_add_peer_duplicate_rejection` is updated to the new behavior.
3. **Reject config where failsafe current exceeds the group budget** — one islanded member at `failsafe_safe_current=32` against `group_max_current=10` could exceed the group max on its own. Cross-field validation (incoming-else-stored) at config write.
4. **Enforce member failsafe with an EvseManager claim** — the important one. `checkMemberFailsafe()` set `_failsafe_active` but placed no claim, so a member that lost its controller kept charging uncapped and a manual override ran uncontested (32 A pilot with `/claims` empty). The poller now applies the same Limit-priority claim the allocation path uses (safe-current cap, or `Disabled` in `disable` mode) whenever failsafe engages. Since Limit (1100) outbids Manual (1000), an override is now capped at the failsafe current while islanded — pilot drops 32→6 in the test.
5. **Consult peer priority in scarcity subset selection** — the `priority` field (lower value = higher priority, per the `AllocationInput` contract) now orders subset selection ahead of the id tiebreak.
6. **Rotate equal-priority members under scarcity** — new `loadsharing_rotation_interval` (default 1800 s, 0 disables): when the budget can't cover all minimums, the front of each equal-priority run rotates on the interval so both vehicles make progress overnight instead of the lowest id starving the rest. Time is a parameter, so the allocator stays a pure, simulator-testable function.
7. **Failsafe-timing and override drills** — outcome-recording integration tests that reproduce the pre-fix findings and pass on fixed firmware.

**Verification on my bench:** divert_sim suite 65 passed (including the new priority + rotation scenarios); the new `test_loadsharing_fixes.py` 3/3 and drills 3/3; `test_charging.py` 24/24; `test_loadsharing_peer_status.py` went from 10/6 to 16/16 with these fixes. `test_loadsharing_peer_management.py` still has the pre-existing mDNS discovery flakiness you've been chasing — unrelated to this series as far as I can tell.

Happy to split any of this out, reorder, or rework to fit where the branch is heading. One open question worth deciding together at some point: how load-sharing claims should layer against the charge manager's timer-window claims from #1112, since both sit at Limit priority today.

### DIFF
--- a/divert_sim/data/scenarios/loadsharing_priority_wins.json
+++ b/divert_sim/data/scenarios/loadsharing_priority_wins.json
@@ -1,0 +1,44 @@
+{
+  "meta": {
+    "id": "loadsharing_priority_wins",
+    "title": "Load Sharing - priority selects the winner under scarcity",
+    "category": "loadsharing",
+    "profile": "default"
+  },
+  "group": {
+    "safety_factor": 1.0,
+    "failsafe_mode": "safe_current",
+    "failsafe_peer_assumed_current": 6.0
+  },
+  "simulation": {
+    "duration": 7200,
+    "tick_interval": 60,
+    "nominal_voltage": 240
+  },
+  "supply": {
+    "max_pwr": 2400,
+    "live_pwr": 0
+  },
+  "peers": [
+    {
+      "id": "evse-001",
+      "min_current": 6,
+      "max_current": 32,
+      "voltage": 240,
+      "priority": 10,
+      "ev": {"battery_capacity_kwh": 75, "initial_soc": 40, "max_charge_rate_kw": 7.2},
+      "initial": {"online": true, "vehicle": true},
+      "events": []
+    },
+    {
+      "id": "evse-002",
+      "min_current": 6,
+      "max_current": 32,
+      "voltage": 240,
+      "priority": 0,
+      "ev": {"battery_capacity_kwh": 75, "initial_soc": 40, "max_charge_rate_kw": 7.2},
+      "initial": {"online": true, "vehicle": true},
+      "events": []
+    }
+  ]
+}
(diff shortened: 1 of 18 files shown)


jeremypoulter:
@RAR are you able to look at the conflicts and test failures?

RAR:
Rebased onto the current jeremypoulter/issue940. The main adaptation was the member failsafe fix: with allocations now routed through the shaper, it applies the failsafe via shaper.setLoadSharingLimit() (safe_current cap, or force-disabled in disable mode) instead of a direct EvseManager claim, so it still outranks a manual override via the shaper's Safety-priority claim. The integration test now asserts on the Shaper claim, polling for it since the shaper applies the limit from its own task loop rather than synchronously.

The priority and rotation commits carried over on top of the new simulator scenarios without functional changes.

Local runs on the rebased branch: all 10 divert_sim load-sharing tests pass (the 8 on the base branch plus the priority and rotation scenarios), all 36 integration tests pass against the native build, and openevse_wifi_v1 compiles.

PR #1159, opened 2026-07-12
Fix silently dropped RAPI commands: LCD text, controller RTC sync, Ohm Connect

chris1howell:
## Problem

Since the load-sharing rework (8ffc467a, merged to master 2026-07-09) `EvseManager` owns a private `OpenEVSEClass _openevse` instance (needed so divert_sim can run multiple EVSE peers), and the **global `OpenEVSE` object is never `begin()`-ed**. `OpenEVSEClass` methods silently no-op when `_sender` is null, so everything still using the global object stopped working with no error:

- **Character LCD is completely dead** — `lcd.cpp` sends all `$F0`/`$FB`/`$FP` through `EvseManager::getOpenEVSE()`, which still returned the global. Verified on real hardware (openevse_wifi_v1_16mb, master build `851b3a60`): RAPI polling flows normally, zero `$F*` commands ever sent.
- **Controller RTC never gets set from NTP** (`time_man.cpp` `OpenEVSE.setTime/getTime`)
- **Controller time polling + WiFi state signalling** (`input.cpp`)
- **Ohm Connect sleep/enable** (`ohm.cpp`)
- **Legacy timer import** (`legacy_support.cpp`)

## Fix

- `EvseManager::getOpenEVSE()` returns the monitor-initialised `_openevse` member instead of the global.
- Migrate the five remaining direct users of the global `OpenEVSE` object to `evse.getOpenEVSE()`.

## Testing

- `pio run -e openevse_wifi_v1` builds clean.
- Root cause confirmed on hardware by A/B: master build shows no `$F*` traffic in the RAPI spy (`/evse`) while all monitor polling works; `$F0 1` from the console proves the LCD/controller side is healthy.

### DIFF
--- a/src/evse_man.h
+++ b/src/evse_man.h
@@ -579,9 +579,11 @@ class EvseManager : public MicroTasks::Task
       return _sender;
     }
 
-    // Get the OpenEVSE API
+    // Get the OpenEVSE API. Must be the instance the monitor initialised with
+    // the RAPI sender; the global OpenEVSE object is never begin()-ed so its
+    // commands are silently dropped.
     OpenEVSEClass &getOpenEVSE() {
-      return OpenEVSE;
+      return _openevse;
     }
 
     // Register for events
--- a/src/input.cpp
+++ b/src/input.cpp
@@ -169,7 +169,7 @@ handleRapiRead()
 {
   Profile_Start(handleRapiRead);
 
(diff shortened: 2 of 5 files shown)


jeremypoulter (approved):
Good spot, probably need to check that the old global insurance is just not created to avoid this, sorry I missed it

PR #1165, opened 2026-07-18
WebSocket endpoints streamed device data without any credential check

### LINKED ISSUE #1: ESP32: non-blocking RAPI & "controller not found message"
If possible the RAPI loop should be non-blocking and if the open evse controller is not detected a message should be displayed:

"OpenEVSE controller not detected, please check connections" 

- [x] RAPI comms non-blocking
- [x] Display message in UI when OpenEVSE is not connected

chris1howell:
# ws_auth branch changes

Fixes a **missing-authentication** flaw in the ESP32 web server: three
WebSocket endpoints streamed device data without any credential check, even
when HTTP Basic auth was configured and correctly enforced on every REST route.
Also hardens the Basic-auth comparison against a timing side channel.

Branched from `master`. Verified against a live station (`openevse-c620`,
`openevse_wifi_v1_16mb`).

Source of the report: coordinated-disclosure advisory
"Unauthenticated WebSocket Endpoints Expose Device Telemetry, RFID Credentials,
and Serial Console Streams" (CWE-306 / CWE-200 primary, CWE-208 secondary).

## Background / the problem

Every REST handler runs through `requestPreProcess()`, which performs the HTTP
Basic-auth check. The three WebSocket endpoints registered only
`onFrame()` / `onConnect()` callbacks and therefore never passed through that
helper. In the underlying `ArduinoMongoose 0.0.26` dispatch, the WebSocket
handshake shares its branch with ordinary HTTP requests and the application
auth hook is only invoked when an `onRequest` handler is set — for a
WebSocket-only route it was not, so the handshake completed unauthenticated.

Result: any host able to reach the device's HTTP port could complete a
WebSocket handshake with **no credentials** and passively receive:

| Endpoint | Data exposed |
|---|---|
| `/ws` | Full device status on connect + all broadcast events, **including RFID card UIDs at enrollment** |
| `/evse/console` | Live RAPI serial TX/RX stream to/from the EVSE controller |
| `/debug/console` | Internal firmware debug log stream |

The RFID UID exposure is the highest-impact element: RFID authorization in this
firmware is a plaintext UID equality check (`rfid.cpp`), so a captured UID is
directly replayable to gain charging authorization.

Secondary: `MongooseHttpServerRequest::authenticate()` compares credentials with
`strcmp()`, which short-circuits on the first differing byte and is therefore
not constant-time (CWE-208). Low severity — network jitter dominates the
per-byte delta — but it lives in the same auth path.

### 1. Authenticate the WebSocket handshake

A single gate function is registered via `->onRequest()` on all three WebSocket
endpoints:

```c
void onWsAuthenticate(MongooseHttpServerRequest *request)
{
  if(!isAuthenticated(request)) {
    request->requestAuthentication(esp_hostname);
  }
}
```

Because `onRequest` runs during `MG_EV_WEBSOCKET_HANDSHAKE_REQUEST` — **before**
the `101 Switching Protocols` response — a failed check calls
`requestAuthentication()`, which sends a `401` and sets `MG_F_SEND_AND_CLOSE`.
Mongoose then skips handshake completion entirely (`mongoose.c`, the
`MG_EV_WEBSOCKET_HANDSHAKE_REQUEST` handler only sends the 101 and fires
`HANDSHAKE_DONE` when the connection is not flagged for close). No `onConnect`
fires and no data is ever pushed to an unauthenticated client.

**Why the handshake-request stage, not `onConnect`:** the advisory's suggested
`onConnect` gate fires at `HANDSHAKE_DONE` (after the 101), the
`MongooseHttpWebSocketConnection` class in this library version has no `close()`
method, and the two console endpoints have no `onConnect` handler at all.
Gating at the request stage is one uniform mechanism that covers all three
endpoints before any data flows.

### 2. Single source of truth for auth + constant-time comparison

The Basic-auth check was factored out of `requestPreProcess()` into a shared
`isAuthenticated()` used by both the REST path and the WebSocket gate:

```c
bool isAuthenticated(MongooseHttpServerRequest *request)
{
  if(net.isWifiModeApOnly() || www_username == "") {
    return true;                       // AP provisioning / no admin password set
  }
  MongooseString authHeader = request->headers("Authorization");
  if(!authHeader) {
    return false;
  }
  mg_str hdr = authHeader.toMgStr();
  char user_buf[64] = {0};
  char pass_buf[64] = {0};
  if(0 != mg_parse_http_basic_auth(&hdr, user_buf, sizeof(user_buf),
                                   pass_buf, sizeof(pass_buf))) {
    return false;
  }
  return credentialsMatch(www_username.c_str(), user_buf) &&
         credentialsMatch(www_password.c_str(), pass_buf);
}
```

`credentialsMatch()` compares full length without an early exit, folding in the
length difference, so the check no longer leaks how many leading bytes matched
(CWE-208). Parsing reuses the library's tested `mg_parse_http_basic_auth()`;
only the comparison changed. The buffers are zero-initialised (the library
original could leave `pass` uninitialised if the header had no colon).

The `strcmp()` itself lives in `ArduinoMongoose`, a **PlatformIO registry
dependency** (`jeremypoulter/ArduinoMongoose@0.0.26`) that cannot be patched
persistently in this repo — so the fix was applied at the firmware layer, where
the firmware no longer routes credential comparison through the library's
`strcmp`. A constant-time fix upstream in ArduinoMongoose remains worthwhile.

## Not changed, and why

- **RFID UID broadcast (advisory §6.2, defence-in-depth).** Not implemented.
  Masking the enrolled UID would break the "Add RFID" workflow — the enrolling
  admin's browser needs the full UID to save the card — and per-connection
  delivery is not feasible because enrollment is triggered by a stateless REST
  call (`/rfid/add`) with no associated WebSocket. After fix #1 this data is
  already restricted to authenticated clients; limiting it to only the enrolling
  session is a GUI/product decision.

## Compatibility notes

- **The GUI keeps working.** Browsers automatically resend cached HTTP Basic
  credentials on same-origin requests, including the WebSocket handshake
  (`GET` upgrade). This is distinct from the JS `WebSocket` API's inability to
  set *custom* headers, and is confirmed on-device (see below).
- **Unprovisioned devices are unaffected.** With no admin password set (or in
  AP-only provisioning mode) `isAuthenticated()` returns `true`, so WebSockets
  stay open exactly as before.

## Testing

Build: `pio run -e openevse_wifi_v1` and `-e openevse_wifi_v1_16mb` both clean.

On-device (`openevse-c620`, live), a temporary admin password was set, the
before/after captured, then the password cleared to restore the original open
state:

| Test (admin password set) | Before fix | After fix |
|---|---|---|
| REST `/status`, no creds | 401 | 401 |
| WS `/ws`, no creds | **101 + full status leak** | **401** |
| WS `/evse/console`, no creds | **101** | **401** |
| WS `/debug/console`, no creds | **101** | **401** |
| WS `/ws`, valid creds | — | 101 + status data |
| WS `/evse/console` + `/debug/console`, valid creds | — | 101 |
| WS `/ws`, wrong user or wrong password | — | 401 |

With the password cleared, `/ws` returns to `101` (open) — confirming no
regression for unprovisioned/AP-provisioning devices.


RAR:
Reviewed against ArduinoMongoose 0.0.26 source and the full route table — the fix is sound and I'd take it. Confirmed the core claims: `authenticate()` really does `strcmp()` (MongooseHttpServer.cpp), the three WS routes are the *only* ones registering `onFrame`/`onConnect` so coverage is complete, and gating at `MG_EV_WEBSOCKET_HANDSHAKE_REQUEST` works because `requestAuthentication()` sets `MG_F_SEND_AND_CLOSE` + `_responseSent` before the 101. Nice catch on the zero-init of `pass_buf` too — `mg_parse_http_basic_auth()` returns success when `sscanf` matches only the username (no colon) and leaves `pass` untouched, so `{0}` is what fails that case closed.

One thing worth fixing before merge, in `credentialsMatch()`:

```c
unsigned char diff = (unsigned char)(la ^ lb);
```

Truncating the length difference to an `unsigned char` drops all but the low 8 bits. If a stored credential's length differs from the supplied one by a nonzero multiple of 256 *and* the overlapping bytes match, `diff` starts at 0 and stays 0 → false accept. The supplied side is bounded to 63 by `pass_buf[64]`, but `la = strlen(www_password.c_str())` is unbounded config, so e.g. a 300-char stored password vs a 44-char input matching its first 44 bytes (`0x12C ^ 0x02C = 0x100`) would authenticate. Only reachable with a pathological 256+ char stored password so severity is low, but it's a latent bypass inside the auth path and the fix is free — keep the length diff full-width:

```c
size_t diff = la ^ lb;
size_t n = la < lb ? la : lb;
for (size_t i = 0; i < n; i++) {
  diff |= (unsigned char)(a[i] ^ b[i]);
}
return 0 == diff;
```

Minor, non-blocking:

- `ws_auth_changes.md` is the PR description committed at repo root — probably belongs in the PR body or under `docs/` rather than as a permanent file.
- `n = min(la, lb)` plus the `strlen`s means the compare isn't strictly constant-time w.r.t. length, but that's fine for this threat model (you note network jitter dominates anyway).

Net: endpoint coverage, the handshake-stage design, and the before/after table all check out. Approve after the `size_t diff` one-liner.

chris1howell:
Both review items are addressed and pushed to origin/ws_auth (commit 964dd793):

The blocking fix — credentialsMatch() now keeps the length difference full-width (size_t diff = la ^ lb; instead of truncating to unsigned char). The analysis was right: la = strlen(www_password.c_str()) is unbounded config, so a 256+ char stored password whose length delta from the input landed on a multiple of 256, with matching overlapping bytes, would have started diff at 0 and false-accepted. Full-width diff closes that. The comment records why.

The minor doc note — moved ws_auth_changes.md → docs/ws_auth.md (100% rename), matching the docs-in-docs/ convention.

The other non-blocking note — that the compare isn't strictly constant-time w.r.t. length because of the strlens and min(la,lb) — I left as-is, since the reviewer explicitly accepted it for this threat model (network jitter dominates), consistent with what the doc already says.

PR #1166, opened 2026-07-19
fix(web): HTML-escape reflected RAPI parameter in /r console (XSS)

RAR:
## Summary

The `/r` RAPI web console reflects the user-supplied `rapi` query parameter (and the controller's response string) into a `text/html` response without output encoding. A crafted link such as `/r?rapi=<script>…</script>` therefore executes attacker-controlled script in the origin of the device. Both the success and error branches reflect the raw value, so it fires regardless of the `isRapiCommandBlocked()` allowlist outcome.

This came in via a coordinated-disclosure report (CWE-79). This PR addresses that reflected-XSS finding only — see "Scope" below.

## Fix

- Add a small file-local `html_escape()` helper (escapes `& < > " '`).
- Escape the reflected `rapi` parameter in both the success and error HTML branches, plus the dynamic controller response string, before concatenation into the response.

The JSON branch (`?json=1`) is unchanged: it returns `application/json`, not HTML, so it is not an XSS sink. (It does build JSON by string concatenation, which is a separate correctness nit worth a follow-up, but not part of this security fix.)

## Verification

Built and run on the `native_openevse` host target:

```
# before: reflects raw <script>…</script>
# after:
$ curl -s 'http://localhost:8000/r?rapi=<script>alert(1)</script>' | grep -c '<script>alert(1)</script>'
0
# the probe now reflects as: &lt;script&gt;alert(1)&lt;/script&gt;
```

`pio run -e native_openevse` → SUCCESS.

## Scope

The disclosure bundled two other items that are intentionally **not** in this PR because they are product decisions rather than clear bugs:

- **Auth-off-when-no-password** (`requestPreProcess`): flipping to fail-closed would lock out existing no-password devices and is at odds with the empty-username semantics kept in the recent WebSocket-auth work. Better handled as a login-page / first-run-setup discussion.
- **Arbitrary-URL OTA** (`/update`): needs a signing/allowlist decision. Its practical severity drops substantially once this XSS is fixed and the endpoint sits behind auth.

Happy to follow up on either separately.

### DIFF
--- a/src/web_server.cpp
+++ b/src/web_server.cpp
@@ -204,6 +204,27 @@ bool isPositive(MongooseHttpServerRequest *request, const char *param) {
   return paramFound >= 0 && (0 == paramFound || isPositive(String(paramValue)));
 }
 
+// -------------------------------------------------------------------
+// Escape a string for safe inclusion in an HTML response body, so that
+// reflected user input cannot be interpreted as markup or script.
+// -------------------------------------------------------------------
+static String html_escape(const String &input) {
+  String out;
+  out.reserve(input.length());
+  for(size_t i = 0; i < input.length(); i++) {
+    char c = input.charAt(i);
+    switch(c) {
+      case '&':  out += F("&amp;");  break;
+      case '<':  out += F("&lt;");   break;
+      case '>':  out += F("&gt;");   break;
+      case '"':  out += F("&quot;"); break;
+      case '\'': out += F("&#39;");  break;
+      default:   out += c;           break;
+    }
+  }
+  return out;
+}
+
 //---------------------------------------------------------------------
 // Build status data
 // --------------------------------------------------------------------
@@ -1111,9 +1132,9 @@ handleRapi(MongooseHttpServerRequest *request) {
       if (json) {
         s = "{\"cmd\":\""+rapi+"\",\"ret\":\""+rapiString+"\"}";
       } else {
-        s += rapi;
+        s += html_escape(rapi);
         s += F("<p>&gt;");
-        s += rapiString;
+        s += html_escape(rapiString);
       }
     }
     else
@@ -1135,7 +1156,7 @@ handleRapi(MongooseHttpServerRequest *request) {
       if (json) {
         s = "{\"cmd\":\""+rapi+"\",\"error\":\""+errorString+"\"}";
       } else {
-        s += rapi;
+        s += html_escape(rapi);
         s += F("<p><strong>Error:</strong>");
         s += errorString;
       }


chris1howell:
## Review

The fix is correct and I'd merge it. The escaping is complete for the context it's used in (text node, and the quote escapes mean it stays safe if anyone later prefills the form `value=`), both HTML branches are covered, and `errorString` being left unescaped is fine — it's a fixed set of `F()` literals from the `RAPI_RESPONSE_*` ladder, not attacker data.

Worth noting for prioritization: `requestPreProcess()` only authenticates when `www_username != ""`, so on a default no-password device this is a **pre-auth** reflected XSS. That argues for merging now and handling the rest as follow-ups.

## Changes I'd ask for before merge

**1. Fix the JSON branch too, not later.** It's three lines and it's in the same function:

```cpp
s = "{\"cmd\":\""+rapi+"\",\"ret\":\""+rapiString+"\"}";
```

`?json=1&rapi=$GS"` produces malformed JSON, and `rapi` can inject arbitrary keys into the object the GUI parses. Not XSS (content type is `application/json`), but it's the same untrusted-value-into-a-structured-format bug, a scanner will re-flag it as a "new" finding after this PR closes, and leaving it open invites a second disclosure round-trip. `StaticJsonDocument<256>` + `serializeJson(doc, s)` kills it.

**2. `handleHttpsRedirect()` has the identical pattern and isn't touched.** Around line 1214:

```cpp
url += request->host().toString();
url += request->uri().toString();
```

...concatenated into a `meta http-equiv="Refresh"` URL, an `href="..."` attribute, *and* a `Location:` header. Browsers percent-encode `<`, `>`, `"` in the path so the drive-by XSS is weak, but `host()` is fully controlled by any non-browser client and it feeds a redirect target — that's a host-header open redirect regardless of escaping. At minimum this should be escaped in the same PR (or explicitly called out in "Scope" so it doesn't look like it was missed). `handleNotFound()` builds its URL from `net.getIp()`, so that one is fine.

## Non-blocking nits

- **`out.reserve(input.length())` under-reserves.** Arduino `String::concat` reallocs to exact size on each append, so every escaped character triggers a fresh realloc — O(n) heap churn inside the network loop on a device where `free_heap` is already a status field. `input.length() + 16` (or `* 2` for the short strings this sees) makes the common path allocation-free.
- **Loop index is `size_t`, `charAt()` takes `unsigned int`.** Harmless, but it's a narrowing conversion that'll warn under `-Wconversion` on the native target. `unsigned int i` matches the rest of the file.
- **Add a unit test.** You just set the precedent for exactly this in `ff8780f` — `src/ota_url_allow.cpp` + `test/test_ota_url_allow/` under `native_test`/doctest. `html_escape` is a pure function with no framework dependency; pulling it into a small TU with a doctest covering `&`/`<`/`>`/`"`/`'`, empty string, and no-op input costs almost nothing and means the next refactor can't silently regress it. Right now the only verification is a manual `curl` in the PR description.
- **`<meta charset="utf-8">` on the `/r` page.** The page declares no charset; escaping-bypass-via-charset-sniffing is a legacy-browser concern only, but it's one string literal.

## Follow-ups worth filing (out of scope here)

- **`/r` is a state-changing GET with no CSRF protection.** `` from any page disables charging on a no-password device. Arguably higher practical impact than the XSS this PR fixes, and unaffected by it.
- **`X-Content-Type-Options: nosniff`** in `requestPreProcess()` — one line, covers every endpoint, and is the cheap insurance behind the "JSON isn't an XSS sink" argument.

I agree with punting auth-off-when-no-password and arbitrary-URL OTA — those are product decisions, and #1167 already owns the OTA half.

chris1howell (approved):

PR #1167, opened 2026-07-19
Signed Firmware

chris1howell:
# signed_ota branch changes

Addresses the coordinated-disclosure report "Chained Reflected XSS, Insecure
Default Configuration, and Unsigned Arbitrary-URL Firmware Update" (F1 insecure
default auth, F2 unsigned arbitrary-URL OTA, F3 reflected XSS in `/r`).

Branched from `master`.

> **Status (after on-device testing):**  F2a (URL allowlist) are
> done and sound. F2b **signed OTA is not production-ready** — the on-device
> verifier is proven working, but delivering a correctly-signed image hits a
> core-library flash-erase bug on the exact-size streaming path. Keep signing
> **disabled** (no `OTA_SIGNING_KEY` secret) until this is resolved. See the
> Testing section for detail.

## Decisions (given the "no forced password" constraint)

Forcing a web password was explicitly rejected (support burden when users
forget). So **F1 is intentionally left as-is** — the device stays open on the
LAN by default. The report's real teeth are the *chain*, so this branch breaks
the chain instead of the default:

- **F2** (unsigned arbitrary-URL OTA) gets two independent controls: an
  HTTPS-only GitHub allowlist on the URL-fetch path, and **signed firmware
  images** verified on-device. Signing is the one control that closes the
  unauthenticated-reflash path *even on a password-less, open-by-default
  device*, because a LAN attacker cannot produce a validly-signed image.

Residual, by design: with no password and no signing, a LAN host can still
upload an image via the authenticated-but-open file-upload path. Signed OTA
closes that on the boards that support it (see core-3.x note); the 4MB boards
retain the LAN-trust posture you're choosing.

## F2a — URL-fetch allowlist (`src/http_update.cpp` / `.h`)

`http_update_url_allowed()` requires `https://` and a host of `github.com` or
`*.githubusercontent.com` (GitHub redirects release downloads to the
`githubusercontent` CDN). It is enforced at the top of `http_update_from_url()`,
which runs on the initial fetch **and** on every 30x redirect target, so a
device can only pull firmware from a trusted origin. Userinfo (`user@host`) is
stripped to defeat `github.com@evil.example` tricks. Rejected fetches return
`HTTP_UPDATE_ERROR_URL_NOT_ALLOWED`.

The **file-upload** path (choosing a `.bin` in the GUI) is unchanged — this is
how local/dev builds are flashed, and it satisfies "updates come from OpenEVSE
GitHub" (the browser downloads from GitHub, then uploads).

The allowlist is one editable function; relax or make it configurable upstream
if self-hosted update URLs need to be supported.

## F2b — Signed OTA (`src/ota_signing.*`, `scripts/`, CI)

Uses the Arduino-ESP32 **core-3.x** `Update` library's built-in verifier
(`UpdaterRSAVerifier` + `Update.installSignature()`), so the on-device code is
small.

- `src/ota_signing.cpp/.h` embeds the RSA-2048 **public** key and, when built
  with `-DREQUIRE_SIGNED_OTA`, installs the verifier. `http_update_start()` —
  the single `Update.begin()` choke point for **both** the URL and file-upload
  paths — calls it just before `begin()`, and fails closed if the verifier
  can't be installed or the Content-Length is unknown (needed to locate the
  signature trailer).
- Image format (matched exactly to the core verifier): `firmware || signature ||
  zero-pad`, a fixed **512-byte** trailer. The signature is **RSA-2048 PSS**
  over `SHA-256(firmware)`, MGF1-SHA256, **salt length 222** (`key_len − 32 −
  2`), 256 bytes, zero-padded to 512. `begin(total)` is given
  `len(firmware)+512` (the served Content-Length).
- `scripts/sign_firmware.py` produces exactly this (verified: a signed test
  image validates against the public key with device-identical PSS params, and
  a one-byte tamper is rejected). `scripts/generate_ota_key.sh` regenerates the
  keypair for rotation.

### How "on in CI, off locally" works

Enforcement is decided by the firmware **running on the receiving device**, not
by how an incoming image was built:

- Local/dev builds have no flags → accept unsigned images (flash over USB as
  usual, or OTA a dev build to a dev device).
- CI release builds (once enabled) compile with `-DUPDATE_SIGN
  -DREQUIRE_SIGNED_OTA` and the image is signed → a device running such a build
  rejects any unsigned OTA. To put an unsigned local build on such a device, use
  USB/serial.

### CI (dormant until you opt in)

`.github/workflows/build.yaml` signs and enforces **only** when the
`OTA_SIGNING_KEY` secret is present *and* the env is in `SIGNED_OTA_ENVS`.
Until the secret is added, builds are byte-for-byte unchanged (unsigned) — so
merging this does not alter the fleet on its own.

### core-3.x limitation

The verifier API exists only in the **core-3.x / pioarduino / IDF5** core (the
16MB+ boards, e.g. `openevse_wifi_v1_16mb`). The older-core 4MB envs
(`openevse_wifi_v1`, …) have no `Updater_Signing.h` and **cannot** verify
signatures via this mechanism; `ota_signing.cpp` `#error`s if the flag is set on
such an env, and `SIGNED_OTA_ENVS` lists only signed-capable envs. 4MB boards
rely on the URL allowlist + F3 fix.

## Operator action items (to actually turn signing on)

1. The public key is already embedded (`src/ota_signing.cpp`). The matching
   **private** key was generated during this work and is **not** in the repo
   (`*.pem` is gitignored). Store its PEM as the GitHub Actions secret
   `OTA_SIGNING_KEY`, or regenerate with `scripts/generate_ota_key.sh` and
   replace the embedded public key.
2. Extend `SIGNED_OTA_ENVS` as more core-3.x boards are added.
3. **Rollout is a one-way door:** once a device runs a signed-required build,
   all future OTAs must be signed. If the key is lost, recovery is USB-only.
   Consider a phased rollout (ship *signed* images before flipping
   `REQUIRE_SIGNED_OTA`), and keep the private key backed up and access-limited.

## Testing

- Builds (16MB, core-3.x): unsigned default **and** `-DUPDATE_SIGN
  -DREQUIRE_SIGNED_OTA` both clean.
- Build (4MB, old core): unsigned default clean; signed flags correctly
  `#error` out (guard verified).
- Signing round-trip verified off-device: `sign_firmware.py` output validates
  against the public key with the device's exact PSS parameters (salt 222,
  256-byte sig, 512-byte trailer); a tampered image is rejected.

### On-device signed-OTA test (openevse-c620, 16MB) — verifier PASS, delivery BLOCKED

An enforcing build (`-DUPDATE_SIGN -DREQUIRE_SIGNED_OTA`) was flashed to c620
and images were pushed over the web `/update` file-upload path:

- **Verifier works.** An **unsigned** image is rejected with
  `Update failed: 14 (Signature Verification Failed)` — the on-device mbedtls
  RSA-PSS verifier ran and correctly rejected it. This validates the public
  key, the PSS parameters, and the `sign_firmware.py` trailer format against
  real silicon.
- **Delivery is blocked by a core-library flash-erase bug.** To verify a
  *correctly-signed* image the exact image size must be given to
  `Update.begin()` (so the 512-byte signature trailer can be located). With the
  exact size, the core Update library flushes the final partial sector *during*
  `write()` (Updater.cpp:791, `_bufferLen == remaining()`) instead of at
  `end()`, and that erase fails with `Update failed: 2 (Flash Erase Failed)` in
  the streaming callback context. Without the exact size the erase succeeds but
  the signature hash covers the wrong byte range, so a signed image can't be
  verified that way either.

**Status: signed OTA is NOT production-ready.** The verifier is proven, but
there is no working end-to-end delivery of a signed image on this core via the
OpenEVSE streaming update path. The `?size` file-upload hint that exposed this
was reverted. Resolving it needs one of:

1. Root-cause the exact-size final-sector erase (possibly a flash/cache-context
   issue with erasing inside the network callback) and fix delivery so
   `end()`-time flushing is used, or
2. Verify the signature in OpenEVSE code instead of via the core
   `installSignature()` API — hash incrementally over the streamed firmware,
   hold back the last 512 bytes, and RSA-PSS-verify with mbedtls at `end()` —
   avoiding the exact-size `begin()` requirement entirely.

Until then, keep signing **disabled** (no `OTA_SIGNING_KEY` secret).
F2 URL allowlist are unaffected and stand on their own.

Note: because an enforcing build rejects every image it can't verify, c620 was
left on the enforcing test build and recovered by **USB flash** of a clean
non-enforcing image (there is no working signed-OTA path to recover it over the
network — the same delivery bug).


RAR (commented):
Review focuses on **F2a (URL allowlist)** and **F2b (signed OTA)** plus the CI/doc changes.

**F3 (reflected XSS) is handled in #1166** — the `htmlEscape()` helper, the `handleRapi()` escaping, the baseline headers in `requestPreProcess()`, and the `/r` CSP overlap with that PR. Suggest dropping the F3 hunks from this branch and rebasing on #1166 so the two don't diverge, leaving this PR to be just the OTA controls.

---

## 🔴 F2a allowlist is bypassable: `https://github.com:<x>@evil.example/…`

`http_update_url_allowed()` (`src/http_update.cpp:50-66`) extracts the host by scanning for the first `/`, `?`, **or `:`**, and only *then* strips userinfo at the last `@`:

```cpp
for(int i = start; i < url.length(); i++) {
  char c = url[i];
  if(c == '/' || c == '?' || c == ':') { end = i; break; }   // stops at ':'
}
String authority = url.substring(start, end);
int at = authority.lastIndexOf('@');                          // never sees the '@'
String host = (at >= 0) ? authority.substring(at + 1) : authority;
```

A `:` can appear **inside userinfo**, before the `@`. So for:

```
https://github.com:anything@evil.example/firmware.bin
```

`authority` is truncated to `github.com` at the first `:`, `lastIndexOf('@')` returns −1, and `host` becomes `github.com` → **allowed**.

But Mongoose parses the same URL per the URL spec — everything before `@` is userinfo — and connects to **`evil.example`** (`mg_parse_uri` in `mongoose.c`, then `mg_connect_http_base`). HTTPS here runs with `ssl_ca_cert = "*"` (no cert verification), so the attacker's host needs no valid cert. This defeats the allowlist entirely — the exact `github.com@evil` trick the comment says it prevents, just with a `:` in front.

(The plain `https://github.com@evil.example/` form *is* caught — no `:` before the `@`, so the scan runs to `/` and the userinfo strip works. Only the `:`-prefixed variant slips through.)

**Fix** — locate the authority on `/` and `?` only, strip userinfo at the last `@`, *then* cut the port:

```cpp
int end = url.length();
for(int i = start; i < (int)url.length(); i++) {
  char c = url[i];
  if(c == '/' || c == '?') { end = i; break; }   // NOT ':'
}
String authority = url.substring(start, end);
int at = authority.lastIndexOf('@');
String host = (at >= 0) ? authority.substring(at + 1) : authority;
int colon = host.indexOf(':');                    // strip port from the real host
if(colon >= 0) host = host.substring(0, colon);
```

Worth adding a unit/round-trip test with these vectors: `github.com:1@evil.example`, `github.com@evil.example`, `evil.example/github.com`, `raw.githubusercontent.com`, `notgithub.com`.

## 🟡 F2b: file-upload path can't supply the correct size for the verifier

`handleUpdateUpload` (`src/web_server_update.cpp:119`) calls `http_update_start(filename, request->contentLength())`. For a multipart upload, `contentLength()` is the size of the **whole multipart envelope** (boundaries + part headers + the `.bin`), not the firmware image. The verifier needs `len(firmware)+512` exactly to locate the 512-byte trailer, so — independent of the flash-erase bug you documented — the file-upload path structurally can't feed it the right size. Whoever picks up F2b will need to parse the part length here, not just fix the erase. Consistent with the "keep signing disabled" status; noting it so it isn't lost.

## 🟢 Looks good

- Verifier install + `total == 0` guard both **fail closed** (`http_update.cpp:252-264`).
- The `#error` guards in `ota_signing.cpp` (needs `UPDATE_SIGN`, needs `Updater_Signing.h`) prevent silently building a non-verifying "signed" image on the 4MB core.
- CI gating (`SIGN_ENABLED` + `SIGNED_OTA_ENVS`) keeps released images byte-identical until the secret exists; `PLATFORMIO_BUILD_FLAGS` isn't otherwise used in the workflow, so nothing gets clobbered.
- `sign_firmware.py` PSS params (salt 222 = `key_len − 32 − 2`, 256-byte sig, 512-byte trailer) match the mbedtls verifier; the mutually-exclusive `--key`/`--key-env` and RSA-type check are clean.

## 🔵 Minor

- **`signed_ota_changes.md`** — this is essentially the PR description plus a timestamped status memo ("NOT production-ready", the c620 recovery note). It'll rot in-tree once F2b is resolved. Prefer keeping it in the PR body or under `docs/` rather than as a tracked root file.
- **`for(int i = start; i < url.length(); ...)`** compares signed `int` to an unsigned length — harmless here, but a compiler warning; cast or use `size_t`.

---

**Recommendation:** the allowlist bypass should be fixed before merge — as written, F2a provides no protection against a crafted URL, and it's a small self-contained change. F2b is correctly shipped disabled, so its file-upload gap can be a follow-up. Everything else is in good shape once F3 is rebased onto #1166.

chris1howell:
🔴 Allowlist bypass — fixed
 a : inside userinfo truncated the host before the @ was seen. I extracted the logic into a pure, framework-free unit (src/ota_url_allow.cpp → ota_url_host_allowed()) with spec-correct parsing — authority ends at the first / or ? (not :), userinfo stripped at the last @, then the port stripped at the host's :. Scheme matching is now case-insensitive too. http_update_url_allowed() is a thin String wrapper over it.

I added test/test_ota_url_allow/ (doctest / native_test) covering all the vectors plus look-alikes and wrong-scheme. The native_test toolchain isn't on this Windows box (CI runs it on Linux), so I validated the algorithm locally with a faithful Python port: all 27 vectors pass, including https://github.com:1@evil.example/… → rejected.

🟡 F2b file-upload size gap — noted, not fixed
Correct that it's structural (contentLength() is the whole multipart envelope), separate from the erase bug. Signed OTA is shipped disabled, so I captured this in the doc as a second thing F2b's owner must address, alongside the erase issue — not silently lost.

F3 — dropped
Restored web_server.cpp to the branch base (verified byte-identical), so this branch no longer touches it and won't diverge from #1166. This PR is now just the OTA controls.

🔵 Minor — both done
Moved signed_ota_changes.md → docs/signed_ota.md, trimmed to OTA-only and updated to reflect the allowlist fix (out of the repo root).
The signed/unsigned int vs length comparison is gone — the new unit uses std::string::size_type throughout.
🟢 The "looks good" items are untouched
Verifier fail-closed guards, the ota_signing.cpp #error guards, CI gating, and sign_firmware.py PSS params all stayed as-is.

Both cores build clean (16MB and 4MB). One cosmetic wart: the commit got split into two same-titled commits because my first git add aborted on the already-removed doc path — harmless, both are pushed and the branch is consistent. Everything's on origin/signed_ota, in sync. The one thing still outstanding from earlier is unchanged: c620 is on the enforcing test build awaiting your USB reflash.

PR #1168, opened 2026-07-19
feat(auth): browser login page + session cookie (F1 backend)

RAR:
## Summary

Firmware backend for a browser **login page with "remember me"**, addressing finding **F1** (auth silently disabled by default; browser-cached HTTP Basic credentials auto-attach to same-origin requests) of the chained-XSS/insecure-defaults/OTA advisory. This is the maintainer-preferred direction — a real login/session rather than a forced password (no support burden, no lockouts).

Builds directly on the `isAuthenticated()` helper merged in #1165: this PR **extends it in place** with a session-cookie branch, so both the REST path (`requestPreProcess`) and the WebSocket gate (`onWsAuthenticate`) accept a logged-in browser session with no repoint.

**HTTP Basic auth is preserved** for machine clients (Home Assistant, MQTT, the app, scripted API) — nothing there changes.

## How it works

- **Stateless HMAC-signed cookie**: `oevse_session = v1.<exp>.<HMAC-SHA256(secret, "v1."+exp)>`. No server-side session table.
- **`server_secret`**: 32 random bytes (`esp_random`), stored as a `ConfigOptSecret` in the existing EEPROM config (masked from `GET /config`). **Rotated whenever the web password changes**, which invalidates every outstanding session — a free "sign out everywhere".
- **Cookie flags**: `HttpOnly`, `SameSite=Strict`, `Path=/`; remember-me sets `Max-Age` (30 d) vs a session cookie (6 h). (`Secure` is left TLS-gated for a follow-up, since most units are plain-HTTP LAN.)
- **CSRF**: cookie-authenticated non-GET requests must also carry `X-Requested-With: OpenEVSE` (the SPA sets it; a cross-origin form can't). Basic-auth clients never send our cookie, so they're unaffected.
- **`POST /login`**: constant-time credential check, escalating lockout after 5 failures. **`POST /logout`** clears the cookie.

## Scope / what's not here yet

- **GUI is a companion PR (in progress)** in `gui-nightshift`: the login route, the 401→login interceptor, the logout control, and the **first-run "set a password (recommended) / skip" wizard nudge** — that nudge is the part that actually turns F1's silent-insecure-default into an informed choice.
- `Secure`-cookie-when-TLS and a host-side sign/verify round-trip are noted follow-ups.

## Validation

- **Unit tests** (host `native_test`, doctest): portable HMAC-SHA256 against the **RFC 4231** known-answer vector; token mint/verify with tamper, expiry, wrong-secret, version, and garbage rejection; constant-time compare; cookie parsing incl. userinfo/substring-key edge cases. All green.
- **Builds clean**: `openevse_wifi_v1` (94.2% flash) and `native_openevse`.
- **Live end-to-end auth flow is pending on-hardware validation.** The `native_openevse` host target runs in AP-only mode (`isWifiModeApOnly()` is true with no STA link), so `isAuthenticated()` is open by design there — the gate can't be exercised on the host build. Needs a STA-connected unit with a password set.

Crypto note: the HMAC-SHA256 is a small vendored public-domain implementation compiled into **both** host and device (no `#ifdef` divergence) specifically so the token codec is host-testable with a real KAT; the device could swap to mbedTLS behind the same one-line function later.

@chris1howell — this is the F1 half; complements your F2 work in #1167 and my F3 fix in #1166. Review welcome, especially the cookie/CSRF model and the `server_secret` rotation.


chris1howell:
## Highlights

- **The vendored HMAC-SHA256 is right.** I compiled it standalone and differential-tested `hmac_sha256_hex()` against Python's `hmac`/`hashlib` on 155 vectors (keys 0–200 bytes including >block-size, messages 0–400 bytes). All match. The RFC 4231 KAT in the suite passes, and `extern "C"` linkage on `sha256.h` is correct — I confirmed by compiling `sha256.c` as actual C and linking against the C++ side, which is the failure mode that a native-only test can mask.
- **All 10 unit tests pass** (18 assertions). `session_token_verify()` fails closed on garbage, tamper, wrong secret, wrong version, and short input.
- **`auth_constant_time_equals()` is full-width `size_t`** — the `unsigned char` truncation bug from #1165 wasn't reintroduced.
- **`server_secret` is properly masked.** Both `GET /config` (`web_server_config.cpp:29`) and the MQTT config publish call `config_serialize(..., hideSecrets=true)`, and `ConfigOptSecret::set()` rejects the dummy value — so a config backup/restore can't install a known HMAC key. That's the trap I went looking for and it's closed.
- **Rebased on current master, merges cleanly.**

## Blocking

### 🔴 1. Session expiry is unenforceable — the ESP32 has no clock at boot

`session_token_verify()`'s only freshness check is `now < exp`, with `now = (uint32_t)time(nullptr)`. The ESP32 has no battery-backed RTC, so after every power cycle `time(nullptr)` returns ~0 until SNTP syncs. In that window, **every previously issued token — including long-expired ones — verifies**, because `0 < 1.7e9`. An attacker who captured a cookie months ago replays it after any reboot, and reboots are trivially induced (`GET /restart`).

Worse, on a LAN with no internet or SNTP blocked, the clock never syncs and expiry is **never enforced at all** — the "6 hour session" is permanent.

The mirror image is a UX bug from the same root: a token minted before sync gets `exp ≈ 21600` (1970), so logging in right after a reboot silently stops working the moment NTP lands.

Cheapest fix is to fail closed on an unknown clock (but maybe this locks us out and prevents setting up the clock):

```cpp
static bool clockIsSane() { return (uint32_t)time(nullptr) > 1700000000; }
// in hasValidSessionCookie(): if(!clockIsSane()) return false;
// in handleLogin():           if(!clockIsSane()) -> 503 "clock not set"
```

`timeManager` already knows whether time has been set, so wiring to that is cleaner than a magic constant.

### 🔴 2. The login lockout is bypassed entirely by HTTP Basic

`s_loginFails` / `s_lockUntil` only gate `POST /login`. Basic auth goes through `isAuthenticated()`, which has **no rate limiting whatsoever** — an attacker brute-forces `www_password` at full speed against `GET /status` and never touches `/login`. The lockout protects the door the attacker won't use.

Since #1165 the credential comparison is already centralised, so the counter belongs on the failed-Basic path in `isAuthenticated()`, not just in the login handler.

Related, and I'd fix it in the same pass: **the lockout state is global and re-armable**, so any LAN host can lock the admin out indefinitely by posting one bad login every 15 minutes. `s_loginFails` only resets on success, which the admin can't achieve while locked. At minimum let the counter decay, and consider not extending the window on attempts made *during* a lockout.

## Should fix before merge

**🟡 Username oracle from short-circuit.** 

```cpp
bool ok = auth_constant_time_equals(user, www_username) &&
          auth_constant_time_equals(pass, www_password);
```

`&&` skips the password compare on a username miss, reintroducing exactly the timing signal the constant-time helper exists to remove. Evaluate both into locals first, then `&&`.

**🟡 State-changing GETs are exempt from the CSRF check.** The guard is `request->method() != HTTP_GET`, but `/reset` (factory reset + reboot), `/restart`, `/apoff`, `/divertmode`, `/shaper`, and `/r?rapi=$FS` all actuate on GET. The entire defense for those is `SameSite=Strict`, which does hold on current browsers — but it's the only layer, on the endpoints with the worst consequences. Either convert them to POST or extend the header requirement to that route list explicitly.

**🟡 The 401 still carries `WWW-Authenticate: Basic`.** `requestAuthentication()` is unchanged, so a browser hitting a protected route gets the native Basic modal — which will preempt the SPA's 401→login interceptor. This needs resolving jointly with `gui-nightshift` before either lands; the backend probably needs to omit `WWW-Authenticate` when the request looks like a browser XHR.

**🟡 Password-without-username still leaves the device wide open.** `isAuthenticated()` early-returns `true` when `www_username == ""` (inherited from #1165). If the first-run wizard lets someone set only a password, they get a login page, a working session cookie, and zero actual protection. Given the whole point of this PR is turning F1's silent-insecure-default into an informed choice, this is the one to get right in the wizard — and arguably the firmware should treat "password set, username empty" as auth-enabled with a default username rather than as auth-disabled.

## Minor

- Secret rotation lives inside `#if ENABLE_CONFIG_CHANGE_NOTIFICATION`. Only the divert-sim env sets it to 0 and that env doesn't build the web server, so it's fine today — but a security control shouldn't be inside a feature-notification guard.
- `web_auth_get_secret()` can call `config_user_commit()` (a flash write) from inside a network callback if the secret is ever empty at request time. Same hazard class as the erase-in-callback bug in #1167. Generate at boot only and fail closed at request time.
- `hasValidSessionCookie()` runs twice per non-GET request (CSRF check, then `isAuthenticated`) — two HMACs plus the `std::string` copies inside `hmac_sha256()`. Compute once and pass the result.
- No `Cache-Control: no-store` on the `/login` response carrying `Set-Cookie`; `/login` and `/logout` skip `requestPreProcess` so they get none of the baseline headers.
- Login CSRF: `/login` has no CSRF guard, and a cross-site `text/plain` form POST is a simple request. Only exploitable by someone who already has credentials, so low — worth a comment noting it's deliberate.
- Tokens carry no identity or session ID, so individual sessions can't be revoked — only "rotate the secret and sign everyone out." Reasonable for a single-admin device; worth stating in the PR body so nobody assumes otherwise later.
- **Conflicts with #1167** on `platformio.ini` (both extend the `native_test` `build_src_filter`). Trivial, but whichever lands second needs the fixup.

RAR:
Thanks — genuinely thorough, and the differential HMAC run (155 vectors) + the `extern "C"` C-linkage check are appreciated. Agree with essentially all of it. I've pushed `0215b2c9` addressing the mechanical items; the four with real tradeoffs I'd like to align on before coding so we don't pick the wrong policy.

**Fixed in `0215b2c9`:**
- **Username oracle** — both `handleLogin` and `isAuthenticated` now evaluate both compares into locals before the `&&`; no short-circuit.
- **Secret lifecycle** — `web_auth_get_secret()` is read-only; `web_auth_ensure_secret()` does the generate+persist and runs at boot only. No flash write from a request path anymore. `handleLogin`/`hasValidSessionCookie` fail closed if the secret is somehow empty.
- **Rotation** moved out of the `#if ENABLE_CONFIG_CHANGE_NOTIFICATION` guard.
- **Cookie check** computed once per request (via a `usedCookie` out-param) and reused by the CSRF guard.
- **XHR-aware 401** — a request carrying `X-Requested-With: OpenEVSE` gets a plain 401 (no `WWW-Authenticate`), so the SPA's interceptor shows the login page instead of the browser's Basic modal; curl/machine clients still get the Basic challenge. This pairs with the `gui-nightshift` 401 interceptor (branch `feat/login-page`).
- **`Cache-Control: no-store`** on `/login` and `/logout`.
- Will add the PR-body notes (no per-session revoke; rotate-to-sign-out-all) and rebase the `platformio.ini` `build_src_filter` overlap with #1167 for whichever lands second.

**Four I want to settle with you first — each has a tradeoff:**

1. **Clock/expiry.** You're right that `now < exp` is meaningless before SNTP. But failing closed on an unset clock locks out air-gapped LAN units permanently. Proposing: gate cookie mint+verify on a synced clock (via `timeManager`, not a magic constant); when the clock is unset, cookies simply don't apply and users authenticate with Basic until NTP lands. Blocks the replay-after-reboot path without a hard lockout — at the cost of no cookie sessions on a truly offline device. Acceptable, or do you want an offline escape hatch?

2. **Lockout.** Agree it's bypassed by Basic and the global counter is a re-armable DoS. The wrinkle: extending the lockout to the Basic path can lock out Home Assistant/MQTT (correct creds) during an attacker-induced window. Thinking a decaying counter that rate-limits *failed* attempts without ever blocking a correct credential. Want your read on the machine-client impact before I commit to a shape.

3. **State-changing GETs.** Agree `/reset`, `/restart`, `/apoff`, `/divertmode`, `/shaper`, `/r?rapi=$FS` lean entirely on `SameSite=Strict`. Blanket "require the header on GET" would break the `/r` console for direct browser use. I'd gate the destructive actuators (`/reset`, `/restart`, `/apoff`) via header-or-POST and leave `/r` to SameSite — or convert those three to POST-only if you'd rather (cleaner, breaks API compat). Which?

4. **Password-without-username.** Agree — `www_username==""` ⇒ open (from #1165) means "password set, username blank" is a login page over zero protection. Proposing firmware treats password-set/username-empty as auth-enabled with a default username (`admin`), plus the wizard requires a username. Sound right?

Happy to take any of these in whatever direction you prefer.

chris1howell:
1. Clock/expiry - I like your approch. Should there be a 1 or 2 minute maxamum wait for the clock to settle? So racing a reboot on a truly online device aould not be allowed.
2. Lockout. Sounds good.
3. State-changing GETs. no preferance whichever is easier to implement.
4. Password-without-username. Thats fine, if we attribute to a default name, lets use something like openevseadmin, so its not a common target like root or admin. If it is just blank not applicable.

RAR:
All four settled and pushed (`44a2a00a` + review follow-up `c83eb350`). Device builds 94.4%, native `test_web_auth` 15/28 green (5 new throttle cases). I ran an adversarial security pass over the diff before pushing; the one real issue it surfaced (below) is fixed in `c83eb350`.

**1. Clock/expiry** — `clockIsSane()` gates both mint and verify. `hasValidSessionCookie()` returns false and `handleLogin` returns `503 {"msg":"clock"}` until `time()` clears the epoch floor (lifted by SNTP, the EVSE RTC, *or* a manual set — deliberately source-agnostic, since a `timeManager` "SNTP-synced" flag would miss a clock set from the controller).

On your max-wait question: I don't think a 1–2 min timer is needed, because the cookie is **inert** while the clock is unset rather than *accepted-without-expiry-check*. That closes the reboot race by construction: reboot → `time()`≈0 → the replayed old cookie can't verify (Basic only), and once the clock lands its `exp` is already in the past → still rejected. There's no "open window" to time-box. A truly offline unit just stays on Basic forever, which is safe. Happy to add the `timeManager`-backed cap as belt-and-suspenders if you'd still prefer it, but I read it as redundant here.

**2. Lockout** — replaced the per-`/login` counter with a shared **decaying** throttle (`web_auth`, native-tested) fed by *both* failed Basic (in `isAuthenticated`) and failed `/login`. A correct credential always authenticates and resets the counter, so HA/MQTT holding the right password can't be locked out. Failed attempts get `429` while hot (`LOCK_AT=10`); the counter decays ~1/min, killing the re-armable global DoS. Timing is `millis()`-based (monotonic), immune to the SNTP jump.

**3. State-changing GETs** — went with header-or-POST (easiest, no API-compat break): `/reset`, `/restart`, `/apoff` now require a non-GET method or `X-Requested-With: OpenEVSE` (`actuatorMethodAllowed`). The GUI already sends the header on every request, so it's unaffected; a bare cross-site `GET` can't actuate them.

**4. Password-without-username** — auth is now keyed on the **password**. Password-set + username-blank ⇒ auth enabled with default user **`openevseadmin`** (per your steer). Both blank ⇒ open (n/a). I also routed the **static-shell handler** through the shared `isAuthenticated()` so it honours the same Basic-or-cookie, password-keyed policy — otherwise a password-set/username-blank device would still have served the shell open, and it fixes a cookie-reload → Basic-reprompt gap.

**Review follow-up (`c83eb350`):** unifying the static handler onto `isAuthenticated()` initially let a browser replaying a *stale* Basic header (e.g. right after a password change) record one throttle failure per asset — ~10 assets in a reload could self-trip the soft-lock. Added a `recordThrottle` flag (default true); the static handler passes false, so only the API and `/login` (which actually adjudicate credentials) feed the counter. The shell still enforces the policy, it just doesn't count.

**Two notes for the changelog rather than code changes:**
- The actuator guard is a behavior change for anyone who scripted a bare `GET /reset|/restart|/apoff` — they'll need the header or a POST now.
- Tokens still carry no per-session identity, so revocation remains "rotate the secret → sign everyone out" (fine for the single-admin model); and the throttle counter is process-global (one hot attacker makes *wrong*-credential requests from others get `429`, but correct creds always pass). I'll fold both into the PR body.

Still to do on my side: the PR-body notes above, the `platformio.ini` `build_src_filter` fixup vs #1167 for whichever lands second, and HW validation of the live `POST /login → 200 + Set-Cookie` flow (native runs AP-only so the gate is open there). The `gui-nightshift` side (`feat/login-page`: X-Requested-With + 401→login interceptor, standalone login page, logout, wizard) is ready and pairs with the XHR-aware 401 — I'll open that PR once the endpoints are confirmed on hardware.

RAR:
GUI side is up as a companion: OpenEVSE/openevse-gui-nightshift#64 (login page, `/logout`, the `X-Requested-With` + 401→login interceptor that pairs with the XHR-aware 401 here, and the first-run password nudge). It depends on this PR — the `/login`+`/logout` endpoints and the suppressed `WWW-Authenticate` don't exist until #1168 merges — so it's opened for review but flagged not-to-merge-first. Live end-to-end validation is done jointly once this is on hardware.

chris1howell (dismissed):

RAR:
Thanks for putting it on hardware — that surfaced the exact integration gap you flagged in review (WWW-Authenticate vs the SPA login). Root cause for each thing you hit:

**1. Basic modal instead of the login page — this was the bug.** A top-level navigation to `/` goes through the static file handler, which was calling `requestAuthentication()` → `WWW-Authenticate: Basic`, so the browser popped its native dialog *before* the SPA could load and render `Login.svelte`. The XHR-aware 401 only helps once the app is already running; the shell load happens first, so it never got there.

Fixed in `de68b57e`: the static shell (HTML/JS/CSS/assets) is now served **unauthenticated**. It carries no secrets — every piece of data and every action stays behind the authenticated API/WebSocket. That's the standard SPA model and the only way the login page is reachable on the first navigation. **Heads-up — posture change:** on Basic-only setups the UI *shell* used to be Basic-gated and is now open (data still protected). That trade is inherent to having a login page; say the word if you'd rather gate it some other way.

**2. "Charger failed to accept settings" / couldn't factory-reset** — same root cause, compounded by the mid-session enable. Setting a password immediately turned auth on, but that browser session had no cookie and no Basic creds, so its next writes (config save, `/reset`) started 401ing — which looked like the charger rejecting everything. With #1 fixed, that 401 now routes the session to `/login` → cookie → you can continue or reset, instead of hitting the Basic wall.

**3. The partial password saving before you were done — this is the real trap, and it's GUI-side.** The HTTP username/password fields live-save independently on blur (the general config-form pattern), so `username` + the first couple password chars persisted as you typed, enabling auth with a truncated password → locked out. I don't think credentials should live-save per field. My proposal: in both the wizard security step and the Settings auth section, hold the pair locally and write `www_username`+`www_password` **together on one explicit "Set password" action**, require a username when a password is set (the D4 GUI half), and add a confirm-password field so a typo can't lock you out. That changes the live-save pattern for that one section — do you want me to go that way, or keep live-save but just write the pair atomically and block password-without-username? Your call on the UX.

**4. Phone wouldn't reconnect after the wizard reboot** — I read this as the normal AP→station handoff rather than an auth issue: once the device joins your WiFi it drops its own AP, so a phone still on the device AP (192.168.4.1) loses it and has to rejoin the home network — which is why the PC, already on that network, could finish. If it looked different from that (device never came back at all, etc.), tell me and I'll dig in.

`de68b57e` builds 94.4%, native suite green. GUI companion is OpenEVSE/openevse-gui-nightshift#64. None of this touches the four decisions we settled — it's the shell-serving gap underneath them.

RAR:
Put it on hardware (16MB WROOM at a real STA address) and reproduced the Basic modal after setting a password — `de68b57e` (serve the shell open) was necessary but **not sufficient**. There was a second bug underneath it, now fixed and HW-verified in `f9bd6326`.

**Root cause:** `MongooseString::c_str()` returns the raw `mg_str` pointer, which is **not null-terminated** — it points into the header buffer where the value is followed by `\r\nHost: …`. So `strcmp(xrw.c_str(), "OpenEVSE")` was comparing `"OpenEVSE\r\nHost:…"` against `"OpenEVSE"` and **never matched**. Every `X-Requested-With` check fell through:
- the XHR-aware 401 sent `WWW-Authenticate: Basic` anyway → the browser modal you (and I) saw;
- the cookie CSRF guard would have 403'd every SPA mutation;
- the actuator guard would have 403'd `GET /reset|/restart|/apoff` even with the header.

It was mine, in since the first XHR-401 commit (`0215b2c9`). Fix is the library's length-aware `MongooseString::equals()` (`mg_strcmp`) — the same idiom the ETag check already uses.

**Verified on the device with a password set:**
```
GET  /status  -H 'X-Requested-With: OpenEVSE'  → 401 {"msg":"auth"}   (no WWW-Authenticate)
GET  /status  (no header)                      → 401 + WWW-Authenticate: Basic
POST /login   (bad creds)                      → 401 {"msg":"invalid"}, Cache-Control: no-store
GET  /                                         → 200 (shell served open)
```
So the SPA now gets the plain 401 → its interceptor routes to the login page instead of the browser popping Basic; curl (machine clients) still get the Basic challenge. Pushed to `feat/web-login-session`. The last-mile visual (login page renders, correct password lands a cookie + dashboard) is being eyeballed in a browser now, but the HTTP contract is confirmed on-device.

Separately, that `c_str()`-isn't-null-terminated footgun is worth being wary of elsewhere in the tree — anywhere a header value is `strcmp`'d via `c_str()` has the same latent bug.

chris1howell (approved):

PR #1171, opened 2026-07-28
chore(ai): sandbox policy for local AI agent development

jeremypoulter:
## Context

This repo already has good AI-agent *guidance* (`AGENTS.md`, `docs/ai/invariants.md`, Copilot instructions) but nothing that makes an agent productive on it without constant permission prompts, and no isolation to make that safe.

The PR has two separable halves, and they can be judged independently:

1. **`scripts/openevse_test.sh`** — a test/diagnostic harness that starts, uses and tears down firmware + emulator instances inside a single invocation. Useful whether or not anyone runs a sandbox.
2. **Agent policy** — a small checked-in set of guards, plus an **opt-in** sandbox config that lives in each developer's gitignored `.claude/settings.local.json`.

## 1. The harness

Every sandboxed Bash call gets its own network namespace, so a server started in one command is gone by the next. Each subcommand therefore launches what it needs, runs your command, and tears everything down in one go — pass the command after `--`, read instance URLs from `$EVSE_URL` / `$EVSE_URL_<i>`.

```bash
scripts/openevse_test.sh all                  # unit + divert + gui
scripts/openevse_test.sh native -n 2          # 2 firmware instances, sandboxed
scripts/openevse_test.sh native -- curl -s "$EVSE_URL/status"
scripts/openevse_test.sh emulator -n 2        # 2 firmware + emulator pairs (needs docker)
scripts/openevse_test.sh integration          # the checked-in pytest suite
scripts/openevse_test.sh launch -i 1          # one long-lived pair, held until Ctrl-C
```

`launch` is the interactive one: a single pair for a given instance ID, held in the foreground so a REST client (`test/*.http`) or a second terminal can drive it. Ports, chip ID, hostname and working directory all derive from `-i`, and each has an override.

Every instance gets its own hostname and chip ID so several are addressable at once — needed for the load-sharing work in #1027, where members join a group by mDNS name. Worth one note on how: ESPAL treats the chip ID as a MAC, so `getLongId()` reverses the bytes and shifts right 16, and `getShortId()` keeps the last four hex digits — which come from **bytes 2-3**. Varying the low byte leaves every instance sharing a short ID, and with it the default hostname, the softAP SSID and the MQTT announce topic. The harness varies the `0x5678` field instead.

## 2. Policy — what ships, and what does not

Following review feedback, the sandbox is **not** imposed by the repo.

**Checked in** (`.claude/settings.json`, 22 lines) — only the portable parts:

- **Read/Edit denials** for `*.pem` (OTA signing keys), `.env` files, and `.vscode/settings.json` (holds test WiFi credentials). These tools are *not* covered by the Bash sandbox, so they need explicit rules.
- **Prompts retained** on `git push` and the device OTA flash (`curl … /update`).

Neither needs a sandbox, neither assumes anything about your machine, and neither changes your security posture.

**Opt-in** (`docs/ai/sandbox.md` → paste into your gitignored `.claude/settings.local.json`) — the egress allowlist, filesystem/credential guards, `excludedCommands`, and a starting-point `allow` list.

Turning the sandbox on decides how much autonomy an agent has on *your* machine, which is not the repo's call to make. It also is not free to adopt: it restricts egress to a fixed domain list, so anyone behind a corporate proxy or a PyPI/npm mirror that is not on it gets build failures that do not obviously point back at the config, and Ubuntu 24.04+ needs the AppArmor `bwrap` workaround on top. And it is Claude-Code-only, while `AGENTS.md` and `docs/ai/invariants.md` serve Copilot/Codex/Cursor users equally.

`.gitignore` gains `.claude/settings.local.json`.

## Known limits, stated rather than solved

- **The egress allowlist is not a containment boundary.** `github.com` and `api.anthropic.com` have to be reachable for the agent to work at all, so a determined exfiltration path stays open — and the proxy allows by hostname without inspecting TLS. Treat it as a guard against accidental reach and dependency confusion.
- **The settings schema moves quickly**, and renamed or removed keys degrade silently rather than erroring. A stale block can look protective while doing nothing.
- **The deny globs are bounded-depth on purpose** — `./*/`, `./*/*/` rather than a recursive `./**/`. Patterns resolve against the real tree and that resolution follows symlinks; `lib/` symlinks point at sibling checkouts whose `examples/*/lib/` directories symlink back to their own root, forming cycles. A recursive `**` walk never terminates on that shape — observed ~20 GB RSS before the OOM killer in the V3.x tree. Every real secret lives at depth ≤ 3, so bounding costs no coverage. **Please do not "fix" these back to `**`.**
- **For unattended / `--dangerously-skip-permissions` runs this host-level sandbox is not enough** — that needs a dev container with a default-deny egress firewall, or a VM. Deliberate follow-up, not set up here.

## Verification

- Harness exercised end to end: `native -n 2`, `emulator -n 2` (Docker), `launch` with and without an emulator, console mirroring, and Ctrl-C teardown leaving no containers, processes or run directories behind.
- Distinct identity per instance confirmed against a live pair — `mqtt_announce_topic` `openevse/announce/7856` vs `/7956`, firmware banner `OpenEVSE WiFI 7956`.
- Argument validation, `--help`, and JSON validity of both `.claude/settings.json` and the paste-in block.
- The test WiFi credentials in `.vscode/settings.json` were checked and are **not** committed (`.vscode` is gitignored), so this is not a historical leak. The deny rule is defence-in-depth.

Not yet run: the full GUI/divert-sim validation gate under the opt-in policy on a warm toolchain cache. Checklist is in `docs/ai/sandbox.md`.


jeremypoulter:
@RAR / @chris1howell please take a look, I think this will allow Cluade to just get on with stuff a bit more

RAR:
Had a proper read through this. The harness half I'd take today; the checked-in `.claude/settings.json` I'd rather we didn't ship as-is.

**Worth merging on its own:** `scripts/openevse_test.sh`, the AGENTS.md test section, and the `.gitignore` line. The one-invocation start/use/teardown shape is useful whether or not anyone runs a sandbox, and `native -- <cmd>` is a nicer entry point than the current pile of separate commands. None of it depends on the policy file.

**Concerns with `.claude/settings.json` as a shared, checked-in file:**

1. It contains `"Read(//home/jpoulter/.claude/**)"` — a hardcoded personal home path, which suggests this began life as a local settings file.

2. A repo-level settings.json applies to every Claude Code user who opens the repo, and it turns `sandbox.enabled` + `autoAllowBashIfSandboxed` on for them. That's a security-posture decision taken on the contributor's behalf. The failure mode is also unfriendly: anyone behind a corporate proxy or a PyPI/npm mirror that isn't among the 13 allowed domains gets builds failing in ways that don't obviously point back at this file, and Ubuntu 24.04 users additionally need the AppArmor `bwrap` workaround the doc describes.

3. It's Claude-Code-only. Contributors on Copilot, Codex, Cursor et al. get nothing from it, while `AGENTS.md` + `docs/ai/invariants.md` already serve all of them.

4. The allow list is brittle and machine-shaped — `divert_sim/.venv/bin/pytest`, eight near-duplicate `curl localhost:…` entries. Claude Code's settings schema also moves quickly, and renamed or removed keys degrade silently rather than erroring, so a stale policy can look protective without doing anything.

5. The egress allowlist's security value is modest by the PR's own reasoning: with `github.com` and `api.anthropic.com` allowed, the exfiltration path the list is meant to close is still open.

**What I'd suggest instead** — keep the parts that are portable and cost nothing:

- Keep the `deny` block (`*.pem`, `.env*`, `.vscode/settings.json`) and the `ask` block (`git push`, `curl *update*`) in the checked-in file. Those are good defence-in-depth for anyone with the repo, need no sandbox, and carry no machine-specific assumptions.
- Move the `sandbox` block (network allowlist, filesystem, credentials, excludedCommands) and the personal `allow` list into `docs/ai/sandbox.md` as a "paste this into your `.claude/settings.local.json`" starting point. Same knowledge captured, no policy imposed on people whose environment differs.
- Drop the `/home/jpoulter` entry either way.

That keeps the useful parts and the documented rationale, without the repo dictating each contributor's agent sandbox.

jeremypoulter:
Thanks, that's a fair read and I've taken most of it.

**Split as you suggested.** `.claude/settings.json` is now 22 lines: the `deny` block (`*.pem`, `.env*`, `.vscode/settings.json`) and the `ask` block (`git push`, OTA flash). Nothing else. The whole `sandbox` block and the `allow` list moved into `docs/ai/sandbox.md` as a "paste this into your `.claude/settings.local.json`" starting point. The personal `/home/jpoulter` entry is gone.

**On (2)** — agreed, and it was the strongest point. Nobody now inherits `sandbox.enabled` or `autoAllowBashIfSandboxed` by opening the repo, so the proxy/mirror and Ubuntu-AppArmor failure modes can't bite someone who never opted in.

**On (4)** — the allow list doesn't ship at all now. I also trimmed it 42 → 27 in the paste-in: the eight near-duplicate `curl localhost:…` globs collapse to two, and the entries that only made sense on my machine are gone (`divert_sim/.venv/bin/pytest`, both `pkill` rules, `python3 -c ' *`, the lvgl binary). `docs_coverage.py` and `sync_screenshots.py` stayed — they're CI-enforced workflows documented in AGENTS.md, not local habits. Your schema-drift sub-point I can't fix, so I've documented it instead: renamed keys degrade silently, and a stale block can look protective while doing nothing.

**On (5)** — you're right and I've stopped implying otherwise. The doc now says plainly that `github.com` and `api.anthropic.com` have to be open for the agent to work at all, so the allowlist guards against accidental reach and dependency confusion, not exfiltration.

**On (3)** — partially. What's left checked in is still a Claude-Code-only file that does nothing for Copilot/Codex/Cursor users. That's the shape your own suggestion asked for, so I've kept it, but I don't want to claim the concern is gone. If you'd rather even the deny/ask block were documentation instead of config, say so and I'll move it.

One thing worth flagging before you re-read: **the deny globs are bounded-depth on purpose** — `./*/`, `./*/*/` rather than `./**/`. Patterns resolve against the real tree and follow symlinks, and our `lib/` symlinks point at sibling checkouts whose `examples/*/lib/` symlink back to their own root. A recursive `**` walk never terminates on that — ~20 GB RSS before the OOM killer in the V3.x tree. Every real secret is at depth ≤ 3 so it costs no coverage, but it reads like a mistake in the JSON and JSON can't carry the explanation. Rationale is in `docs/ai/sandbox.md`.

The harness half is unchanged in shape, and has since grown `launch` for driving an instance from a REST client. PR description is rewritten to match.

RAR (commented):
Re-read at `70c436c5`. The split is what I was asking for: 22 lines checked in, no `sandbox` block and no `allow` list imposed on anyone, `/home/jpoulter` gone. I'd take this. Four small things from the re-read, none of them structural.

**1. `status_dump` prints `(request failed)` after a request that succeeded.**

```bash
curl -fsS --max-time 5 "${inst_url[$i]}$path" 2>/dev/null | head -c 400 || printf '(request failed)'
```

With `set -o pipefail`, once the body is larger than what curl manages to write before `head -c 400` exits, curl dies on `EPIPE` and the pipeline reports failure — so the error string gets appended right after 400 perfectly good bytes. Reproduced deterministically on a ~200 KB body; `/status` happens to squeak through today because it fits in one write, which makes this look intermittent rather than wrong. Redirect to a temp file and `head` that, or use `--max-filesize` and drop the pipe.

**2. `--help` breaks on a relative invocation from a non-root cwd.**

`usage()` reads `"${BASH_SOURCE[0]}"`, but the script has already `cd`'d to `$repo_root`, so a relative `BASH_SOURCE` no longer resolves:

```
awk: fatal: cannot open file `./openevse_test.sh' for reading: No such file or directory
```

Resolving the path once, before the `cd`, fixes it.

**3. The doc claims a little more than the config delivers.**

`docs/ai/sandbox.md` says "the built-in Read/Edit tools are *not* covered by the Bash sandbox, so `*.pem` (OTA signing keys), `.env` files, and `.vscode/settings.json` … are denied directly", and the PR description says "Read/Edit denials". The shipped file has only `Read(…)` rules. In practice `Edit` is gated anyway because it demands a prior `Read`, but `Write` isn't — a `.pem` can still be overwritten. Either mirror the globs as `Edit(…)`/`Write(…)`, or soften the wording so the file and the doc agree.

**4. `launch` drops emulated-flash state into a tracked directory.**

Default workdir is `$repo_root/test/$id`, and `test/` is where the checked-in `.http` clients live. The native binary writes `epoxyfsdata` / `epoxyeepromdata` into its cwd, and `.gitignore` only covers `divert_sim/epoxy*` — so every `launch` leaves untracked files inside a tracked directory. You're already touching `.gitignore` here, so one more line closes it.

Two things I checked rather than assumed, both fine: the bounded-depth deny globs are correct and the comment explaining them is worth keeping — the `lib/` → sibling-checkout → `examples/*/lib/` cycle really doesn't terminate under `**`. And the harness's port bases do match `tests/integration/conftest.py` (8080 emulator web, 8023 RAPI), as the comment claims.

jeremypoulter:
Good catches — three of the four were real and are fixed in `47659e51`. The fourth I can't reproduce; details below.

**1. `status_dump` printing `(request failed)` after a successful request** — confirmed and fixed. Reproduced exactly as you described: on a ~200 KB body it printed 40 good bytes then the error string; a small body was clean. It now fetches to a file and `head`s that, so curl's own exit status decides the message rather than the pipeline's. Your read on why it looks intermittent rather than wrong was spot on.

**2. `--help` on a relative invocation** — confirmed and fixed. `cd scripts && ./openevse_test.sh --help` gave exactly the awk error you quoted. The path is now resolved once, before the `cd`. Checked from `scripts/`, the repo root, an absolute path, and an unrelated cwd.

**3. Doc claiming more than the config delivers** — you're right, and I fixed the config rather than softening the wording, since your point about `Write` being the actual gap is the important one. All 11 globs are now mirrored across `Read`, `Edit` and `Write`, and the doc says explicitly that `Edit` is already gated by its own requirement to `Read` first, so `Write` is the verb doing the real work.

Given your earlier point about the schema degrading silently, I checked the verbs against the published schema rather than assuming. Worth flagging for anyone who repeats the exercise: `Write(` does **not** appear in the schema's `examples` list, which looks alarming at first, but the `permissionRule` `pattern` enumerates the valid tool names and `Write` is in it. All 35 rules validate against that pattern.

**4. `launch` dropping emulated-flash state into a tracked directory** — this one doesn't reproduce, and I think the premise is off by a couple of lines. `.gitignore` doesn't only cover `divert_sim/epoxy*`; lines 111-112 carry bare `epoxyfsdata` and `epoxyeepromdata` patterns, and a bare pattern matches at any depth:

```
$ git check-ignore -v test/1/epoxyfsdata/emeter.json
.gitignore:111:epoxyfsdata	test/1/epoxyfsdata/emeter.json

$ git status --short --untracked-files=all test/
(empty)
```

So nothing untracked is left behind and the extra `.gitignore` line would be a no-op.

The observation underneath it still stands, though: `launch` does accumulate `test/0/`, `test/1/` … inside a directory of checked-in `.http` clients. That's clutter rather than a correctness problem, so I've left the documented default alone for now — but if you'd rather it lived somewhere obviously ephemeral (`.harness/<i>`, say) I'm happy to move it, it's a one-line change plus a doc update.

Thanks also for checking the deny globs and the port bases rather than assuming — the `lib/` cycle is exactly the trap it looks like, and it's useful to have someone else confirm the comment earns its place.

PR #1176, opened 2026-08-02
fix: handle the D9 controller fault states across the ESP32 stack

RAR:
## Problem

`OpenEVSE_Lib` 0.0.19 added four controller fault states:

| Value | State |
|---|---|
| `0x0E` | `RELAY_CLOSURE_FAULT` |
| `0x0F` | `PP_SHORTED` |
| `0x10` | `PP_MISSING` |
| `0x11` | `EEPROM_FAILURE` |

`getStateColour()` was updated when the lib was bumped, but every other consumer was missed.

The significant one is `EvseMonitor::isError()`:

```cpp
return OPENEVSE_STATE_VENT_REQUIRED <= _evse_state && _evse_state <= OPENEVSE_STATE_OVER_CURRENT;
```

The controller's fault states are **not contiguous** — `0x0C` and `0x0D` are reserved — so all four new states fall outside the range and `isError()` returns `false` for every one of them. That propagates:

- `EvseMonitor::updateEvseState()` — error path skipped
- `evse_man.cpp:356` — fault logged as `Information` instead of `Warning`
- `LedManagerTask.cpp:315,654` — neither the fault colour nor the fault blink rate

The user-visible result of a PP fault is a charger that silently refuses to charge, showing a dimmed `--` on the display, with nothing anywhere reporting a problem.

## Change

`isError()` becomes an explicit list rather than a range, so a future state addition surfaces as a missing case in review instead of silently reading as "not an error". Then the six consumers that switch on EVSE state:

| File | What |
|---|---|
| `lcd.cpp` | debug state names + 16x2 status lines |
| `lcd_lvgl.cpp` | `stateKeepsAwake()` — a fault holds the backlight on |
| `lvgl_tft/screen_common.cpp` | `state_word()` |
| `screens/screen_charge.cpp` | error icon |
| `screens/screen_manager.cpp` | backlight timeout suppression |
| `ocpp.cpp` | error code mapping |

The PP faults get their own labels (`PP MISSING` / `PP SHORTED` on the TFT, `CABLE ERROR` on the 16x2) rather than the generic `FAULT`, so the display names the actual wiring condition.

For OCPP: `RELAY_CLOSURE_FAULT` joins `PowerSwitchFailure`, `EEPROM_FAILURE` maps to `InternalError`, and the PP faults report `OtherError` with free-text detail — OCPP 1.6 has no dedicated code for proximity pilot. All three strings are in the 1.6 `ChargePointErrorCode` enumeration.

No behaviour change for any pre-existing state.

## Testing

- `openevse_wifi_tft_v1` builds clean; all seven modified files compile in that env
- `isError()` logic checked against the real `OPENEVSE_STATE_*` values from the pinned `OpenEVSE_Lib#0.0.19` — all 20 cases, including that reserved `0x0C`/`0x0D` remain non-errors
- Flashed to a TFT unit and confirmed healthy through a normal charge session, which exercises the edited switches without regression

Not yet exercised on hardware: the PP fault path itself. Reproducing it needs a controller in `PP_MISSING`.

## How it was found

A SAMD NXT running `9.0.0.SAMD` with PP auto-ampacity enabled and no PP wire connected: the controller entered `PP_MISSING` and the TFT showed `--`.


chris1howell (approved):

jeremypoulter:
Kindof sounds like this is a bug in the safety firmware, but guess there is no option but to fix it at a higher level. But I do wonder if the isError should be moved to the OpenEVSE library?

PR #1178, opened 2026-08-03
feat(rfid): add RC522 SPI reader support

tooraj2011:
## RC522 RFID Reader Support

This PR adds MFRC522/RC522 SPI RFID reader support as an alternative to the existing PN532 NFC reader.

### Changes
- Added `ENABLE_RC522` RFID backend
- Added `openevse_wifi_v1_rc522` PlatformIO environment
- Added RC522 SPI driver implementation
- Added RC522 build and user documentation
- Added MFRC522 clone version `0x88` support
- Improved reader timeout handling and false failure detection
- Fixed RC522 reset pin conflict with MCP9808 I2C (GPIO22 → GPIO4)

### Validation
- Built successfully:
- python -m platformio run -e openevse_wifi_v1_rc522

Build result:

SUCCESS

### Hardware
Default RC522 wiring:
- SS: GPIO5
- RST: GPIO4
- SCK: GPIO18
- MOSI: GPIO23
- MISO: GPIO19

PN532 support remains unchanged.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -63,6 +63,10 @@
 #include "rc522.h"
 #endif
 
+#if defined(ENABLE_PN532) && defined(ENABLE_RC522)
+#error "ENABLE_PN532 and ENABLE_RC522 are mutually exclusive — use separate PlatformIO envs"
+#endif
+
 #include "LedManagerTask.h"
 #include "event_log.h"
 #include "evse_man.h"
@@ -200,8 +204,8 @@ void setup()
   pn532.begin();
   rfid.begin(evse, pn532);
 #elif defined(ENABLE_RC522)
-  rc522.begin();
   rfid.begin(evse, rc522);
+  rc522.begin();
 #else
   rfid.begin(evse, rfidNullDevice);
 #endif
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -59,6 +59,10 @@
 #include "pn532.h"
 #endif
 
+#if defined(ENABLE_RC522)
+#include "rc522.h"
+#endif
+
 #include "LedManagerTask.h"
 #include "event_log.h"
 #include "evse_man.h"
@@ -195,6 +199,9 @@ void setup()
 #if defined(ENABLE_PN532)
   pn532.begin();
   rfid.begin(evse, pn532);
+#elif defined(ENABLE_RC522)
+  rc522.begin();
+  rfid.begin(evse, rc522);
 #else
   rfid.begin(evse, rfidNullDevice);
 #endif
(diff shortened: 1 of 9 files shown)


tooraj2011:
Friendly ping. When someone has time, I'd appreciate a review of this PR. Thank you!

src/main.cpp
```diff
@@ -195,6 +203,9 @@ void setup()
 #if defined(ENABLE_PN532)
   pn532.begin();
   rfid.begin(evse, pn532);
+#elif defined(ENABLE_RC522)
+  rfid.begin(evse, rc522);
+  rc522.begin();
```
jeremypoulter: I think the driver should probably be initialised before passing into the RFID module
```suggestion
  rc522.begin();
  rfid.begin(evse, rc522);
```

src/rc522.cpp
```diff
@@ -0,0 +1,211 @@
...
+    // one-shot can false-negative (reader still powering up) and would otherwise
+    // report absent until reboot. readerPresent() returns this cached flag; probeReader()
+    // actively refreshes it over SPI.
+#if defined(RC522_SPI_SCK) && defined(RC522_SPI_MISO) && defined(RC522_SPI_MOSI)
+    _spi->begin(RC522_SPI_SCK, RC522_SPI_MISO, RC522_SPI_MOSI, _ss_pin);
+#else
+    _spi->begin();
+#endif
+
+    _mfrc522.PCD_Init(_ss_pin, _rst_pin);
```
jeremypoulter, line 67: This seems to duplicate the code in initialise, should this just call initialise then do the check?

src/rc522.cpp
```diff
@@ -0,0 +1,211 @@
...
+    // even when RFID is disabled in config.
+    probeReader();
+    MicroTask.startTask(this);
+}
+
+bool RC522Reader::probeReader() {
+    if (_initialized && !_failure) {
+        // Already communicating — avoid resetting the chip mid-session.
+        return true;
+    }
```
jeremypoulter: Is this correct? Does initialised with no failures equate to a reader being present, if initialised and there was a failure should we try again?

Maybe should be:

```suggestion
    if (_initialized) {
        // Already communicating — avoid resetting the chip mid-session.
        return _present;
    }
```

src/rc522.cpp
```diff
@@ -0,0 +1,211 @@
...
+        lcd.display("RFID chip not found", 0, 1, 5 * 1000, LCD_CLEAR_LINE);
+        _failure = true;
+        _initialized = false;
+        _present = false;
+    }
+
+    if (!_initialized || _failure) {
+        initialize();
+        return POLL_DELAY;
+    }
```
jeremypoulter, line 203: Seems like this should be at the top of the loop or probably in `begin` or `setup`

PR #1179, opened 2026-08-03
heap fragmentation instrumentation + reboot fixes

RAR:
My EVSE was rebooting repeatedly — ESP32 only, the attached controller stayed up across every event (controller-sourced `session_elapsed` was 2099s while ESP32 `uptime` was 44s), so not power.

`/status` reported `free_heap` flat at ~60.7k right up to two observed reboots, which made it look like memory was fine. It wasn't: `free_heap` cannot distinguish "plenty of memory" from "plenty of memory in fragments too small to allocate from". Adding the metric that can tell those apart is what found everything else here.

## Measured on hardware (openevse_wifi_tft_v1, 16MB WROOM, no PSRAM)

| | before | after |
|---|---:|---:|
| `free_heap` | 60,276 | 91,024 |
| `heap_largest` | 19,444 | 61,428 |
| `heap_largest_min` | 11,252 | 59,380 |
| **`heap_min`** | **1,620** | **56,248** |
| static `.bss` | 106,372 | 93,836 |

`heap_min` is the all-time low free heap. This unit was surviving on **1,620 bytes** of worst-case margin. During an OTA upload with the Home Assistant integration polling, it was measured at **412 bytes** — which is very likely the long-standing "OTA dies at 2–7%" problem.

Note the largest block gained 27.6 KB from a 15 KB saving: the LVGL draw buffer wasn't just consuming memory, it was fragmenting the heap. Removing it let the regions either side coalesce.

## Diagnostics (`src/diagnostics.{h,cpp}`)

New `/status` fields: `heap_largest`, `heap_largest_min`, `heap_min`, `reset_reason`, `reset_reason_name`, `stack_loop_min`, `stack_events_min`, `ws_conns`, `ws_send_max`, `ws_reaped`, `lv_used_max`, `lv_frag_max`.

`reset_reason_name` is the important one — an unexplained reboot now names itself (`panic` / `task_wdt` / `brownout`) instead of just resetting the uptime counter.

Also caps the websocket send backlog. Nothing in the broadcast path applied backpressure, so a client that stopped reading accumulated every event indefinitely; at ~60k free that is survivable for well under two minutes. Clients past the cap are closed and counted.

## Crash fixes

- **Four unvalidated inputs that reboot the board.** `EvseState::fromString` and `LimitType::fromString` read `value[0]` with no null check, and callers pass a JSON value straight through — ArduinoJson yields `nullptr` for a missing or non-string key, so a POST to `/override`, `/claims` or `/limit` with a non-string value dereferenced null. Two `std::stoull` calls throw uncaught; one of them parses the stored `www_certificate_id` **on the boot path**, so a corrupted value boot-loops the unit with no way in over the network.
- **Loop watchdog armed at the top of `setup()`.** Arduino's `loopTask` feeds the task watchdog once per `loop()` iteration and never during `setup()`, so the whole of setup ran under a 5s panic timer with nothing resetting it — `LittleFS.begin(FORMAT_LITTLEFS_IF_FAILED)`, TSDB recovery and DNS all sit on that path. Self-sustaining: panic, reboot, same slow path. Moved to the end of `setup()`.
- **`onWsConnect` sent a truncated status to every client.** Its document was sized for 40 members while `buildStatus()` emits ~95, so it silently overflowed on every connect (ArduinoJson stops adding members rather than erroring). It also called `web_server_event()`, which broadcasts to *all* open websockets — one client reconnecting pushed a full status to every other one. Now correctly sized via a shared `STATUS_JSON_CAPACITY` and sent only to the connecting client.

## Memory

- LVGL draw buffer 32 → 16 lines (15 KB contiguous, `DRAW_BUF_LINES` override)
- `LV_MEM_SIZE` 32 KB → 24 KB — sized from measurement (`lv_used_max` peaks at 8.8 KB), not estimated
- `GET /certificates` built the whole store in one 32 KB document — larger than the biggest block this board could allocate, so the endpoint *failed* rather than merely costing a lot. Now streamed per certificate.
- StreamSpy rings 2 KB → 1 KB each; `SCHEDULER_MAX_EVENTS` 50 → 16 for this board only, via build flag (4,344 bytes `.bss`)

## Testing

All eight commits built, flashed and verified on hardware, one change per flash.

- Each malformed input fired at the live unit: `POST /override {"state":123}`, `POST /claims/1 {"state":true}`, `GET /certificates/zz`, `GET /certificates/ffffffffffffffffffffff`. All return normally; uptime monotonic, no reboot.
- Websocket verified with a raw client: both clients receive all 97 keys on connect, and an already-connected client receives nothing when a second connects (previously it would have been sent a full status).
- Clean boot confirmed over serial after each flash — no panics, no watchdog trips.

**Not verified:** the `LimitType::fromString` guard could not be reached — `/limit` rejects malformed bodies earlier at `limit.set()`, so it is defence-in-depth rather than a demonstrated fix. The boot-path `strtoull` fallback isn't tested directly (it needs a deliberately corrupted config); it uses the same pattern verified working at the `/certificates` site.

**Caveat on the numbers:** the "after" column is from lightly-loaded units. The minima will fall once several clients are polling again. The `free_heap` and `.bss` figures are stable; the minima need days of real traffic to mean much.

## Still open

`std::nothrow` on the ArduinoMongoose accept path — currently a failed allocation there calls `abort()`, so OOM becomes a panic reboot instead of a dropped request. That's in the pinned dependency, not this repo. The unsynchronised MicroTasks message list in `net_manager.cpp:347` and a double-free in `web_server_update.cpp` are both real and untouched; which one to prioritise depends on what `reset_reason` reports on the next unexplained reboot.


RAR:
## Update: the load evidence the draft was waiting on

The caveat in the description was that the "after" figures came from lightly-loaded units and the minima needed days of real traffic to mean anything. Two days of running has now happened.

**Live unit** (`openevse_wifi_tft_v1`, real Home Assistant + EVCC traffic, occasional browser): **26 hours uptime**, against roughly 6.5 minutes between reboots before the watchdog fix. `heap_largest` stepped down from 61,428 to 34,804 over the first several hours and then held flat for 12+ hours.

**Second unit under synthetic load** approximating the same client mix — two `/status` pollers at 5s and 13s, `/claims/target` and `/override` at 30s, `/config` at 120s, an index fetch and a forced websocket reconnect every 300s:

```
22h elapsed, 0 reboots, 0 missed polls
6,484 /status   829 /claims/target   829 /override   206 /config
83 index fetches   83 websocket connects   1,744 frames
0 errors
```

`heap_min` and `heap_largest_min` did not move from their boot values across the entire run.

### A theory this disproves

The description notes an unexplained residual: small permanent steps down in `heap_largest`, seven of about 1.2 KB in 17.6 hours, event-driven rather than time-driven. The obvious suspect was a per-websocket-reconnect leak, so the load generator was given a fixed 300s reconnect specifically to make that countable.

**83 forced reconnects produced no step at all**, and `ws_conns` stayed pinned throughout. Whatever the residual is, it is not websocket connection churn. It remains open, and it does not affect any claim in this PR — the fixes here are what took the largest block from 19,444 to 61,428 and the worst-case margin from 1,620 bytes to 56,248.

### Also worth recording

The 412-byte worst-case margin measured during an OTA is now better understood. There is a separate bug in `handleUpdateUpload()` that aborts an update with `UPDATE_ERROR_ERASE` after every byte has been written, whenever the declared upload length matches the image exactly — filed as #1181. It is independent of anything in this branch, but it is part of the same "OTA is unreliable on this board" picture that motivated the heap work.

Taking this out of draft.

RAR:
## Correction to the load figures above

The "26 hours uptime" in my previous comment was a stale reading. I took it before the unit rebooted and quoted it afterwards without re-checking. What actually happened:

```
17:41:38  uptime 95997  heap_largest 34804  heap_largest_min 10740  heap_min 164   wdt
17:41:48  uptime     6  heap_largest 61428  heap_largest_min 61428  heap_min 47252  panic
```

The live unit reached **26.7 hours and then panicked**, on 2026-08-04. It did not simply stay up. The comparison against the pre-fix behaviour still holds — roughly 6.5 minutes between reboots before the watchdog change, 26.7 hours after — but "26 hours uptime" read as a steady state, and that was wrong.

### What the current boot shows

17 hours in, on the same firmware and the same real traffic:

| | at boot | now (17h) |
|---|---:|---:|
| `heap_largest` | 61,428 | 42,996 |
| `heap_largest_min` | 61,428 | 17,396 |
| `heap_min` | 47,252 | **1,988** |

The largest block settles at 47,092 for about twelve hours, then steps to 42,996. The worst-case free heap has been down to **1,988 bytes**. The boot before this one recorded `heap_min` of **164 bytes** before it panicked.

### What this changes

The diagnostics in this PR are doing their job — a panic now names itself and the margin is measurable rather than guessed. The watchdog fix is a large, real improvement. But I should not have presented it as the unit being fixed:

- the residual heap decline is still unexplained, and it is not websocket reconnects (that was tested and ruled out)
- a `heap_min` of 164 bytes is not a safe margin, and something is still panicking at the end of a long run
- the `.bss` and `free_heap` gains in the description are stable and unaffected by any of this

I would treat this branch as instrumentation plus a set of genuine fixes that made the failure much rarer and now visible — not as a closed investigation.

chris1howell (approved):

PR #1182, opened 2026-08-05
fix(web): drop the built-in upload form, guard the remaining GET-actuated endpoints

RAR:
Two related changes that came out of auditing what the session-cookie CSRF guard actually covers.

### 1. Drop the built-in firmware upload form

`GET /update` served a bare multipart form:

```html
<form method='POST' action='/update' enctype='multipart/form-data'>
```

That POST cannot work for a browser authenticated by the session cookie. The CSRF guard in `requestPreProcess()` requires `X-Requested-With` on non-GET requests, and a plain HTML form cannot set a header. So the page has been broken since the login work landed, for exactly the users it existed to serve, and nobody has reported it.

It is also on course to break outright rather than partially: #1155 drops multipart upload support in favour of raw bodies, which would leave this page rendering fine and failing every time — worse than not offering it at all.

Uploading belongs to the web app; machine clients POST to the endpoint directly. `GET /update` now answers 405.

### 2. Guard the remaining GET-actuated endpoints

`actuatorMethodAllowed()` was applied to `/reset`, `/restart` and `/apoff`. Four other endpoints change state straight off a query parameter with no method check at all:

| endpoint | effect |
| --- | --- |
| `/divertmode` | `divert.setMode()` |
| `/shaper` | `shaper.setState()` |
| `/settime` | sets the clock |
| `/rfid/add` | puts the reader into enrolment mode |

The CSRF check never sees them, because it only fires on non-GET requests. `SameSite=Strict` means they are not reachable with a live cookie session today, so this is not a live exploit — it is the same defense in depth the other three already had, extended to the rest of the family.

`/rapi` and `/r` are deliberately left alone. They are the most powerful endpoint in the firmware, but the RAPI console page submits a GET form, so guarding them needs that page dealt with first — the same decision as the upload form above, and worth taking separately.

**Compatibility note:** a bare `GET /divertmode` or `GET /shaper` now answers 403. Anything driving the device from a script should use POST, or send `X-Requested-With: OpenEVSE`.

### Web app is unaffected

It POSTs `/divertmode`, does not call `/shaper` or `/settime`, and its `GET /rfid/add` goes through `httpAPI()`, which sets `X-Requested-With` on every request.

### Verification

Flashed to a 16MB ESP32-WROOM and exercised directly:

```
GET  /update                            -> 405 POST firmware to this endpoint
GET  /divertmode?divertmode=1 (bare)    -> 403 Forbidden: use POST or the app
GET  /divertmode?divertmode=1 (+XRW)    -> 200 Divert Mode changed
POST /divertmode divertmode=1           -> 200 Divert Mode changed
GET  /shaper?shaper=0   (bare)          -> 403 Forbidden: use POST or the app
GET  /settime           (bare)          -> 403 Forbidden: use POST or the app
GET  /rfid/add          (bare)          -> 403 Forbidden: use POST or the app
GET  /status                            -> 200 {"mode":"STA",...
```

### Not addressed here

The CSRF guard binds cookie sessions only:

```cpp
if(request->method() != HTTP_GET && usedCookie) {
```

A browser holding cached Basic credentials resends them cross-site — `SameSite` does not apply to Basic auth — so a cross-origin form POST to `/config`, `/override`, `/limit` or `/claims` passes both checks. Closing that needs either a real CSRF token or a decision not to treat Basic auth as sufficient for state changes, which seemed worth discussing before implementing.


RAR:
Pushed the submodule bump this PR was waiting on: **gui-nightshift 2026.30.1 → 2026.30.2** (`db22e923`), with `web_static` regenerated.

That release fixes local firmware upload in the web app, which answered `403 {"msg":"csrf"}` for any user logged in via the session cookie on a password-protected device — `uploadFirmware()` is a raw `fetch` with a `FormData` body and did not set the `X-Requested-With` header that `httpAPI()` adds to every other request (OpenEVSE/openevse-gui-nightshift#70).

That mattered here specifically. This PR removes the built-in upload form on the grounds that uploading belongs to the web app, so the web app's uploader had to actually work first. Merging without it would have left password-protected devices with no browser upload path at all.

Re-embedded assets: index, vendor and css chunks changed hash; the language chunks are byte-identical and keep theirs. Three stale headers removed.

The bump also carries the dependency updates merged since 2026.30.1 — playwright 1.62.0, svelte 5.56.8, jsdom 30.0.1, jest-dom 7.0.0, ws 8.21.1, vite 8.1.5, tailwindcss/vite 4.3.3, actions/setup-node 7.

Built and OTA'd to a 16MB ESP32-WROOM; device rebooted into it and serves the new bundle.

RAR:
End-to-end verified in a browser, which was the one thing the earlier evidence did not cover.

On a 16MB ESP32-WROOM running this branch with the 2026.30.2 bundle embedded: password set through the web app, logged in via the login page, then Settings → Firmware → choose a local `.bin` → Install. The upload ran at normal speed and the device rebooted into the new image.

Before GUI 2026.30.2 that same action returned `403 {"msg":"csrf"}` and the upload never started, because the page's `fetch` did not set `X-Requested-With`.

So the sequence this PR depends on now holds on real hardware: the built-in form is gone, and the web app's uploader — the thing it defers to — works for a cookie-authenticated user on a password-protected device.

chris1howell (approved):

PR #1187, opened 2026-08-08
fix(evse): reconcile cached controller state instead of latching it once

RAR:
Three symptoms on a live unit turned out to be one shape of bug: **controller state read once into a member, then never reconciled with its source.**

| symptom | cached thing | latched at |
|---|---|---|
| diode + vent checks read disabled in the UI while the controller had them enabled | `_settings_flags` | `evseBoot()` |
| `service: 1` reported on a unit running L2 | same flags word | same |
| `vehicle: 1` for ~an hour after the car was unplugged | `_sleepForDisable` → pauses land in DISABLED, where the controller stops tracking plug/unplug | flags config-change handler |

### 1. Settings flags never re-read

`_settings_flags` is written in exactly three places: `evseBoot()`, after `setServiceLevel()`, and after `enableFeature()`. Nothing re-reads `$GE` during normal operation, so once cache and controller disagree they stay that way.

Observed, controller firmware `9.0.0.SAMD`:

```
$GE  →  $OK 16 1121
```

`0x1121` = `CGMI | MONO_LCD | AUTO_SVC_LEVEL_DISABLED | L2`. Neither `ECF_DIODE_CHK_DISABLED (0x0002)` nor `ECF_VENT_REQ_DISABLED (0x0004)` is set — both checks enabled on the controller — while `/config` reported `diode_check: false`, `vent_check: false`, `service: 1`.

The tell: toggling *either* check in the UI appears to fix *both*. That isn't the write landing, it's `enableFeature()` refreshing the flags as a side effect.

Fixed by polling `$GE` every 60 counts, offset off the tick shared by the state and temperature polls (both 30, which 60 is a multiple of). `_settings_changed` fires only on an actual change, so the steady state costs one RAPI read per minute and pushes no events.

I checked whether the controller could just tell us instead — `ECVF_CHANGED_TEST`, which drives async `$ST` notifications, covers `AUTH_LOCKED | EV_CONNECTED | TIME_LIMIT | CHARGE_LIMIT | HARD_FAULT`. The `ECF_*` settings flags aren't in it, so there's no event to subscribe to and polling is the only option against existing controller firmware.

### 2. Pause mode latched, putting the EVSE in DISABLED instead of SLEEPING

`_sleepForDisable` was written only by the config-change handler, never at boot, so it held the constructor default until some unrelated `flags` key happened to be written — and a value latched at the wrong moment stuck for the life of the boot.

Measured across two boots of the same unit, same config, same firmware, only a power cycle between them:

```
boot A:  92,094 samples at state 254 (SLEEPING),  0 at 255
boot B:   2,420 samples at state 255 (DISABLED),  0 at 254
```

`sleepForDisable()` now reads config at the point of use and the member is gone. `setSleepForDisable()` keeps its job of re-applying the choice to a pause already in progress, but checks the controller's actual state first, since the config handler fires for every flags key rather than just this one.

### 3. `EV_CONNECTED` trusted while the pilot is at N12

This is downstream of # 2, and it's the one a user actually notices.

```c
void J1772EVSEController::Disable() {
    m_Pilot.SetState(PILOT_STATE_N12);   // −12 V
    m_EvseState = EVSE_STATE_DISABLED;
    // ECVF_EV_CONNECTED is not cleared
}
void J1772EVSEController::Sleep() {
    m_Pilot.SetState(PILOT_STATE_P12);   // +12 V — detection still works
}
```

`ECVF_EV_CONNECTED` is defined in that same file as *"valid only when pilot not N12"*. `Disable()` sets exactly that and leaves the flag stale, so `/status` reported `vehicle: 1` for about an hour after the car was unplugged. `$GS` read `ff 2937 02 0500` — `EV_CONNECTED | EV_CONNECTED_PREV`. Waking the EVSE for three seconds cleared it to `ff 2937 01 0200`.

`isVehicleConnected()` now returns false when the EVSE is DISABLED. SLEEPING is deliberately excluded — it holds P12 and keeps detecting, so the flag stays trustworthy there.

**This guard is a deliberate trade.** While DISABLED we genuinely cannot know whether a vehicle is present, and reporting "not connected" is wrong if one is plugged in. I picked it as the safer error: a stale "connected" is what caused the observed bug, `pause_uses_disabled` defaults to false so the path is opt-in, and a controller holding the pilot at -12V has electrically disconnected from the vehicle anyway.

**To be clear, the controller is not at fault here.** It already exposes `$G0`, which returns 0/1/**2 = unknown** and explicitly checks `GetPilot()->GetState() == PILOT_STATE_N12`. Confirmed against the live unit while it was in this state:

```
$G0  ->  $OK 2          (unknown - correct)
$GS  ->  $OK ff 2937 01 0200
```

`$GS`'s vflags are raw internal state, documented as conditionally valid; `$G0` is the API that answers this question properly. We were reading the wrong source.

The strictly-correct fix is therefore to consume `$G0`, which would also cover the two transient N12 paths (boot-time service-level detection, and `Calibrate()`). I used the DISABLED guard instead because `$G0` costs an extra RAPI round-trip on every state read, and DISABLED is the only N12 state that persists -- the other two are momentary. Happy to switch to `$G0` if you'd rather have exactness than the saved round-trip.

I had initially planned an open_evse change to clear the flags in `Disable()` and checked it against `dev` first. It would have been wrong twice over: `$G0` already covers the case, and `ClrEvConnected()` additionally sets `ECVF_SESSION_ENDED`, so clearing both connect flags would make the next re-enable look like a fresh connect and reset `m_AccumulatedChargeTime`/`m_ElapsedChargeTime` at `J1772EvseController.cpp:1479`. Anyone pausing via DISABLED would lose accumulated session time on resume. No controller change is needed.

### Cost and testing

`openevse_wifi_v1`: +340 B flash total, no RAM change. Builds clean.

**Not yet hardware-validated.** The unit that exhibits all three is mid-way through an unrelated memory-fragmentation investigation and restarting it would discard that run. I'll confirm on hardware once that resolves and report back here rather than imply validation it hasn't had.


RAR:
Hardware-validated on the unit that exhibited all three symptoms (controller fw `9.0.0.SAMD`). OTA'd this branch, then measured. Reporting what was actually proven and what wasn't.

### Fix 2 (pause mode) — conclusive

Same boot, before and after the OTA, sampling `/status` every 2s:

```
before:  4,026 samples at state 255 (DISABLED),   0 at 254
after:     134 samples at state 254 (SLEEPING),   0 at 255
```

`$GS → $OK fe 2937 01 0200`. It now honours `pause_uses_disabled: false` and pauses into SLEEPING. This is the fix that matters most, because SLEEPING holds the pilot at P12 and the controller keeps tracking plug/unplug.

### Fix 1 (settings flags) — symptom resolved, poll confirmed, heal not directly proven

All previously-disagreeing fields now match the controller:

```
$GE -> $OK 45 1121        (0x1121: DIODE_CHK_DISABLED and VENT_REQ_DISABLED both clear, L2 set)

              controller   esp32 /config
diode_check      enabled   true    (was false)
vent_check       enabled   true    (was false)
service level         L2   2       (was 1)
```

`$GE` is visible in the `/evse` ring buffer interleaved with the `$SY` heartbeats, so the new poll is running at its cadence.

**What I could not prove:** that a divergence appearing mid-run gets healed. To test that properly I needed to change a setting on the controller *behind* the ESP32's back, and the web RAPI endpoint correctly blocks writes (`RAPI_RESPONSE_BLOCKED` on `$FF B 0`). Going through `/config` instead would prove nothing, since that path already refreshes the flags as a side effect. Post-OTA agreement is also confounded by the reboot, since `evseBoot()` reads `$GE` anyway. So: symptom gone and poll confirmed live, mechanism inferred from the code rather than demonstrated end-to-end. Flagging that rather than overclaiming.

### Fix 3 (EV_CONNECTED guard) — not exercised, by design

With Fix 2 working the EVSE no longer enters DISABLED, so the stale-flag path doesn't arise. `$G0 → $OK 0` and `/status vehicle: 0` agree with the car being unplugged. The guard stays as protection for anyone who sets `pause_uses_disabled: true`, where DISABLED is entered deliberately.

### Note on the OTA

Went through cleanly (HTTP 200, 2,505,548 B in 33s) even with `heap_largest` down at 14,836 on a unit that had been up ~5.6h.

chris1howell (approved):

PR #1188, opened 2026-08-09
fix(ota): defer GitHub release redirects until the previous TLS connection is freed

KipK:
## Summary

Currently, GitHub-based HTTP OTA updates fails, getting stuck at:

> Starting update — the charger is downloading the release…

with the progress bar remaining at 0% and no visible error.

## Root cause

GitHub release downloads use a redirect chain:

1. `github.com`
2. `release-assets.githubusercontent.com`

The firmware started the redirected HTTPS request directly from the first
request's response callback. At that point, Mongoose had not yet destroyed the
first connection or freed its TLS context.

This temporarily required two TLS contexts. On memory-constrained ESP32 builds,
allocating the second connection could fail silently. The `/update` endpoint had
already returned `{"msg":"started"}`, so the UI continued waiting for an OTA
event that would never arrive.

@RAR probably linked to your research on heap memory issues.

The issue was latent in the redirect flow and became reproducible as the
firmware's available heap margin decreased.

## Changes

- Store redirect information in the current HTTP update state.
- Start the redirected request from `http_update_loop()`, after
  `Mongoose.poll()` has returned and the previous TLS connection has been
  destroyed.
- Support HTTP redirect codes 301, 302, 303, 307, and 308.
- Keep progress, success, and error callbacks in the request state instead of
  copying them into every callback lambda.
- Preserve URL allowlist validation for every redirect target.

## Result

Only one outbound OTA TLS connection is alive at a time. GitHub release
downloads can follow their CDN redirect without requiring two simultaneous TLS
contexts or silently stalling at 0%.


RAR (approved):
The diagnosis matches what I saw in the heap investigation: two concurrent mbedTLS contexts is exactly what a low `heap_largest` can't absorb, and the failure is silent because `/update` has already answered.

Verified the mechanism against the library source: `MongooseHttpClient` sets `MG_F_CLOSE_IMMEDIATELY` after delivering the reply, so `onClose` fires promptly, the TLS context is freed when the handler returns, and `http_update_loop()` after `Mongoose.poll()` starts the next hop with strictly one context alive. Allowlist is re-validated on every hop since `http_update_from_url` is re-entered, and a 3xx without a `Location` degrades to an allowlist reject rather than a hang.

One non-blocking nit: there's no redirect-hop cap, so two allowlisted hosts pointing at each other would loop forever (sequentially, so no heap harm — just an OTA that never ends). A counter in the state capped at ~5 would close it; fine as a follow-up.

Was this run against a real GitHub release URL on hardware? Just want that on record before merge.

KipK:
> Was this run against a real GitHub release URL on hardware? Just want that on record before merge.

Yes I had it tested from UI directly, it fixes the issue correctly.

PR #1195, opened 2026-08-21
feat(evse_man): device-timed claim expiry (duration/remaining)

RAR:
Adds an optional, firmware-enforced expiry to EVSE claims: a claim (or `/override`) may carry `duration` (seconds), and when the timer elapses the firmware releases the claim exactly as if the client had called release — version counters bump and every connected client converges.

## Why

The web UI's Boost button forces charging for a preset time, but the timer lives in a browser `setTimeout`. Close or refresh the tab mid-boost and the override never ends — the charger stays forced on until someone flips the mode. A second client (another phone, HA) sees the override but has no countdown and no idea when it ends. Root cause: claims have `auto_release` (release at session end) but no "release at time T". Moving the timer into the firmware makes Boost durable across refresh/close and consistent across clients — and any claim client (REST, HA, OCPP) gets timed auto-release for free.

## API

`POST /override` (and `POST /claims/{client}`):
```jsonc
{ "state": "active", "charge_current": 32, "auto_release": false, "duration": 900 }
// seconds from now; 0/omitted = no timer (unchanged); "clear" = cancel on re-claim
```

`GET /override` / `GET /claims` while a timer is live:
```jsonc
{ "state": "active", "charge_current": 32, "auto_release": false, "remaining": 842 }
```

- `remaining` is relative (integer seconds, ceil'd) so it is correct even with unsynced NTP; clients derive their own end time.
- Re-claiming always restarts the timer ("press Boost again" = restart); re-claiming without `duration` removes it. `duration` deliberately never persists across deserializations — it is a one-shot request-side instruction (this also keeps it from sticking to the persistent `EvseProperties` objects on the MQTT set path).
- Clamped to 7 days: past ~24.8 days a `millis()`-based deadline crosses the signed rollover horizon and would read as already expired.
- Expiry = release, not restore: the EVSE falls back to its next-priority claim or default state.
- `auto_release` is orthogonal and unchanged; claims remain RAM-only, so a reboot still clears them.

## Implementation

- `EvseProperties` gains `_duration` (parsed from JSON, excluded from equality — it is not a target-state property).
- `Claim` stores a `millis()`-based deadline; the rollover-safe math lives in a new pure header `src/claim_timer.h`.
- `EvseManager::loop()` ticks at 1 Hz while any claim has a deadline (otherwise sleeps at `Infinate` as before), collects expired claims, and releases them through the existing `release()` path.
- `handleOverrideGet` and the claim serializers report `remaining`. The `/claims` endpoints needed no changes — they already round-trip through `EvseProperties`.

## Tests / verification

- New host suite `test/test_claim_timer/` (doctest, `native_test`): 8 cases / 21 assertions covering the deadline boundary, ceil rounding, the `millis()` wrap, and the clamp. All native suites pass.
- `openevse_wifi_v1` and `openevse_wifi_tft_v1` build clean.
- Verified on hardware (wifi_tft_v1 + 9.0.0 SAMD controller): a 60 s boost armed over REST counted down in `GET /override`, the vehicle charged for the window, and the firmware released the claim at +60 s (±1 s tick) with the state falling back as expected. Also exercised: same-duration re-POST restarts, absent-key removes the timer, `"clear"`, clamp saturation, junk values, and `DELETE` mid-timer.

## Companion

OpenEVSE/openevse-gui-nightshift#89 moves the GUI Boost button onto this contract (deletes the browser timer, renders the countdown from `remaining`, and falls back gracefully on firmware without this feature).


jeremypoulter:
This isn't the correct way to solve that problem, the boost feature should be implemented using limit (or a variant of it as I don't remember if the rules have been changed to not claim immediately)

RAR:
Thanks @jeremypoulter. So I started from Limit, but Boost diverges from it in three ways that I couldn't make work without changing Limit's behavior for existing users:

1. Clock base — limitTime() compares against getSessionElapsed(), i.e. charging minutes since the session started. Pressing Boost 90 minutes into a session with a 60-minute boost would expire immediately. Boost needs wall-clock time from activation.
3. Expiry action — on limit reached, Limit claims EvseState::Disabled at Priority_Limit and holds that until the vehicle disconnects. Boost wants release, not disable: when the boost window ends, the scheduler/eco logic should regain control (e.g. an off-peak schedule can still start charging later that night). Holding a Disabled claim until unplug blocks that.
4. Pause behavior — the limit only ticks while isCharging(), so a paused session stretches the window indefinitely, where a boost is a fixed window.

So the duration on a claim is essentially the variant you mention: a wall-clock timer that releases (rather than disables) on expiry. I put it at the claims layer instead of inside Limit because it then works for any client and keeps Limit's disable-on-reached semantics untouched — and remaining falls out naturally in GET /override and /claims for the UI countdown.

That said, if you'd rather the timer live in Limit, I'm happy to rework it — something like a wall-clock limit type with release-on-expiry, plus remaining reporting added to /limit. The trade-off is the UI has to make two coupled writes (claim Active + set limit) with no atomicity, and clearing one without the other leaves a stale half-state, which is what pushed me toward a single property on the claim.

Lemme know what direction you'd prefer?

jeremypoulter:
@rar yes I guess they are good points I hadn't considered.

I think boost should still live as an explicit feature in the backend rather than mostly in the UI, will help others (Home Assistant) support it with the same logic. I also think that we should support all the same features we limit on for boost as well which is largely why I had it in my head that they were the same thing as limit.

RAR:
Yeah, I'm cool with that — a Boost feature in the backend makes sense, and reusing the same dimensions as Limit keeps the two symmetric: Limit = "stop when reached, hold disabled", Boost = "run until reached, then release back to whatever was in control".

Something llke...

- A Boost module alongside Limit, sharing the threshold checks (time/energy/SoC/range), but time is wall-clock from activation and reaching the threshold releases the claim rather than disabling.
- POST /boost with {"type": "time", "value": 3600} (same type/value vocabulary as /limit), GET /boost reporting the active boost + remaining, DELETE /boost to cancel early. The module owns the Manual-priority Active claim, so HA and the web UI drive it the same way.

One question: do you want the generic duration on claims (what this PR currently adds) to survive as the primitive underneath — Boost's time dimension would just use it, and other claim clients get expiry for free — or would you rather the timer be private to the Boost module and this PR be changed entirely into the /boost feature? Either way is just fine, just changes whether I extend this branch or restart it.

jeremypoulter:
@RAR I think I would remove the time from the claim to keep that layer simple. You may however want a layer that both boost and limit sit on top of.

Please also look at adding at least Boost support to the simulator so we can run simulations on how it interacts with the other modules

RAR:
Closing this in favour of #1197, which implements the direction agreed here: duration comes off the claims layer entirely and Boost is its own backend module with the /limit type vocabulary, a shared threshold layer under Limit, and simulator coverage.

PR #1196, opened 2026-08-22
web: serve the UI shell with max-age=86400 to fix iOS 17 splash-reload on every nav tap

RAR:
## Why

On iOS 17 Safari and WKWebView, tapping any bottom-nav item in the web UI does a full page reload — splash screen on every tap, even re-tapping the current tab. iOS 18+ and current desktop Safari are unaffected (fixed in newer WebKit).

Root cause was isolated by A/B bisection against a real charger and a real iOS 17 device, serving the charger's exact GUI bytes from a test server with controlled headers and latency:

| Static serving | Latency | Result |
|---|---|---|
| `max-age=30, must-revalidate` (current firmware) | fast | clean |
| `max-age=30, must-revalidate` | 400 ms (ESP32-like) | splash on every tap |
| `max-age=86400` | 400 ms | clean |

iOS 17 WebKit promotes a same-document `#hash` link tap into a full renavigation when the document's cache entry is stale and needs revalidation against a slow server. The ESP32 is always slow (~100–500 ms/response), and `max-age=30, must-revalidate` makes the shell go stale almost immediately — so on iOS 17 every tap renavigates. The GUI code, the wire responses (all clean 200/304/101 in a full byte capture), and the mDNS/WebSocket layers were ruled out.

## What

One line: the non-immutable Cache-Control branch (index.html, manifest, favicon, icons) goes from `public, max-age=30, must-revalidate` to `public, max-age=86400`. The content-hashed `/assets/` branch is unchanged (already `immutable`), and the ETag/304 path is unchanged — an explicit browser refresh still picks up a new GUI build immediately.

Trade-off accepted: after a firmware/GUI update, plain navigations may show a stale shell for up to 24 h unless the user refreshes. Assets are content-hashed, so a stale shell is self-consistent — it points at its own hashed assets, which remain embedded.

## Verification

On hardware (`openevse_wifi_tft_v1`):
- `curl -sI http://<ip>/` → `Cache-Control: public, max-age=86400`
- `curl -sI http://<ip>/assets/<hash>.js` → `public, max-age=31536000, immutable` (unchanged)
- `curl -sI -H 'If-None-Match: <etag>'` → `304 Not Modified`
- iOS 17 device: load, reload once to prime the cache, then tap between bottom-nav tabs incl. re-tapping the current tab — no splash reloads.


RAR:
Pushed a second commit: it turns out the max-age change alone only holds for exactly one day. A bare `304` doesn't renew the cached entry's freshness (RFC 9111 §4.3.4 renews it *from the 304's headers*, and we sent none), so 24h after the last full 200 the shell goes permanently stale — every revalidation succeeds and none of them helps, and the iOS 17 splash-on-every-tap behaviour comes back, curable only per-visit with a manual reload. That's exactly what we observed in the field: the fixed device regressed ~24h after flashing.

The fix sends the 304 through the normal response object so it carries `Cache-Control` (resetting the freshness clock on every revalidation) plus the quoted `ETag` (RFC 7232 §4.1). The explicit zero `Content-Length` matters: the response object's default would serialize the 304 as `Transfer-Encoding: chunked` with no terminating chunk.

Verified on hardware:
```
$ curl -sI -H 'If-None-Match: "<etag>"' http://<charger>/
HTTP/1.1 304 Not Modified
Cache-Control: public, max-age=86400
Etag: "<etag>"
Content-Length: 0
```

chris1howell (approved):

PR #1200, opened 2026-08-24
deps: consume esp_tsdb from upstream instead of vendoring it

RAR:
### What

Stop vendoring `esp_tsdb` and reference the upstream `v2.3.0` tag as a PlatformIO git dependency. Deletes `components/esp_tsdb/` — 1,426 lines.

### Why the copy existed, and why it no longer needs to

It was vendored when the fixes it needed lived only on a fork. Those are all merged upstream now:

- [zakery292/esp_tsdb#1](https://github.com/zakery292/esp_tsdb/pull/1) — handle-based multi-instance API
- [#2](https://github.com/zakery292/esp_tsdb/pull/2) — wrapped-ring query iteration
- [#3](https://github.com/zakery292/esp_tsdb/pull/3) — `MALLOC_CAP_8BIT` for the internal-RAM pool (IRAM `LoadStoreError` on xtensa)

I diffed the vendored tree against the `v2.3.0` tag: **every source file is byte-identical**. The only extra file was an added `library.json`. So the copy carried no local changes worth keeping.

A future bump is now a one-line ref change instead of a re-vendor.

### The one caveat

Upstream ships no `library.json` — it is packaged as an ESP-IDF component (`CMakeLists.txt` + `idf_component.yml`). PlatformIO therefore reports the package as `esp_tsdb@0.0.0+<sha>` and cannot do version resolution; **the git ref is what pins it**, which it does exactly. The LDF still picks up `src/` and `include/` by convention, and `host_test/` correctly stays out of the build.

If you would rather have real version metadata, I am happy to open a PR upstream adding a `library.json`; the pin works without it.

This does make the build reach the network for this dependency where the copy built offline — consistent with the rest of `lib_deps`, but worth naming.

### What v2.3.0 fixes

This started as a version bump, and those reasons still stand:

- **2.2.0 — O(log n) query start seek.** The reason I opened this. A full `/energy/daily` query on a ~100-day store ran long enough to trip the task watchdog and reset the device. After: HTTP 200 in ~1.7 s.
- 2.2.0 — header/index overlap fix; strict `num_params` check on open; schema migration.
- 2.3.0 — free-space guard, close-lifecycle use-after-free fixes, `tsdb_resize()`.

### On-disk compatibility

Existing databases open unchanged, verified on hardware before and after:

- 7 columns (`TSDB_NUM_COLS`) keeps it a V3 file; the V4 wide format only applies above 16 columns.
- The sparse-index offset is stored per file and honoured, so existing 512-offset databases still open (2.2.0 moved *new* files to 1024).
- `num_params` matches the config, so 2.2.0's strict open check passes.

After flashing: `tsdb_ready` 1, `tsdb_err` 0, `littlefs_used` unchanged (no recreate), and the aggregate endpoints return **byte-identical** results to a pre-upgrade capture — md5-matched across three separate reflashes, including data written by the old engine.

### Testing

Built for `openevse_wifi_tft_v1` and `openevse_wifi_v1_16mb`. Run on two 16 MB units, one with a two-day store and one with ~100 days; the `/energy/daily` watchdog reset is gone.


jeremypoulter:
Does this have to be copied into the source tree? Can we reference the source as a library/depedancy in Platform IO? aGit submodule?

RAR:
I think originally I did it that way because it was off my fork.  Thats not the case anymore, so I think we can just include it.  Lemme go back and check my notes tho.

RAR:
@jeremypoulter yup, that was the only reason why.  Swapped.

jeremypoulter:
@rar cool, I wonder if including as a git submodule would be better though, just so dependabot can keep it up-to-date. Or are there any other options for that?

RAR:
I think it checks dependencies?  I'd have to look.

jeremypoulter:
@RAR only with known package management systems, Platform IO isn't supposed, but we do have a workflow that supports libraries. But this isn't covered by either

RAR:
Looked at the options:

1. **Publish `esp_tsdb` to the PlatformIO registry** — it already has a `library.json`, so it's just `pio pkg publish` on their side. The dep then becomes `zakery292/esp_tsdb@^2.3.0` and the existing platformio_dependabot workflow picks it up with no new infrastructure. I'll reach out to the esp_tsdb maintainer and ask.
2. Git submodule — covered by the existing Dependabot config, but it bumps to latest branch commit rather than release tags, so we'd lose the deliberate version pinning.
3. A small scheduled workflow comparing the pinned tag against upstream tags — works, but it's more bespoke automation to maintain.

Going with 1; I'll leave this PR on the git tag pin until it's published, then swap to the registry version.

One caveat either way: a niche C library won't get GitHub advisory coverage under any manifest format — the vendored copy was ironically the only version CodeQL could scan. Freshness tracking is the realistic win here.

jeremypoulter (approved):

PR #1203, opened 2026-08-25
feat(mqtt): make the $SYS/broker/version probe optional

RAR:
## Problem

`Mqtt::onConnect()` subscribes to `$SYS/broker/version` unconditionally to
populate `mqtt_broker_version`. That works on Mosquitto and most self-hosted
brokers, but managed brokers have no `$SYS` tree.

AWS IoT Core treats a subscribe to an unauthorised topic as a policy violation
and answers by **closing the connection**, not by returning a failed SUBACK. The
client then reconnects, subscribes again, and is dropped again — so a device
configured against IoT Core never establishes a working session.

This is why it can't be handled as "subscribe and tolerate the failure": there
is no failure to tolerate, only a dropped connection, and nothing in the reply
distinguishes it from an ordinary disconnect.

## Change

Adds a `mqtt_sys_query` option (short name `msq`) gating the probe. One new
flag bit, one accessor, one `if`.

**Stored with inverted sense** — bit set = disabled — matching the existing
`mqtt_reject_unauthorized` / `CONFIG_MQTT_ALLOW_ANY_CERT` pattern. This is
load-bearing rather than stylistic: `flags` is read from NVS wholesale and
`flags_changed` only records options a user has explicitly set, so a newly
added bit reads 0 on every existing install. With normal sense, every current
user would silently stop seeing their broker version after upgrading. Inverted,
existing installs keep probing exactly as they do today and only a deliberate
opt-out changes behaviour.

## Testing

Hardware, against Mosquitto 2.1.2:

| | `mqtt_sys_query` | `mqtt_broker_version` |
|---|---|---|
| default | `true` | `"mosquitto version 2.1.2"` |
| after `POST {"mqtt_sys_query": false}` | `false` | unset |

The reconnect after toggling is clean and `mqtt_connected` stays 1, so no
`$SYS` subscribe goes out and the field correctly reads unknown rather than
going stale.

Builds green on `openevse_wifi_v1` (4 MB, the tightest env): flash 94.8%,
RAM 23.7%.

## Note for the GUI

The option is exposed on `/config` but has no UI yet — a checkbox alongside
the other MQTT settings would finish it. Defaulting on means nothing changes
for existing users until they turn it off.

### DIFF
--- a/src/app_config.cpp
+++ b/src/app_config.cpp
@@ -327,6 +327,7 @@ ConfigOpt *opts[] =
   new ConfigOptVirtualMaskedBool(flagsOpt, flagsChanged, CONFIG_SERVICE_MQTT, CONFIG_SERVICE_MQTT, "mqtt_enabled", "me"),
   new ConfigOptVirtualMaskedBool(flagsOpt, flagsChanged, CONFIG_MQTT_ALLOW_ANY_CERT, 0, "mqtt_reject_unauthorized", "mru"),
   new ConfigOptVirtualMaskedBool(flagsOpt, flagsChanged, CONFIG_MQTT_RETAINED, CONFIG_MQTT_RETAINED, "mqtt_retained", "mrt"),
+  new ConfigOptVirtualMaskedBool(flagsOpt, flagsChanged, CONFIG_MQTT_NO_SYS_QUERY, 0, "mqtt_sys_query", "msq"),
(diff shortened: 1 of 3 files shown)


jeremypoulter (commented):
So we need an option for this? Shouldn't it just fail and the data not being available? Think it is just information right?

RAR:
Yup, just information, but some brokers will disconnect you if you ask for something you don't have access to leading to a connection/disconnect loop.

chris1howell (approved):

PR #1204, opened 2026-08-26
build: Arduino as an ESP-IDF component for the core-3 envs (+ the memory it unlocks)

RAR:
### What

Switches the two core-3 envs (`openevse_wifi_tft_v1`, `openevse_wifi_v1_16mb`) from Espressif's prebuilt Arduino archives to `framework = arduino, espidf` — ESP-IDF built from source with the Arduino core as an ordinary IDF component. The Arduino API is unchanged; every `lib_deps` library still builds through the normal LDF. The 4 MB core-2 envs are untouched (the CMake scaffolding is inert for them).

Draft because it deliberately bundles the sdkconfig-unlocked memory work on top of the build change, and some of it may want splitting out (notably the gui bump).

### Why

Three problems with the prebuilt archives, all hit on the no-PSRAM TFT board:

1. **sdkconfig is read-only.** Half of IDF's memory/TLS knobs silently do nothing (`custom_sdkconfig` values that don't exist in the prebuilt config just drop). With IDF built from source, a committed `sdkconfig.defaults` actually governs the build.
2. **`--wrap` link flags don't propagate.** mbedTLS dynamic buffers need 12 `-Wl,--wrap=mbedtls_ssl_*` flags that IDF components publish as INTERFACE link flags; with prebuilt archives they must be hand-maintained in `build_flags` and silently rot when IDF adds one. In hybrid mode the CMake File API carries them automatically — verified by `nm` on the output ELF with zero hand flags.
3. **`custom_sdkconfig` thrashes CI.** pioarduino rebuilds libs in place in the shared framework package, and any env *without* `custom_sdkconfig` triggers a full framework wipe/re-download — so alternating env builds reinstall the toolchain every time. Gone entirely in hybrid mode.

### Mechanism

- Root `CMakeLists.txt` + `src/CMakeLists.txt` (required by pioarduino's espidf builder; no-ops for core-2 envs).
- Committed `sdkconfig.defaults`: the verbatim seed from the prebuilt package's config, plus a clearly-marked **PROJECT OVERRIDES** section (last value wins) so the diff against stock Arduino is auditable.
- Per-env overlays via `board_build.cmake_extra_args = -DSDKCONFIG_DEFAULTS=sdkconfig.defaults;sdkconfig.defaults.tls` — the TFT env layers TLS settings the 16 MB env doesn't need.
- Vendored `esp_tsdb` moved `components/` → `vendor/` so IDF's component auto-scan doesn't double-build it (interim state; #1200 removes the vendor copy entirely).

One caveat: a **nonexistent** Kconfig symbol still drops silently (negative-tested). What hybrid buys is that *real* options take effect, with dependency auto-selection. Verification is the generated `sdkconfig` + `nm`, not a green exit.

### Memory results (bench-measured on the TFT board, WROOM-32, no PSRAM)

| knob | effect |
|---|---|
| `CONFIG_BT_ENABLED=n` | +22.5 K largest free block — the BT DRAM region is **not** auto-reclaimed at boot; firmware never calls `esp_bt_controller_mem_release` and never uses BT |
| `CONFIG_SPIRAM=n` | no PSRAM on these boards; drops probe + support structures |
| WiFi static RX 8→6, dynamic RX/TX 32→16 | ~3 K DRAM; workload is small JSON/MQTT frames |
| mbedTLS dynamic buffers (TLS overlay) | TLS sessions release their 16 K buffers between records; direct-TLS MQTT now works on this board |
| `CONFIG_SPI_FLASH_DANGEROUS_WRITE_ALLOWED=y` | enables the in-place 4→16 MB repartition path discussed in #1136 (ESPHome-proven mechanism); no behavior change otherwise |
| coredump-to-flash (ELF) | came free with the seed; already used in anger to pull and symbolize a real panic from the coredump partition |

Net on the bench TFT: free heap ~64 K → **~101 K**, boot heap-min watermark 608 B → **26.7 K**. (Fragmentation itself is not fixed by any of this — that work is separate.)

### Also in the branch

- `mqtt_sys_query` config flag: makes the `$SYS/broker/version` probe optional — some brokers (AWS IoT) kill the session on a `$SYS` subscribe.
- gui-nightshift bump to current main + regenerated embedded web assets (happy to split this out).
- The design doc and implementation plan under `docs/superpowers/`.

### Costs / status

- Cold build compiles IDF from source (minutes, once); incremental rebuilds are ~2.5 min, faster than the prebuilt flow's reinstall thrash. CI wall time for the two hybrid envs will go up on cold caches.
- Bench HW-validated end-to-end (boot, WiFi, web, OTA, MQTT/mqtts, display); multi-hour soaks run with flat heap. Draft until a longer soak completes and reviewers weigh in on the bundling.


RAR:
Recreating from an in-repo branch so CI runs without the fork-approval gate — see the replacement PR.

jeremypoulter:
I don't think this is an issue. We used to build IDF from source, IIRC we had to switch back because the config became out of sync and made updates hard, so would be good to have a plan for that. Should probably make sure the IDF build is cached in the workflows so we can scale up the CI/CD builds core 3. Which reminds me can we just make everything build with core 3 and some of the other new features, TSDB, LVGL, etc, I think the old code/support can just be removed simplifying the code base

Issue #1202, opened 2026-08-25
Websocket eventing audit: /status fields that never event, and evented fields absent from /status


RAR:
### Summary

Split out from a side discussion on #1199, where the stated model for clients was:

> You should not be depending on stuff repeatedly calling status, as nothing should be doing that, well behaved clients should call status once on load then use the websocket to receive updates as they happen. For that matter really on connect to ws you should just get the same content as `/status`.

I agree with that model, so this issue audits how far the firmware currently is from it. Everything below is against `master` (ace7bb56).

Note on the last sentence of that quote: on-connect parity is already fixed in #1199. `onWsConnect()` sized its document for 40 members while `buildStatus()` emits ~95, so every websocket connect was pushing a silently truncated status (ArduinoJson does not report this — it just stops adding members). Both call sites now share one capacity constant. No action needed here for that part.

The rest of the gap falls into two classes:

- **A.** fields in `/status` that are never evented, so they go stale until the client reloads
- **B.** fields that are only ever evented, so a client connecting later never learns them

#### A1 — event on change; cheap and exact

`mode`, `rapi_connected`, `ocpp_connected`, `service_level`, `rfid_reader`, `rfid_failure`, `ota_update`, `gfcicount`, `nogndcount`, `stuckcount`, `tsdb_ready`, `tsdb_err`, `littlefs_free`, `littlefs_used`

These change rarely — a fault, a reconnect, a settings write — so a compare-and-emit in a background poll is exact and costs nothing at idle. The fault counters in particular are what you end up staring at during a support conversation, and today they are frozen at whatever they were when the page loaded.

(`littlefs_*` is already sampled on a timer as of #1199, so it only needs the comparison added.)

#### A2 — divert inputs

`solar`, `grid_ie`

These are inputs, set via `POST /status` and over MQTT. `divert.cpp` events `available_current` / `smoothed_available_current` when it recalculates, but never echoes the raw inputs back, so a UI showing them renders whatever was current when the socket opened. Best evented at the point of set rather than by polling.

#### A3 — continuously varying; needs a policy decision

`free_heap`, `heap_largest`, `heap_largest_min`, `heap_min`, `stack_loop_min`, `stack_events_min`, `lv_used_max`, `lv_frag_max`, `ws_conns`, `ws_send_max`, `ws_reaped`, `comm_sent`, `comm_success`, `packets_sent`, `packets_success`, `mqtt_last_rx`

[...]

PR #1209, opened 2026-08-30
Read relay contact-life health from the controller, bump OpenEVSE_Lib…

chris1howell:
# Relay contact-life health estimation

Estimates how much life is left in the main charging contactor and surfaces
it through the stack: controller firmware → RAPI → client library → this
gateway firmware. Diagnostic only — never gates or alters charging logic at
any layer.

Spans three repos:

| Repo | Branch/tag | What it adds |
|---|---|---|
| `open_evse` (ATmega/SAMD controller) | `dev` @ `D9.3.0`) | The estimation model itself: `RelayHealth` module, new RAPI commands |
| `OpenEVSE_Lib` (client library) | `OpenEVSE9` `v0.0.22` | `getRelayHealth()` / `resetRelayHealth()` wrappers for the new commands |
| `openevse_esp32_firmware` (this repo) | `relay_health` @ `6856978` | Polls the new data, caches it, exposes it via `/config` JSON |

## Background / the problem

Relay life splits into two budgets that are easy to conflate: **mechanical**
life (cycles with no current flowing — springs/bearings, normally
10⁷–10⁸ cycles) and **electrical** life (cycles under load — contact erosion
from arcing, 10⁴–10⁵ cycles at rated current). For an EVSE contactor the
electrical budget is the one that matters, and the single biggest lever is
whether the contact opens hot (current still flowing) or cold (the car has
already ramped down). A handful of hot disconnects — fault interrupts,
e-stops, GFCI trips — can consume more life than years of normal cold
cycling.

The firmware already measured charging current, starting and peak
temperatures, current on relay open and whether a stop was clean or
fault, but nothing accumulated that into a life estimate, and nothing 
gave early warning of the one failure mode a cycle count can't predict: 
a welded contact.

## Model (`open_evse` firmware)

Cumulative-damage (Miner's rule): each relay open consumes a fraction of a
life budget, summed over the relay's life.

- **Cold open** (current already at/below a zero threshold): `d = 1/N_m`
- **Hot open** (current still flowing): `d = (1/N_e) · (I/I_r)² · α_load · α_temp`

Where `N_e`/`N_m` are the rated electrical/mechanical life (cycles, from the
relay's datasheet), `I_r` is rated current, and `α_load`/`α_temp` are
load-character and temperature penalty multipliers. In a well-behaved
installation the cold-cycle term rounds to ~0 (mechanical life is normally
untouchable at any realistic cycle count) and the reported percentage is
really tracking hot-switch events — exactly what you want surfaced.

Two more independent signals feed into the picture, both self-baselined per
station since enclosure thermals and hardware vary:

- **Coil drop-out (open) transit-time drift** — a lengthening time between
  the open command and the contact physically opening (measured via the
  load-side AC-sense pin, CGMI hardware only) is an early symptom of
  welding, ahead of the hard `EVSE_STATE_STUCK_RELAY` fault that ultimately
  catches it.
- **Thermal index `H = ΔT / I²`** (requires `TEMPERATURE_MONITORING`) —
  proportional to contact resistance. Heating at the contact is `I²·R`;
  normalizing by `I²` removes the load dependence, so `H` stays flat while
  the contact is healthy and climbs as it erodes/oxidizes.

Not implemented: direct contact-resistance measurement via closed-contact
voltage drop — this hardware has no differential voltage sensor across the
contacts to measure it.

### New source: `firmware/open_evse/RelayHealth.{h,cpp}`

Gated by `RELAY_HEALTH`, auto-enabled wherever `RELAY_ZC_SWITCH` + `AMMETER`
are (every current build target). New EEPROM state at offsets 44–55.

### New RAPI commands

| Command | Direction | Purpose |
|---|---|---|
| `$SZ ma` | set | Relay-open current-zero threshold (mA) — runtime-tunable per-unit ammeter noise floor |
| `$GZ` | get | Now returns **two** fields: `freqx100 zerothreshma` (was frequency-only) |
| `$GW` | get | Relay wear diagnostics: `hotswitchcnt lastopenma closetransitms opentransitms` |
| `$GL` | get | Relay life/health: `pctremain coldopencnt elecdamagex1e6 transitbaselinems transitdrift thermalx100 thermalbaselinex100 thermalwarn` |
| `$FH` | set | Reset the health accumulator/baselines (use after a physical relay replacement) |

Full field-by-field reference: `open_evse/firmware/open_evse/rapi_proc.h`.

## Client library (`OpenEVSE_Lib` 0.0.22)

Added `OpenEVSEClass::getRelayHealth(callback)` and `resetRelayHealth(callback)`
in `src/openevse.{h,cpp}`, mirroring the existing `getFrequency()` /
`getRelayStatus()` / `resetFaultCounters()` D9 extensions (same
`isD9Supported()` protocol gate, same token-parsing style). Added the
`OPENEVSE_RELAY_HEALTH_NOT_AVAILABLE` (`0xffff`) sentinel for fields that
aren't available yet (baseline not established) or at all (thermal fields
need `TEMPERATURE_MONITORING` on the controller).

`getFrequency()` itself was **not** changed to expose the new `$GZ`
zero-threshold field, to avoid a breaking signature change to an existing
public method — see below for how this gateway firmware gets that value
instead.

# Stuck-relay recovery support

Client-library support for the controller-side stuck-relay auto-recovery
feature (`open_evse` firmware 9.3.0, `dev` @ `5b3a755`). `OpenEVSE9` @
`2c7c3f3`, `library.json` 0.0.22.

Firmware 9.3.0 added an automatic recovery attempt for a welded/stuck main
contactor: when `EVSE_STATE_STUCK_RELAY` is entered with no EV connected,
the controller cycles the relay (3 rounds of 5 on/off toggles, each shorter
than the last) instead of going straight to an unrecoverable hard fault.
It also added `$FK` to run that same cycle manually, and extended `$GL`
(relay health) with a cumulative attempt counter. Full firmware-side
derivation: `open_evse/docs/stuck_relay_recovery.md`.

## `getRelayHealth()` - new field

The callback signature gained a 10th parameter,
`uint32_t stuck_relay_recovery_count`:

```cpp
OpenEVSE.getRelayHealth([](int ret, uint8_t life_remaining_pct,
                            uint32_t cold_open_count, uint32_t elec_damage_x1e6,
                            uint32_t transit_baseline_ms, bool transit_drift_warning,
                            uint32_t thermal_index_x100, uint32_t thermal_baseline_x100,
                            uint8_t thermal_warning_level,
                            uint32_t stuck_relay_recovery_count)
{
  if (ret == RAPI_RESPONSE_OK) {
    Serial.printf("Recovery attempts: %u\n", stuck_relay_recovery_count);
  }
});
```

Backward compatible on the wire: against a pre-9.3.0 controller, `$GL`
still only returns 8 fields, and `stuck_relay_recovery_count` reads back
`0` rather than the call failing.

## `runStuckRelayRecovery()` - new method

```cpp
void OpenEVSEClass::runStuckRelayRecovery(std::function<void(int ret)> callback);
```

Sends `$FK` to run the recovery cycle on demand - the same routine the
controller runs automatically. The controller NAKs it if an EV is
connected (unsafe to cycle the relay under load).

Uses an explicit **35 second** `RapiSender` timeout, not the library
default (`RAPI_TIMEOUT_MS`, 500ms). The controller call itself is blocking
and can legitimately take up to ~30s (3 rounds x 5 toggles x up to 2s
on+off, plus zero-cross/transit overhead) before it responds - the default
timeout would give up long before that.

Doesn't itself report whether the relay came free; check `getStatus()`
(state) or `getRelayHealth()` (the recovery counter, or re-poll relay
status) afterward.

## `resetRelayHealth()` - behavior change, no API change

`$FH` now also clears the recovery counter server-side, alongside the
existing damage-accumulator and baseline resets. No client-side change
needed - just documented.

## Requirements

Same D9-protocol gate (`isD9Supported()`) as the other linco-work
extensions, and firmware 9.3.0+ / `ADVPWR` for `runStuckRelayRecovery()` and
the recovery counter specifically (older/non-ADVPWR controllers NAK `$FK`
with `RAPI_RESPONSE_FEATURE_NOT_SUPPORTED` and report the counter as 0).

## Gateway firmware (this repo, `relay_health` branch)

`src/evse_monitor.{h,cpp}` — `EvseMonitor` gained:

- **`readRelayHealth()`** — calls `OpenEVSE.getRelayHealth()`, caches all
  eight fields, exposed via `isRelayHealthKnown()` +
  `getRelay*()`/`isRelay*()` getters.
- **`readFrequency()` rewritten** — reads `$GZ` raw via `EvseMonitor`'s own
  `RapiSender*` instead of the library's single-field `getFrequency()`, so it
  now also captures the zero-cross threshold from the same response instead
  of round-tripping `$GZ` twice.

### Polling cadence

| When | Why |
|---|---|
| At boot (`evseBoot()`) | Populate immediately on connect |
| After every charge session (`updateEvseState()`, the existing relay-open `!isCharging()` branch) | `$GL` is only meaningful once an open has occurred — that's when the controller's accumulator actually updates |
| Every ~60s (`getSettingsFromEvse()`, existing settings-poll cadence) | Catches changes from other RAPI clients and long-running sessions with no session boundary, without adding RAPI traffic |

### Exposure

`src/app_config.cpp`'s `config_serialize()` (the `/config` JSON), alongside
the existing `zero_cross`/`relay_dc1` fields:

```json
{
  "zero_cross": true,
  "zero_cross_threshold_ma": 300,
  "relay_dc1": true,
  "relay_life_pct": 100,
  "relay_cold_open_count": 42,
  "relay_elec_damage_x1e6": 0,
  "relay_transit_drift_warning": false,
  "relay_transit_baseline_ms": 18,
  "relay_thermal_warning_level": 0,
  "relay_thermal_index_x100": 12,
  "relay_thermal_baseline_x100": 11
}
```

Each field/block is **omitted** (not defaulted to 0/false) until the
controller has actually answered — same pattern already used for
`relay_dc1`/`relay_dc2`/`relay_ac`, so the GUI can distinguish "not
supported by this controller" from a real zero.

## Verification status

Firmware side (`open_evse`) built clean on `m328p_core`, `m328p_LCD_WIFI`,
`samd`, `samd_ice` via PlatformIO. Gateway side (`openevse_esp32_firmware`)
was **not** compiler-verified: the shared PlatformIO core dir's
`espressif32` platform install and the `native_simulator` env (no `g++` on
PATH) were both broken/unavailable in the session this was built in. Changes
were reviewed by hand against the exact wire format above and the existing
`readFrequency()`/`readRelayStatus()` call patterns, but a real build should
be run before merging.

### DIFF
--- a/docs/relay_health.md
+++ b/docs/relay_health.md
@@ -1,0 +1,200 @@
+# Relay contact-life health estimation
+
+Estimates how much life is left in the main charging contactor and surfaces
+it through the stack: controller firmware → RAPI → client library → this
+gateway firmware → GUI. Diagnostic only — never gates or alters charging
+logic at any layer.
+
+Spans four repos:
+
+| Repo | Branch/tag | What it adds |
+|---|---|---|
+| `open_evse` (ATmega/SAMD controller) | `dev` @ `5b3a755` (built on `442a98f`, `544f277`) | The estimation model: `RelayHealth` module, new RAPI commands, plus stuck-relay auto-recovery ($FK) |
+| `OpenEVSE_Lib` (client library) | `OpenEVSE9` @ `2c7c3f3`, tag `v0.0.22` | `getRelayHealth()` / `resetRelayHealth()` / `runStuckRelayRecovery()` wrappers |
+| `openevse_esp32_firmware` (this repo) | `relay_health` | Polls the new data, caches it, exposes it via `/config` JSON |
+| `openevse-gui-nightshift` | `relay_health` @ `c3f49b8` | Displays it: Monitoring → Health tab's Relay Health section |
+
+## Background / the problem
+
+Relay life splits into two budgets that are easy to conflate: **mechanical**
+life (cycles with no current flowing — springs/bearings, normally
+10⁷–10⁸ cycles) and **electrical** life (cycles under load — contact erosion
+from arcing, 10⁴–10⁵ cycles at rated current). For an EVSE contactor the
+electrical budget is the one that matters, and the single biggest lever is
+whether the contact opens hot (current still flowing) or cold (the car has
+already ramped down). A handful of hot disconnects — fault interrupts,
+e-stops, GFCI trips — can consume more life than years of normal cold
+cycling.
+
+The firmware already measured charging current and knew whether a stop was
+clean or a fault, but nothing accumulated that into a life estimate, and
+nothing gave early warning of the one failure mode a cycle count can't
+predict: a welded contact.
+
+## Model (`open_evse` firmware)
+
+Cumulative-damage (Miner's rule): each relay open consumes a fraction of a
+life budget, summed over the relay's life.
+
+- **Cold open** (current already at/below a zero threshold): `d = 1/N_m`
+- **Hot open** (current still flowing): `d = (1/N_e) · (I/I_r)² · α_load · α_temp`
+
+Where `N_e`/`N_m` are the rated electrical/mechanical life (cycles, from the
+relay's datasheet), `I_r` is rated current, and `α_load`/`α_temp` are
+load-character and temperature penalty multipliers. In a well-behaved
+installation the cold-cycle term rounds to ~0 (mechanical life is normally
+untouchable at any realistic cycle count) and the reported percentage is
+really tracking hot-switch events — exactly what you want surfaced.
+
+Two more independent signals feed into the picture, both self-baselined per
+station since enclosure thermals and hardware vary:
+
+- **Coil drop-out (open) transit-time drift** — a lengthening time between
+  the open command and the contact physically opening (measured via the
+  load-side AC-sense pin, CGMI hardware only) is an early symptom of
+  welding, ahead of the hard `EVSE_STATE_STUCK_RELAY` fault that ultimately
+  catches it.
+- **Thermal index `H = ΔT / I²`** (requires `TEMPERATURE_MONITORING`) —
+  proportional to contact resistance. Heating at the contact is `I²·R`;
+  normalizing by `I²` removes the load dependence, so `H` stays flat while
+  the contact is healthy and climbs as it erodes/oxidizes.
+
+Not implemented: direct contact-resistance measurement via closed-contact
+voltage drop — this hardware has no differential voltage sensor across the
+contacts to measure it.
+
+### New source: `firmware/open_evse/RelayHealth.{h,cpp}`
+
+Gated by `RELAY_HEALTH`, auto-enabled wherever `RELAY_ZC_SWITCH` + `AMMETER`
+are (every current build target). New EEPROM state at offsets 44–55.
+
+### New RAPI commands
+
+| Command | Direction | Purpose |
+|---|---|---|
+| `$SZ ma` | set | Relay-open current-zero threshold (mA) — runtime-tunable per-unit ammeter noise floor |
+| `$GZ` | get | Now returns **two** fields: `freqx100 zerothreshma` (was frequency-only) |
+| `$GW` | get | Relay wear diagnostics: `hotswitchcnt lastopenma closetransitms opentransitms` |
+| `$GL` | get | Relay life/health: `pctremain coldopencnt elecdamagex1e6 transitbaselinems transitdrift thermalx100 thermalbaselinex100 thermalwarn stuckrelayrecoverycnt` (9th field, `stuckrelayrecoverycnt`, added alongside stuck-relay auto-recovery) |
+| `$FH` | set | Reset the health accumulator/baselines/recovery counter (use after a physical relay replacement) |
+| `$FK` | set | Run the stuck-relay recovery cycle manually. NAK'd if an EV is connected; blocking on the controller side for up to ~30s |
+
+Full field-by-field reference: `open_evse/firmware/open_evse/rapi_proc.h`.
+
+## Client library (`OpenEVSE_Lib` 0.0.22)
+
+Added `OpenEVSEClass::getRelayHealth(callback)`, `resetRelayHealth(callback)`,
+and `runStuckRelayRecovery(callback)` in `src/openevse.{h,cpp}`, mirroring
+the existing `getFrequency()` / `getRelayStatus()` / `resetFaultCounters()`
+D9 extensions (same `isD9Supported()` protocol gate, same token-parsing
+style). Added the `OPENEVSE_RELAY_HEALTH_NOT_AVAILABLE` (`0xffff`) sentinel
+for fields that aren't available yet (baseline not established) or at all
+(thermal fields need `TEMPERATURE_MONITORING` on the controller).
+`getRelayHealth()`'s callback took a breaking 9th → 10th argument
+(`stuck_relay_recovery_count`) between 0.0.21 and 0.0.22; older callbacks
+still work against a newer controller (the field defaults to 0 if the
+9-argument form is compiled), but a 10-argument callback against an older
+0.0.21-vintage controller response is a compile-time signature mismatch, not
+a silent bug.
+
+`getFrequency()` itself was **not** changed to expose the new `$GZ`
+zero-threshold field, to avoid a breaking signature change to an existing
+public method — see below for how this gateway firmware gets that value
+instead.
+
+## Gateway firmware (this repo, `relay_health` branch)
+
+`src/evse_monitor.{h,cpp}` — `EvseMonitor` gained:
+
+- **`readRelayHealth()`** — calls `OpenEVSE.getRelayHealth()`, caches all
+  nine fields, exposed via `isRelayHealthKnown()` +
+  `getRelay*()`/`isRelay*()` getters.
+- **`readFrequency()` rewritten** — reads `$GZ` raw via `EvseMonitor`'s own
+  `RapiSender*` instead of the library's single-field `getFrequency()`, so it
+  now also captures the zero-cross threshold from the same response instead
+  of round-tripping `$GZ` twice. Gated on `isD9Supported()` (as
+  `getFrequency()` itself is), so a pre-D9 controller isn't sent a `$GZ` it
+  can only NAK.
+- **`resetRelayHealth()`** — wraps `$FH`; re-reads `$GL` on success rather
+  than guessing what a freshly-reset accumulator looks like locally.
+- **`runStuckRelayRecovery()`** — wraps `$FK`; re-reads `$GL`/`$GR` on
+  success. Sets `_relay_recovery_in_flight` for the duration, which
+  `loop()` checks to skip its own periodic RAPI traffic (heartbeat pulse,
+  state/amp/temp/settings polls) — `$FK` can hold the RAPI queue for up to
+  ~30s, and without this guard everything `loop()` would otherwise enqueue
+  during that window gets `RAPI_RESPONSE_QUEUE_FULL`, heartbeat pulses
+  included, which risks tripping the controller's heartbeat-supervision
+  fallback current if enough consecutive pulses are dropped.
+- **Cache invalidation at `evseBoot()`** — `_relay_health_known` and
+  `_zero_cross_threshold_ma` are reset at the top of `evseBoot()` (which
+  runs on every controller connect, not just ESP32 power-on). Without this,
+  swapping in a controller that doesn't support `RELAY_HEALTH`/D9 — or is
+  simply older — would keep serving the *previous* controller's cached
+  values indefinitely, since nothing else clears them.
+
+**Known gap**: `runStuckRelayRecovery()` and `resetRelayHealth()` are both
+correctly wired C++ API now, but nothing in this repo calls them — the GUI's
+"Relay Replaced - Reset Health Values" button and any future recovery
+trigger go through the generic RAPI passthrough (`/r?json=1&rapi=$FH` /
+`$FK`) instead, bypassing this layer (and, for `$FK`, its queue guard)
+entirely. That's fine functionally — the passthrough works — but it means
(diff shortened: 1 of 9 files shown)


jeremypoulter:
Should be using the Platform IO registry so the depedancy management works managehttps://registry.platformio.org/libraries/openevse/OpenEVSE

RAR (commented):
Read through this one carefully — the direction is right, and it closes the gap I hit when I probed a 9.3.0 dev-tip controller: `$GW`/`$GL` exist on the controller with no path to the web UI at all.

I built it, since the description says it wasn't compiler-verified: **`openevse_wifi_v1` builds clean**, 76 s, no warnings in `evse_monitor.cpp` / `app_config.cpp` / `evse_man.h`. Flash 95.9% (1885897 / 1966080), RAM 23.8%. The 10-arg `getRelayHealth()` lambda matches the 0.0.22 signature exactly, and the `$GZ` token indices agree with the library's own parsing (`getToken(1)` is the first value).

Five things worth looking at, most significant first.

### 1. `readFrequency()` silently drops its D9 gate

The library's `getFrequency()` short-circuits on `!isD9Supported()` without touching the bus. The rewrite issues raw `$GZ` through `_sender` with no such check, so a pre-D9 controller — 8.2.3 is still `master` on `open_evse` — now gets a `$GZ` at boot *and* every 60 s, NAK'd every time. Cheap fix: early-return on `!_openevse.isD9Supported()` alongside the existing `!_sender` guard.

### 2. `runStuckRelayRecovery()` has no caller

It's plumbed `EvseMonitor` → `EvseManager` and stops there — no HTTP endpoint, no MQTT, nothing calls it. Either wire it up or leave it in the library until there's a consumer.

The mirror-image gap: `resetRelayHealth()` (`$FH`) is in 0.0.22 and gets no gateway exposure at all. That's the one an installer actually needs, since the health accumulator is meaningless after a physical relay swap until it's reset.

### 3. The 35 s `$FK` will stall the RAPI queue past the heartbeat interval

`RapiSender` runs one command in flight against a 10-deep queue (`RAPI_MAX_COMMANDS`), and `EvseMonitor::loop` enqueues at 1 Hz. Roughly 10 s into a recovery cycle the queue is full and everything after it gets `RAPI_RESPONSE_QUEUE_FULL` and is dropped — including `heartbeatPulse()`, which runs on a 5 s interval (`EVSE_HEATBEAT_INTERVAL`). Seven missed pulses trips the controller's heartbeat supervision to the 6 A limit (`EVSE_HEARTBEAT_CURRENT`).

It self-heals — the `_count % EVSE_MONITOR_STATE_TIME` branch re-enables `$SY` within 30 s — and `$FK` is NAK'd with an EV connected, so nothing charges through the window. But it should be understood before an endpoint gets pointed at it, and it's an argument for suppressing the state/heartbeat polling for the duration rather than letting the queue overflow.

### 4. Cached health is never invalidated

`_relay_health_known` only ever goes true. Swap in a controller without `RELAY_HEALTH`, or one that predates it, and `/config` keeps serving the previous unit's life percentage indefinitely — `readRelayHealth()` runs at `evseBoot()`, but the `FEATURE_NOT_SUPPORTED` return leaves the stale cache intact. Clearing the flag at the top of `evseBoot()` covers it. Same for `_zero_cross_threshold_ma`.

### 5. Most of this is telemetry in a config endpoint

`relay_life_pct`, `relay_cold_open_count`, `relay_elec_damage_x1e6`, the thermal index/baseline and the recovery counter all move over the life of the unit; everything else in `/config` is static. As written, the GUI has to re-poll `/config` to notice a drift warning, and none of it reaches the websocket — related to the eventing gaps in #1202. `zero_cross_threshold_ma` genuinely is configuration and belongs where it is; the rest reads like `/status` to me. Your call on where the line goes.

I checked `/config` capacity while I was in there, since this adds up to 11 keys: a live TFT unit serves 135 members and ~446 bytes of string pool against `JSON_OBJECT_SIZE(128) + 1024`, so there's room — but the headroom is thin enough now to be worth a comment on the constant.

### Minor

- The comment in `getSettingsFromEvse()` says the re-read happens "without adding RAPI traffic". It adds `$GZ` + `$GL` per 60 s cycle. Negligible, but not zero.
- The description is stale against the branch: it describes head `6856978` pinning `OpenEVSE_Lib` 0.0.21, but the head is `de6fe6a1` at 0.0.22.
- The `/config` JSON example in the description omits `relay_stuck_recovery_count`, which the code emits.

chris1howell:
Thank you RAR for the detailed review.


## 1. `readFrequency()` silently drops its D9 gate

**Fixed.** Restored the `!isD9Supported()` check the library's
`getFrequency()` had and the raw-`$GZ` rewrite lost:

```cpp
if(!_sender || !_openevse.isD9Supported()) {
  return;
}
```

A pre-D9 controller no longer gets sent a `$GZ` it can only NAK, at boot and
every ~60s settings-poll cycle.

## 2. `runStuckRelayRecovery()` has no caller / `resetRelayHealth()` has no gateway exposure

**Both fixed, and taken further.** Added `EvseMonitor`/`EvseManager::resetRelayHealth()`
(wraps `$FH`, re-reads `$GL` on success) — the gap you flagged as "the one
an installer actually needs" is closed.

For `runStuckRelayRecovery()`, we went with your other option and then some:
rather than leave it uncalled, we built the consumer. New typed HTTP
endpoints on the gateway:

| Endpoint | Calls |
|---|---|
| `GET /relay/reset` | `EvseManager::resetRelayHealth()` ($FH) |
| `GET /relay/recovery` | `EvseManager::runStuckRelayRecovery()` ($FK) |

Both use the deferred-response pattern already established for `/scan`
(WiFi scan) — the handler holds the `MongooseHttpServerResponseStream` open
and returns immediately; the actual HTTP response is sent from the async
RAPI callback. That matters specifically for `/relay/recovery`: the raw `/r`
passthrough's `sendCmdSync()` busy-loops the calling task until the RAPI
reply arrives, which would have frozen the *entire* Mongoose server —
HTTP, WebSocket, MQTT, everything — for the recovery's ~30s worst case, not
just the one request. The deferred pattern keeps the rest of the server
responsive while the recovery runs. Both endpoints carry the same
`actuatorMethodAllowed()` CSRF guard as `/restart`/`/apoff`/etc.

`gui-nightshift`'s Health tab now points both buttons at these instead of
raw RAPI passthrough — "Relay Replaced - Reset Health Values" (existing,
repointed) and a new "Run Stuck Relay Recovery" button. See point 3: this is
what makes the queue guard there actually take effect.

## 3. The 35s `$FK` will stall the RAPI queue past the heartbeat interval

**Fixed.** Added `_relay_recovery_in_flight`, checked at the top of
`EvseMonitor::loop()`:

```cpp
if(_relay_recovery_in_flight) {
  _count++;
  return EVSE_MONITOR_POLL_TIME;
}
```

Set around the `$FK` call in `runStuckRelayRecovery()`, cleared in its
callback. While a recovery is outstanding, `loop()` skips its own periodic
RAPI traffic entirely for that tick — heartbeat pulse, state/amp/temp/
settings polls — rather than letting them pile into the queue and get
`RAPI_RESPONSE_QUEUE_FULL`'d once it fills, heartbeat pulses included, which
is what risked tripping the controller's heartbeat-supervision fallback
current.

Set unconditionally rather than only on the "will actually run" path: an
EV-connected refusal NAKs almost immediately (the controller checks before
doing anything), so the pause costs nothing on that path and only matters
when the recovery genuinely runs.

This guard is only exercised when `EvseMonitor::runStuckRelayRecovery()` is
actually called — which, per point 2, is now the case: `/relay/recovery` is
what the GUI calls, so the fix isn't inert.

## 4. Cached health is never invalidated

**Fixed.** `evseBoot()` (runs on every controller connect, not just ESP32
power-on) now resets both at the top, before any RAPI round-trips:

```cpp
_relay_health_known = false;
_zero_cross_threshold_ma = OPENEVSE_RELAY_HEALTH_NOT_AVAILABLE;
```

A controller swap to one without `RELAY_HEALTH`/D9 no longer keeps serving
the previous unit's cached life percentage (or zero-cross threshold)
indefinitely.

## 5. Most of this is telemetry in a config endpoint

**Not resolved — ** Didn't move the drifting
fields (`relay_life_pct` and siblings) to `/status`. It's a real design
question, but also a real API change with `gui-nightshift` implications on
the other side (`HealthTab.svelte` currently reads all of it from
`config_store`) — moving it means updating both repos in lockstep.

Did the concrete, low-risk part: added a comment at the `/config`
`DynamicJsonDocument` allocation site (`web_server_config.cpp`) noting the
headroom is thinner now — measured against a live TFT unit's ~135 members,
the ~11 new keys fit, but it's worth rechecking before the next addition
rather than assuming.

## Minor

- **"without adding RAPI traffic"** — this was in `docs/relay_health.md`
  (the closest thing we have to a PR description), not a code comment.
  Corrected: the settings-poll cadence does add `$GZ` + `$GL` to that
  60s cycle, small but not zero.
- **Stale commit/version references** — `docs/relay_health.md` described
  head `6856978` pinning `OpenEVSE_Lib` 0.0.21; the actual head at review
  time was `de6fe6a1` at 0.0.22. Doc rewritten to match, and again since for
  everything in this response.
- **`/config` JSON example missing `relay_stuck_recovery_count`** — fixed
  in the same doc rewrite.

## Verification

`pio run -e openevse_wifi_v1` builds clean at every stage of this response:
after the five fixes (flash 96.0%, 1887925/1966080 bytes), after adding the
typed endpoints (96.1%, 1889937/1966080), and after merging upstream
`master`'s intervening changes into `relay_health` (96.3%, 1892813/1966080).
RAM steady at 23.8% throughout. `gui-nightshift`'s `http_endpoints` branch:
full test suite (958/958), production build, and a Playwright screenshot
against the mock dev server confirming both buttons render and fire the new
endpoints.

RAR (approved):

PR #1216, opened 2026-09-01
fix(eventlog): stop pilot-state flicker from flooding the event log

RAR:
## The bug

The event that drives event logging triggers on any change of `(evse_state, pilot_state, vflags)` (`evse_monitor.cpp:84`):

```cpp
if(_evse_state != evse_state || _pilot_state != pilot_state || _vflags != vflags)
```

but the record written to the log carries `evseState`, `evseFlags` and `pilot` — where `pilot` is the **advertised current in amps**, not the pilot state. The one input that can change is the one never written down, so a pilot-state flicker produces an entry identical to the one above it in every field a reader can see.

## Why it is not cosmetic

The log is capped at `EVENTLOG_MAX_ROTATE_COUNT` files of `EVENTLOG_ROTATE_SIZE` bytes — roughly 40–80 entries. Measured on a live unit with a real controller and a real car:

- **72 entries in 10 minutes of continuous charging**, one every 8.1 s
- of the 80 entries the log retained, **75 were the same charging row**; only 3 distinct states appeared in the entire file
- the retained history spanned 4h44m and recorded essentially nothing — session starts, stops and faults had been rotated out within minutes of a session beginning

So the failure mode is not "the History view is noisy", it is "the event log cannot retain events".

## The fix (commit 1)

Record `pilotState` in the entry, and drop an entry whose visible fields match the last one written until `EVENTLOG_REPEAT_INTERVAL` (300 s) has passed.

Two exclusion decisions carry the design, and both are deliberate:

- **`pilotState` is recorded but not compared.** It is worth seeing, but a flicker in it must not be the thing that keeps the entry — that is the bug.
- **Time, energy, elapsed and temperature are excluded** for the opposite reason: they change on nearly every entry, so including them would suppress nothing. The 300 s interval still leaves a periodic sample of them.

The event type is part of the key, so the first entry of any new state — a fault above all — is always written. Warnings are not exempt beyond that, and that was a correction made during bench testing: an earlier version of this patch exempted them outright, and a sustained GFCI fault with a flickering pilot promptly wrote **eight identical warning rows in eight seconds**, evicting the context needed to interpret them.

## Saying why an entry survived (commit 2)

Suppressing repeats leaves a second problem: the surviving entries still do not say what put them there. A reader has to diff a row against the one above it, which may have rotated out of the window or sit in a different block — and the two fields that move most often during a session, the pilot current and the status flags, are not shown in the History view at all.

The result is rows that still read as duplicates. From the live unit: two "Charging" rows a minute apart were the pilot stepping 47→42 then 42→41 A as the charge was throttled; two "Disabled - Sleeping" rows two seconds apart were the stop followed by the charging relay bit clearing in `evseFlags`. Four distinct events, rendered as two pairs of twins.

The repeat filter already computes this to decide what to drop, so recording it is close to free. Each entry carries the fields that differ from the last one written, stored as a bitmask and served from `/logs` as a list of names:

```json
{ "evseState": 2, "evseFlags": 1280, "pilotState": 2,
  "changed": ["state", "pilot"], "pilot": 47, ... }
```

`boot` means there was nothing to compare against; `periodic` means nothing visible moved and the entry is the interval sample of an unchanging state. `pilotState` is never listed — it is recorded, but a change in it alone does not create an entry, so naming it as a reason would be a lie.

## Compatibility

Additive only; no field removed or renamed. Entries written by older firmware read back as `pilotState: 255` (unknown) and `changed: []`, which was verified against a real four-month-old log on the device rather than in a fixture.

## Structure and testing

The decision lives in `src/event_log_repeat.h`, free of Arduino, LittleFS and ArduinoJson, following the existing `charge_threshold.h` / `fault_text.h` pattern in this repo. That makes it host-testable: 10 cases / 27 assertions covering the interval boundary, each visible field independently, repeated and distinct faults, a backwards clock jump (an NTP correction must not silence the log until real time catches up), the change mask, and the run-of-repeats case where a naive implementation keeps pushing the deadline out and an unchanging session never logs at all.

## Verification

| Check | Result |
|---|---|
| `native_test` | 12/12 suites |
| `openevse_wifi_v1` (core-2, 4 MB) | SUCCESS — 96.0% |
| `esp32-c3-devkitc-02` (RISC-V) | SUCCESS — 92.8% |
| `openevse_wifi_tft_v1` (core-3, arduino+espidf) | SUCCESS |
| `native_openevse` | SUCCESS |

Cost, A/B against master with an identical `BUILD_TAG`: **+976 bytes flash, +32 bytes RAM**. Still 96.0% on the 4 MB board.

**Bench-tested** on an `openevse_wifi_tft_v1` with an in-firmware fake controller driving a 1 Hz pilot-state flicker, against with-fix and reverted-fix builds of the same tree:

| Run | 150 s of flicker while charging | Entries written | Log rotations |
|---|---|---|---|
| without the fix | | ~75, all identical | **18** — the whole window recycled nearly twice |
| with the fix | | **0** | 0 |

Four further minutes spanning both charging and a sustained GFCI fault produced 5 entries total and zero rotations, with the fault logged exactly once and its recovery logged after.

**Hardware-validated** on a live unit (controller 9.3.0.SAMD, real vehicle): 13 minutes across a charge session and its stop produced 4 entries and zero rotations, against 72 entries per 10 minutes before. `changed` and `pilotState` confirmed live, alongside pre-existing entries correctly reading back as unknown.

## Note for GUI work

`/logs` now carries the reason, but no GUI renders it yet, so those rows still look identical on screen. Making them legible wants two things in the web UI: display the `changed` reason, and display the pilot current — it is in every record already and is the single most common reason a row exists, yet no view shows it.

### DIFF
--- a/api.yml
+++ b/api.yml
@@ -1104,6 +1104,9 @@ components:
             managerState: active
             evseState: 1
             evseFlags: 123
+            pilotState: 2
+            changed:
+              - pilot
             pilot: 32
             energy: 1234.5
             elapsed: 123456
@@ -1130,6 +1133,35 @@ components:
         evseFlags:
           type: integer
           format: int32
+        pilotState:
+          description: >-
+            J1772 pilot state at the time of the event. 255 for entries written
+            before this field was recorded.
+          type: integer
+          minimum: 0
+          maximum: 255
+        changed:
+          description: >-
+            Which fields differ from the previous entry, i.e. why this entry
+            exists. "boot" means there was no previous entry to compare
+            against; "periodic" means nothing visible changed and the entry is
+            the periodic sample of an unchanging state. Empty for entries
+            written before this field was recorded. The pilot state is never
+            listed: it is recorded, but a change in it alone does not create an
+            entry.
+          type: array
+          items:
+            type: string
+            enum:
+              - boot
+              - type
+              - manager
+              - state
+              - flags
+              - pilot
+              - divert
+              - shaper
+              - periodic
         pilot:
           type: integer
           format: int32
--- a/api.yml
+++ b/api.yml
@@ -1143,7 +1143,9 @@ components:
         changed:
           description: >-
             Which fields differ from the previous entry, i.e. why this entry
-            exists. "boot" means there was no previous entry to compare
+            exists. "flags" covers only the meaningful evseFlags bits: the
+            session-timing and previous-EV-connected bookkeeping bits are
+            recorded in evseFlags but never create an entry of their own. "boot" means there was no previous entry to compare
             against; "periodic" means nothing visible changed and the entry is
             the periodic sample of an unchanging state. Empty for entries
             written before this field was recorded. The pilot state is never
--- a/src/event_log.cpp
+++ b/src/event_log.cpp
@@ -77,7 +77,7 @@ void EventLog::begin()
   }
 }
 
-void EventLog::log(EventType type, EvseState managerState, uint8_t evseState, uint32_t evseFlags, uint32_t pilot, double energy, uint32_t elapsed, double temperature, double temperatureMax, uint8_t divertMode, uint8_t shaper)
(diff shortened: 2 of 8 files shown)


RAR:
commit two gui side: https://github.com/OpenEVSE/openevse-gui-nightshift/pull/106

chris1howell (approved):

PR #1219, opened 2026-09-03
loadsharing: gate periodic mDNS discovery on the enabled flag, fix the service name and IPv4 selection

RAR:
Stacked on #1218 (the leak fix); the diff shown here will collapse to three commits once that merges.

## 1. Only run periodic discovery while load sharing is enabled

`LoadSharingDiscoveryTask` queried `_openevse._tcp` every 10 s on every unit from boot, enabled or not. With load sharing off nothing consumes the results: the peer poller only polls joined peers and the group state only refreshes online flags on peers already added. Each query is still an mDNS round that every other OpenEVSE on the LAN answers, plus a burst of small allocations that fragments the heap on no-PSRAM boards.

Now: periodic queries run only while `loadsharing_enabled` is true. When off the task idles. `POST /loadsharing/discover` (what the peer-management UI uses to browse for chargers before enabling) still runs a single query through a one-shot trigger flag, so that flow is unchanged. Turning the setting on is picked up on the next 2 s wake.

## 2. Query `_openevse._tcp`, not `openevse.tcp`

Found while verifying (1) with a host-side mDNS listener. The task calls the raw IDF `mdns_query_async_new` with `"openevse"`/`"tcp"`, assuming the API adds the DNS-SD underscores. It does not: mdns.h documents the arguments as `_http`, `_tcp`, and only the Arduino ESPmDNS wrapper prepends them. On the wire every unit was asking for `openevse.tcp.local`, which nothing advertises, while the units themselves (`MDNS.addService("openevse", "tcp")`) answer `_openevse._tcp.local`. **Discovery never found a peer on real hardware.** The integration harness presumably shims mDNS without the underscore distinction.

## 3. Take the first IPv4 address, not the first address

Found once (2) made discovery return results. `r->addr` was read as IPv4 without checking the family. A unit with IPv6 answers with its A record plus link-local and ULA AAAA records, and the ULA came first, so the peer was recorded as `253.75.40.139` (the first four bytes of `fd4b:288b:…`). Walk the list and take the first `ESP_IPADDR_TYPE_V4` entry.

## Verification (bench `openevse_wifi_tft_v1`, host listener on 224.0.0.251:5353)

| state | queries from the bench |
|---|---|
| before, load sharing disabled | continuous, every ~2 s (one search per 10 s with the library's retransmits), name `openevse.tcp.local` |
| after, load sharing disabled, 45 s | 0 |
| after, one `POST /loadsharing/discover` | one search, name `_openevse._tcp.local` |

`GET /loadsharing/peers` after the trigger:

| build | discovered peers |
|---|---|
| master | local device only |
| after (2) | `openevse-2760` at `253.75.40.139`, `openevse-bench32` at `10.75.0.29` |
| after (3) | `openevse-2760` at `10.75.1.157`, `openevse-bench32` at `10.75.0.29` |

Heap stayed flat and the MQTT session stayed up throughout.

### DIFF
--- a/src/loadsharing_discovery_task.cpp
+++ b/src/loadsharing_discovery_task.cpp
@@ -83,6 +83,7 @@ LoadSharingDiscoveryTask::LoadSharingDiscoveryTask(unsigned long cacheTtl,
     _active_query(nullptr),
     _query_start_time(0),
     _query_in_progress(false),
+    _manual_trigger(false),
     _groupState(nullptr),
     _discovery_count(0),
     _last_result_count(0)
@@ -116,9 +117,16 @@ unsigned long LoadSharingDiscoveryTask::loop(MicroTasks::WakeReason reason) {
       }
     }
   } else {
-    // No query in progress, check if we should start a new one
-    if (now - _last_discovery_time >= _discovery_interval_ms || _last_discovery_time == 0) {
-      // Time to start a new discovery
+    // No query in progress, check if we should start a new one. Periodic
+    // discovery only runs while load sharing is enabled: with it off nothing
+    // consumes the results, and every query is an mDNS round that every other
+    // OpenEVSE on the LAN answers plus a burst of heap churn on no-PSRAM
+    // boards. A manual trigger still runs a single query so the UI can browse
+    // for peers before enabling.
+    bool periodic_due = loadsharing_enabled &&
+                        (now - _last_discovery_time >= _discovery_interval_ms || _last_discovery_time == 0);
+    if (_manual_trigger || periodic_due) {
+      _manual_trigger = false;
       DBUGF("LoadSharingDiscoveryTask: Starting discovery iteration %lu", _discovery_count + 1);
       startAsyncQuery();
       _last_discovery_time = now;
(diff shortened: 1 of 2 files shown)


src/loadsharing_discovery_task.cpp
```diff
@@ -12,6 +12,12 @@
 #include <ESPmDNS.h>
 #include <mdns.h>
 #include <algorithm>
+
+// The EpoxymDNS shim used by the native builds carries the esp_ip_addr_t
+// type field but not the IDF constant for it. IDF defines it as 0U.
+#ifndef ESP_IPADDR_TYPE_V4
+#define ESP_IPADDR_TYPE_V4 0U
```
jeremypoulter: Should be fixed in EpoxyDuino not here and also this issue implies a different behaviour in EpoxyDuino regarding adding `_` which also should be fixed

RAR: Agreed on both counts. jeremypoulter/EpoxyDuino#4 moves the constant into the shim and makes `mdns_query_async_new` take the labels verbatim like IDF, so a bare `"openevse"` now fails on the host the same way it fails on the device. The fallback define is dropped here in f03d1465. Since the native envs pull EpoxyDuino unpinned from `develop`, this PR's native build needs #4 merged first; I verified the combination locally (native_openevse builds, `test_peer_discovery_mdns` 3 passed).

jeremypoulter:
I have merged the EpoxyDuino PR

RAR:
> I have merged the EpoxyDuino PR

Should be good to merge.

chris1howell (approved):

Issue #1224, opened 2026-09-05
openevse_wifi_v1 (4MB) exceeds flash partition after ArduinoJson v7 migration


jeremypoulter:
## Summary

The `openevse_wifi_v1` PlatformIO env (4MB boards, 1,966,080-byte app partition) overflows its flash partition once ArduinoJson is bumped from v6.20.1 to v7.4.3 (#1044).

```
Error: The program size (2034601 bytes) is greater than maximum allowed (1966080 bytes)
RAM:   [==        ]  24.0% (used 78632 bytes from 327680 bytes)
Flash: [==========]  103.5% (used 2034601 bytes from 1966080 bytes)
```

Master is already at 99.9% (1,964,049 / 1,966,080 bytes, ~2KB spare) before this bump, and ArduinoJson v7 adds roughly 70KB of flash over v6 on this codebase, so there's no headroom left to absorb it.

## Context

Found while updating #1044 (ArduinoJson v6 -> v7 migration) to current master. Every other CI-built env I checked builds fine with the bump:

- `native_simulator` — OK
- `native_openevse` — OK
- `native_test` (unit tests) — OK, all pass
- `openevse_wifi_tft_v1_dev` (6.5MB partition) — OK, 44.8% flash used
- `openevse_wifi_v1` (4MB partition) — **overflow**

## Options to consider

- Trim flash usage elsewhere in the 4MB build (feature gating, `-Os`, dropping unused libs/strings, etc.)
- Move `openevse_wifi_v1` to a larger partition table if the board has the flash to support it
- Investigate whether ArduinoJson v7 can be configured (e.g. via `ARDUINOJSON_ENABLE_...` flags) to claw back some of the ~70KB delta

This should be resolved before #1044 can merge, since that PR is blocked on this env building within budget.

jeremypoulter:
## Flash review

Measured on `openevse_wifi_v1` at the #1044 merge commit. Partition is 1,966,080; the build is 2,034,585, so **68,505 over**. Master is 1,964,049 (2,031 spare).

### Where the flash goes

From `firmware.map` (`.flash.text` + `.flash.rodata` + `.iram0.text` + `.dram0.data`):

| item | bytes | % partition |
|---|---|---|
| **`web_server_static.cpp.o`** (embedded web GUI) | 289,371 | 14.7% |
| **`libMicroOcpp.a`** (+ `ocpp.cpp.o` 12,223) | 208,724 | 10.6% |
[...]

Issue #122, opened 2020-10-16
Base topic changes


robao23:
Hi guys.

I have a ghost in open evse I want to know if it is a reported error and if so, what solution does it have.

I put a base topic in services, base topic.

And after a few hours, a few days or a week I just change it.

Example:

246f2893d3dcOpenEVSE

Today I realized that the eco mode was loading at 16A and in open evse it was set this: 

246f2893d3dc2246f2893d3dcOpenEVSE

I am a newbie in this, these are the data that I see on the web, in case they are useful or if they need more information.

WiFi Firmware
ESP8266
Version: 3.1.0.dev

Greetings.

glynhudson:
Sorry to hear your having issues. Was the ESP32 module pre-loaded with firmware and purchased via the OpenEnergyMonitor store?

Try updating to the latest version via the web interface: https://github.com/OpenEVSE/ESP32_WiFi_V3.x/releases/tag/3.3.2

robao23:
> Sorry to hear your having issues. Was the ESP32 module pre-loaded with firmware and purchased via the OpenEnergyMonitor store?
> 
> Try updating to the latest version via the web interface: https://github.com/OpenEVSE/ESP32_WiFi_V3.x/releases/tag/3.3.2
thanks for answering


Yes, buy the advance kit in June at the official store.

I will try with that new firmware.


It is valid for my system, now it says this;

WiFi Firmware
ESP8266
Version: 3.1.0.

glynhudson:
Yes, you have an ES32 module so you can flash the latest V3 firmware. Current latest is V3.3.2 

https://github.com/openevse/ESP32_WiFi_V3.x/releases/

Update can be done via the web interface.

PR #1225, opened 2026-09-05
perf(web): re-encode embedded GUI assets to reclaim ~35KB of flash

jeremypoulter:
Reclaims 35,464 bytes of flash on the 4MB boards by re-encoding the embedded GUI assets. No behaviour change, no client change.

**This is now urgent rather than nice-to-have.** After the GUI bump to 2026.37.0 (#1223), `openevse_wifi_v1` on master measures **1,965,665 of 1,966,080 bytes — 100.0%, with 415 bytes spare.** The next commit that touches almost anything will fail the build. It also blocks #1044 (ArduinoJson v7), which needs ~70KB more than v6.

Groundwork for #1224.

## Why the assets

The embedded web GUI is the single biggest thing in the image — 14.7% of the app partition. I reviewed the content first and it is *not* where the waste is:

- All four locale bundles have identical, complete key sets (985 keys, zero missing or extra).
- The 191 keys with no literal reference in the app bundles are all runtime-built paths (`t('days.' + d)`, `dashboard.status.*`, `history.reason.*`) — not dead.
- Zero of 158 English strings >=25 chars appear in `index.js`/`vendor.js`/`charts.js`, so nothing is duplicated out of the locale files.
- Only 139-155 of 985 keys are byte-identical to English, and those are untranslatable tokens (`OpenEVSE`, `IP`, `MAC`, `kW`, `A`, `V`).

The waste is in **how the assets are encoded**, not what they contain.

## What this does

New `scripts/web_assets.py`, hooked into `binary_to_header()` so it runs whenever the checked-in headers are regenerated from a GUI dist:

| pass | saving |
|---|---|
| Text assets: decompress and re-pack with **zopfli** | −9,517 |
| PNG icons: quantize to a **256-entry palette** | −25,947 |

Zopfli emits an ordinary gzip stream, so nothing on the client or in `web_server_static.cpp` changes — it is simply better-packed deflate. The PNG icons ship as 8-bit RGBA but the artwork only uses 431-457 distinct colours, so 256 entries are visually indistinguishable (RMSE 0.0015-0.003); the 512px icons drop from ~15KB to ~5KB.

Each pass keeps its result only when it actually came out smaller, so a future GUI build that already ships optimal assets is a no-op rather than a regression. The ETag is taken over the bytes actually embedded, so re-encoding invalidates stale client caches.

## Result

| | flash | spare |
|---|---|---|
| master (cc434801) | 1,965,665 (**100.0%**) | **415** |
| **this PR** | **1,930,201 (98.2%)** | **35,879** |

## Brotli: wired up, deliberately not enabled

`StaticFile.compressed` (bool) becomes `encoding` (string) that `web_server_static.cpp` sends verbatim, and `WEB_ASSET_ENCODING=brotli` switches the compressor for a further ~26KB (measured 96.8% on the pre-bump GUI).

It is off by default because it needs verifying on hardware first. The server sends `Content-Encoding` unconditionally and never reads `Accept-Encoding`, so the encoding is fixed at generation time rather than negotiated, and there is no affordable runtime fallback:

- Decoding brotli on device would need RFC 7932's mandatory 122,784-byte static dictionary plus a decoder and ring buffer — ~150KB+ of flash to save 35KB. The ESP32 ROM exposes no inflate either.
- Storing both encodings costs +199,622 bytes.
- Storing raw and compressing per request costs +416,866 bytes plus a deflate compressor and its RAM.

Browsers historically only *advertised* `br` over HTTPS and this device is usually reached over plain HTTP on the LAN. They should still decode when explicitly told to, but that is unverified against real Chrome/Firefox/Safari, so flipping the switch should be a separate, tested change.

## CI

`web_assets.py` degrades gracefully when zopfli/brotli/Pillow are missing — it falls back to stdlib gzip and skips PNG optimisation. That would silently make CI binaries larger than local ones, so `scripts/requirements.txt` is added and installed in all three workflow setup steps.

## Testing

GUI rebuilt from source at the current submodule (2026.37.0) and all headers regenerated through the real pipeline. `native_simulator`, `native_openevse` and `openevse_wifi_v1` build; `native_test` passes.

Some `lcd_static` icons only regenerate on the TFT_eSPI envs, so those pick up the PNG pass whenever CI next builds them rather than in this commit.


RAR (commented):
The bytes check out: `openevse_wifi_v1` on the PR head builds at 1,930,285 (within 84 B of the claim), every gzip asset decompresses byte-identical to master, the zopfli output reproduces exactly with zopfli 0.4.1 and 0.4.3, the PNGs keep their alpha channel with RGBA RMSE 0.0016–0.0087, and the ETag follows the embedded bytes. `native_test` passes.

What I'd hold the merge on is how the headers get regenerated after this lands:

1. **`.github/workflows/build_gui.yml` (~70-83) deletes the `web_static` / `lcd_static` headers, runs `pio run`, and auto-commits "Update of pre-built GUI files", but it was not given the `pip install -r scripts/requirements.txt` step that `build.yaml` got in three places** (its pip cache key already hashes `requirements.txt`, so the intent was there). As written, the next GUI release regenerates the headers with stdlib gzip and unoptimised PNGs, commits a ~35 KB regression, and `openevse_wifi_v1` stops linking again.

2. **`scripts/extra_script.py:321-323` regenerates `lcd_static` on every build with no gate.** Reproduced: a plain `pio run -e openevse_wifi_v1` in a tree whose Python lacks Pillow (the PlatformIO penv, for instance) rewrote the three optimised `lcd_static` headers back to master's sizes. Because those headers are committed, "degrade gracefully when optimisers are missing" is the wrong default: a `git add -A` after that loses the saving silently. Regeneration should fail loudly instead. (#1226 removes `gui-tft` and with it this trigger, but not item 1.)

3. `scripts/web_assets.py` `_requantize_png` prefers `LIBIMAGEQUANT` when Pillow has it, which distro Pillow does and the PyPI wheels do not. On such a machine 5 of the 7 icons come out different and larger (logo 5109 vs 3937, button_bar 3095 vs 2529, wifi 1696 vs 1405, ...) with different ETags. The committed bytes reproduce with `FASTOCTREE`; pinning that makes the output deterministic.

Minor: the RMSE range in the description/docstring (0.0007–0.003) understates the lcd icons (0.005–0.009), and `requirements.txt` installs brotli in three CI jobs although brotli is deliberately disabled.

jeremypoulter:
@RAR I am actually wondering if this should go into the GUI build, what do you think

RAR:
Yes, I think it belongs in the GUI build. The firmware already embeds the gzip bytes the GUI produces (vite-plugin-compression2 with `deleteOriginalAssets`), so re-packing them in `extra_script.py` is one layer too late. compression2 2.5.x takes a custom compressor via `defineAlgorithm`, and `@gfx/zopfli` is a WASM build with no native deps, so it is a config change plus a pinned devDependency. The PNG icons are static files in `public/`, so quantize once with pngquant and commit them, no build step.

That also removes the failure modes I raised here: `build_gui.yml` auto-committing larger headers without the pip deps, the `lcd_static` rewrite on every developer build, and the Pillow quantizer nondeterminism. Keep the `encoding` plumbing and ETag change in the firmware so brotli stays a GUI-side switch later; the rest of the Python can go. Only cost is that the saving lands with the next GUI tag and pin bump rather than this PR.

jeremypoulter:
Followed up on this: opened two PRs that split it the way RAR suggested.

- **OpenEVSE/openevse-gui-nightshift#122** — the actual flash savings (zopfli gzip re-encoding + palette-quantized PNG icons), moved into the GUI build itself. Verified against a from-source build at the pinned `2026.37.0` tag: `dist/` drops from 287,850 to 253,564 bytes (−34,286), same ~35KB this PR measured, `npm test` green (135 files / 1002 tests).
- **#1238** — just the `StaticFile.compressed` (bool) → `encoding` (string) plumbing this PR added, so a GUI-side switch to a different `Content-Encoding` later (brotli) doesn't need a firmware change. Everything else here — `web_assets.py`, `scripts/requirements.txt`, the `build.yaml` pip installs — is dropped, since it's redundant once #122 lands and the submodule gets bumped.

Once #122 merges and the GUI tag/pin gets bumped, `build_gui.yml`'s normal regeneration picks up the smaller assets automatically — no firmware-side re-encoding step needed.

Given that, I'd suggest closing this one in favor of #1238 (and the pending GUI bump) — but that's your call as the author, not mine to make.

https://claude.ai/code/session_01UfjGrn2ZgomFGuQrPT4ogL

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1242, opened 2026-09-09
perf(build): disable C++ exceptions to reclaim ~155KB flash

jeremypoulter:
## Context

Follow-up from the flash-usage discussion on #1044 (ArduinoJson v6→v7 migration, #1224). While validating that PR I found that ~58% of the flash regression it triggers isn't ArduinoJson code at all — it's C++ exception unwind tables (`.eh_frame`), which grow with every function the linker keeps regardless of whether that function can throw.

## What's going on

The Arduino-ESP32 PlatformIO builder unconditionally adds `-fexceptions` to `CXXFLAGS`. I checked every library this firmware links against — ArduinoJson, ArduinoMongoose, MicroOcpp, ConfigJson, OpenEVSE, Adafruit GFX — and none of them `throw` or `catch`. The only real usage in the whole tree was one `try`/`catch` in `flash_migrate.cpp`, converting `std::vector::reserve()`'s `bad_alloc` into a clean failure.

So the project was paying for unwind tables on every function without using them anywhere. On `openevse_wifi_v1` that was **106,948 bytes of `.eh_frame`** (5.4% of the 4MB partition) baked into `.flash.rodata`.

## Change

- Add `-fno-exceptions` / `build_unflags = -fexceptions` via `[env]` + `[common].build_flags`, so it applies uniformly to every ESP32 hardware env without per-env special-casing (native/EpoxyDuino envs are unaffected — they build with host gcc, not this framework builder).
- Replace the one real `try`/`catch` in `flash_migrate.cpp` with a precheck against `ESP.getMaxAllocHeap()` — the actual constraint `reserve()` needs (one contiguous block) — before calling the same `reserve()`. Same failure behavior, no exceptions required.

## Numbers

| build | flash | of 1,966,080-byte partition |
|---|---|---|
| master (unchanged) | 1,962,721 | 99.8% |
| master + this change | 1,791,461 | **91.1%** (−171,260) |

The savings compound going forward: removing or simplifying any function elsewhere in the tree also removes its unwind entry, so future flash-reclaiming work gets roughly 3x the effective payoff it would show in a naive before/after byte count.

## Verification

- `native_test` — 13/13 suites pass
- `native_openevse` — builds
- `native_simulator` — builds
- `divert_sim` — all 81 scenarios pass
- `openevse_wifi_v1` — builds clean with zero warnings from project source, at the size above

`openevse_wifi_tft_v1_dev` (core-3.x/pioarduino platform) could not be built in my sandbox — its toolchain fetch hit a network/TLS issue reaching github.com directly, unrelated to this change — so CI should be the first real confirmation there. No library it pulls in (Adafruit GFX included) throws or catches either, and the flag reaches it the same way as every other env (no separate wiring needed).

Refs #1224.

https://claude.ai/code/session_01B3XmFqw1P36RNkaABQ1Uvq

---
_Generated by [Claude Code](https://claude.ai/code/session_01B3XmFqw1P36RNkaABQ1Uvq)_

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -560,6 +560,18 @@ build_flags =
   ; (measured: /override went from 0 to 4,096 bytes of largest-block drop).
   ; The handlers that emit large bodies serialise into a right-sized buffer
   ; and write once instead; see handleStatus and handleConfigGet.
+  ; This is a hybrid Arduino-as-ESP-IDF-component build (`framework = arduino,
+  ; espidf` above): the espidf managed component tree it pulls in includes
+  ; espressif__esp_modem, which genuinely uses C++ exceptions internally
+  ; (see exception_stub.hpp). [common].build_flags' -fno-exceptions (added
+  ; for the 4MB-partition flash budget, OpenEVSE/openevse_esp32_firmware#1224)
+  ; reaches this whole IDF component build, not just our own sources, so
+  ; esp_modem fails with "exception handling disabled". Reassert -fexceptions
+  ; here -- this board is on a 16MB partition with no flash-budget need for
+  ; the savings anyway -- and cancel [env]'s build_unflags so it isn't
+  ; immediately stripped back out.
+  -fexceptions
+build_unflags =
 lib_deps =
   ${common.lib_deps}
   ${common.gfx_display_libs}
@@ -581,6 +593,9 @@ build_flags =
   ${common.debug_flags}
   -D DEBUG_PORT=Serial
   -D RAPI_PORT=Serial2
+  -fexceptions ; see openevse_wifi_tft_v1: espidf's esp_modem needs it; this
+               ; env redefines build_flags wholesale so it doesn't inherit
+               ; the parent's copy of the flag, only its build_unflags override
 upload_protocol = custom
 upload_command = curl -F firmware=@$SOURCE http://$UPLOAD_PORT/update
 build_type = debug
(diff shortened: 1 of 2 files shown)


jeremypoulter:
CI is fully green on `274da06` — all 30 checks passing, including the three core-3/espidf hybrid envs (`openevse_wifi_tft_v1`, `openevse_wifi_tft_v1_dev`, `openevse_wifi_v1_16mb`) that the first push broke.

For the record, what happened: the initial `-fno-exceptions` push (`99d581b`) built cleanly on the classic Arduino-core envs but broke those three hybrid builds — their `espidf` managed-component tree pulls in `espressif__esp_modem`, which genuinely uses C++ exceptions (`exception_stub.hpp`), and `[common].build_flags`' `-fno-exceptions` was reaching that whole IDF component build. Fixed in `274da06` by reasserting `-fexceptions` (and cancelling `[env]`'s `build_unflags`) on just those three envs — all on 16MB+ partitions with no flash-budget need for the change anyway. Verified the underlying flag-ordering mechanics (GCC's last-toggle-wins, PlatformIO's per-env override replacing the parent's default) with a local toy build before trusting CI to confirm the real one.

No open review threads. This is ready for review.

---
_Generated by [Claude Code](https://claude.ai/code)_

jeremypoulter (commented):
Good catch — both this and the deeper TOCTOU race Copilot flagged separately were real. Fixed in `ea57cbc` by removing the dynamic allocation entirely instead of guarding it: the manifest is always ≤`MIGRATE_SCRATCH_CAP` (`consume_chunk()` already rejects anything larger), so `mctx.buf` never needed to be resizable. Replaced `std::vector<uint8_t>` with a fixed `uint8_t[MIGRATE_SCRATCH_CAP]` array + length counter — no allocation, so no fragmentation risk, no race, and no false rejection on retry.

---
_Generated by [Claude Code](https://claude.ai/code)_

chris1howell (approved):

PR #1252, opened 2026-09-12
fix: resolve the MQTT broker and NTP host IPs without blocking loopTask

RAR:
## The bug

Two status fields were filled with `getaddrinfo()`, which blocks for the whole query. Both run on `loopTask`, which feeds the task watchdog:

- `mqtt_broker_ip`, in `Mqtt::loop()`
- the NTP `resolved_ip`, at three sites in `TimeManager`

With `LWIP_IPV6` enabled an `AF_UNSPEC` lookup asks for both A and AAAA, each up to `DNS_MAX_RETRIES` (4) times across every configured server — so a slow or unanswered resolve runs well past the 5 s `CONFIG_ESP_TASK_WDT_TIMEOUT_S` and panics the charger.

Both were introduced on the same day, in `5a387c75` (MQTT) and `3e92a94b` (NTP).

## How it was found

**MQTT.** A unit that had been up 25 h rebooted with `reset_reason_name = task_wdt` and panic `loopTask (CPU 1)`. The `loopTask` stack from the coredump:

```
loopTask -> app_main -> loop() -> MicroTasks dispatch -> Mqtt::loop()
  -> lwip_getaddrinfo (netdb.c:505) -> lwip_netconn_do_gethostbyname
  -> sys_mbox_new -> xQueueSemaphoreTake
```

The broker hostname was a CNAME to AWS IoT that answers AAAA first, which is what made the `AF_UNSPEC` path expensive on that network. `_needsDnsLookup` is re-armed in `onMqttConnect()`, so every MQTT reconnect was another chance to reboot.

**NTP.** The same panic then turned up on an ESP32-S3 **with MQTT disabled**, so it could not be the same path. Its `loopTask` stack:

```
loopTask -> mg_sntp_handler -> MongooseSntpClient::eventHandler
  -> lwip_netconn_do_gethostbyname (api_msg.c:2231)
  -> lwip_netconn_do_dns_found -> dns_timeout_cb (dns.c:444)
```

`MongooseSntpClient::eventHandler` there is the `MG_SNTP_FAILED` dispatch, which reaches `_sntp.onError()`. That handler runs *because* the hostname did not resolve, and then resolves the same name synchronously — so it always takes the full retry path. `dns_timeout_cb` on the stack is that path in progress.

The "early DNS probe" site has the same exposure. Its comment assumes Mongoose has already warmed LwIP's cache, which holds only when Mongoose's own resolve succeeded; when it did not, the cache is empty and the probe blocks.

## The fix

Ask LwIP for the address and let it call back, in both files:

- `dns_gethostbyname()` answers immediately (`ERR_OK`) when the name is in the resolver cache.
- Otherwise it returns `ERR_INPROGRESS` and calls a callback on the TCP/IP thread. The callback formats the address and hands it over with a release store; `loop()` collects it with an acquire load on a later pass.
- Any other return means the query could not be started, which reports `"failed"` as before.

Nothing waits. The status fields behave as before except that on a cache miss the value appears a moment later instead of stalling the loop until it is known. The two NTP sites that run after a *successful* resolve hit the cache and so are unchanged in practice; they go through the same path to keep the blocking resolvers out of the file entirely.

`dns_gethostbyname()` is LwIP raw API, so it is called under `LOCK_TCPIP_CORE()`. `CONFIG_LWIP_TCPIP_CORE_LOCKING` and `CONFIG_LWIP_CHECK_THREAD_SAFETY` are both enabled in the prebuilt Arduino framework for every ESP32 target and in `sdkconfig.defaults`, so the locking is available and runtime-asserted.

`TimeManager::loop()` additionally returns at most 250 ms while a lookup is outstanding, so an answer is not left sitting through an NTP retry back-off of up to 30 minutes. A 30 s deadline stops a lost callback from holding that rate indefinitely.

The comment the MQTT change replaces said `loop()` was safe to block in. Moving that resolve off the Mongoose callback fixed a dropped-connection bug, but traded it for a watchdog reboot — neither context can afford to wait.

The host (`EPOXY_DUINO`) build keeps the inline `getaddrinfo()`: no LwIP resolver, and no watchdog to trip.

## Testing

- `openevse_wifi_v1` builds clean — 91.1% flash, 25.3% RAM, unchanged by either commit.
- `native_openevse` (EpoxyDuino) builds clean, so the host paths still compile.
- `pio test -e native_test`: all suites pass.
- The MQTT half is running on an ESP32 TFT unit: MQTT reconnects and `mqtt_broker_ip` populates with no watchdog reboot and no loop stall.
- The NTP half is not yet hardware-soaked against a deliberately unresponsive DNS server. Review of the async handoff is welcome.

## Summary by CodeRabbit

* **Performance**
  * Improved MQTT broker and NTP server hostname resolution to avoid blocking the application loop.
  * MQTT communication and time synchronization can continue while DNS resolution is in progress.
* **Reliability**
  * Added handling for DNS lookup timeouts and stale results during connection or synchronization retries.
* **Compatibility**
  * Preserved synchronous hostname resolution for desktop host builds.

### DIFF
--- a/src/mqtt.cpp
+++ b/src/mqtt.cpp
@@ -8,6 +8,8 @@
 #include <netdb.h>
 #else
 #include <lwip/netdb.h>
+#include <lwip/dns.h>
+#include <lwip/tcpip.h>
 #endif
 #include "openevse.h"
 #include "divert.h"
@@ -29,6 +31,7 @@ Mqtt::Mqtt(EvseManager &evseManager) :
   _lastRxTime(0)
 {
   _brokerIp[0]      = '\0';
+  _dnsResult[0]     = '\0';
   _brokerVersion[0] = '\0';
   _errorCategory[0] = '\0';
   _errorDetail[0]   = '\0';
@@ -127,14 +130,14 @@ unsigned long Mqtt::loop(MicroTasks::WakeReason reason) {
     }
   }
 
-  // If connected, perform periodic checks and safe deferred DNS lookup
+  // If connected, perform periodic checks and the deferred broker IP lookup
   if (_mqttclient.connected()) {
     if (millis() - _loop_timer > MQTT_LOOP_INTERVAL) {
       _loop_timer = millis();
       checkAndPublishUpdates();
     }
 
-    // DNS lookup deferred from onMqttConnect (safe to block here, not in callback)
+    // Broker IP lookup, deferred from onMqttConnect and started without blocking
     if (_needsDnsLookup) {
       _needsDnsLookup = false;
       struct in_addr addr4; struct in6_addr addr6;
@@ -142,27 +145,15 @@ unsigned long Mqtt::loop(MicroTasks::WakeReason reason) {
                   (inet_pton(AF_INET6, mqtt_server.c_str(), &addr6) == 1);
       if (isIp) {
         strncpy(_brokerIp, mqtt_server.c_str(), sizeof(_brokerIp) - 1);
-      } else if (mqtt_server.length() > 0) {
-        struct addrinfo hints = {}, *res = nullptr;
-        hints.ai_family = AF_UNSPEC;
-        if (getaddrinfo(mqtt_server.c_str(), nullptr, &hints, &res) == 0 && res) {
-          void *addr = res->ai_family == AF_INET
-            ? (void *)&((struct sockaddr_in  *)res->ai_addr)->sin_addr
-            : (void *)&((struct sockaddr_in6 *)res->ai_addr)->sin6_addr;
-          inet_ntop(res->ai_family, addr, _brokerIp, sizeof(_brokerIp) - 1);
-          freeaddrinfo(res);
-        } else {
-          strncpy(_brokerIp, "failed", sizeof(_brokerIp) - 1);
-        }
         _brokerIp[sizeof(_brokerIp) - 1] = '\0';
+      } else if (mqtt_server.length() > 0) {
+        startDnsLookup();
       }
-      if (_brokerIp[0] != '\0') {
-        // WebSocket only — this is a UI status field, not broker data
-        StaticJsonDocument<128> dns_event;
-        dns_event["mqtt_broker_ip"] = _brokerIp;
-        web_server_event(dns_event);
-      }
+      publishBrokerIp();
     }
+
+    // An asynchronous lookup may land on any later pass
+    takeDnsResult();
   }
 
   // Periodic status push so GUI always reflects the real connection state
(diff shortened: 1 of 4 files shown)


jeremypoulter:
Do we really need the MQTT broker IP? This is a HUGE amount of code for something that is for information. If it is really needed we should probably get it from Mongoose. Almost certainly it will be wrong for the use case of AWS as every time you resolve the IP you will get a different value and the chances of getting the same one are slim.

RAR:
> Do we really need the MQTT broker IP? This is a HUGE amount of code for something that is for information. If it is really needed we should probably get it from Mongoose. Almost certainly it will be wrong for the use case of AWS as every time you resolve the IP you will get a different value and the chances of getting the same one are slim.

Yeah, I think it was trying to avoid touching your library.  I can re-focus it.

RAR:
Screenshot 2026-09-12 at 3 14 18 PM
Screenshot 2026-09-12 at 3 14 03 PM

I Believe this is what its used for...

jeremypoulter:
Yeah, I think this is only of value if it gets the IP Mongoose uses, particularly for AWS the IP Mongoose uses can definitely be different to what the firmware would resolve. I would either remove this feature or extend Mongoose to be able to get the IP. Fairly sure it is in the connection so should be easy. That being said I would really like to not don't for v6 and get v7 merged as it is so close.

RAR:
Closing in favour of a better fix.

This PR resolved the hostnames asynchronously so the lookups could no longer block `loopTask`. That works, and it needed no library change — but it kept ~275 lines of concurrency machinery (two TCP/IP-thread callbacks, a release/acquire handoff, a collection poll with a lost-callback deadline, a host/target `#ifdef` split) to answer a question Mongoose could already answer: it resolved the peer to open the connection and stores the address on it.

Replaced by:

- **jeremypoulter/ArduinoMongoose#102** — `remoteAddress()` on the MQTT and SNTP clients (mongoose 6 / `master`)
- **jeremypoulter/ArduinoMongoose#103** — the same on `MongooseSocket` for `mongoose_7`, which also fixes a null dereference in the existing `getRemoteAddress()`
- **#1253** — the firmware change, which reads the address instead of looking it up

That comes out at **68 insertions and 98 deletions against master**: the fixed firmware is smaller than the broken one, and it is more accurate on both paths — an empty peer address is an exact "this name did not resolve" signal, and on success the address reported is the one the reply actually came from rather than whatever a fresh lookup of a pool hostname returns.

The panics and the diagnosis are unchanged and are written up in full on #1253. Note that #1253 cannot merge until #102 lands and the `ArduinoMongoose` pin in `platformio.ini` moves past it, so if the reboot needs fixing before then, this branch is still here and still works.

RAR:
> Yeah, I think this is only of value if it gets the IP Mongoose uses, particularly for AWS the IP Mongoose uses can definitely be different to what the firmware would resolve. I would either remove this feature or extend Mongoose to be able to get the IP. Fairly sure it is in the connection so should be easy. That being said I would really like to not don't for v6 and get v7 merged as it is so close.

Honestly its close enough I'd merge it and we can work out anything remaining on master.

PR #1254, opened 2026-09-12
Add 2-line LCD backlight type (mono/RGB) control, bump OpenEVSE_Lib t…

chris1howell:
# 2-line LCD backlight type (Monochrome / RGB)

Adds a web UI control for the OpenEVSE controller's 2-line character LCD
backlight type, on branch `RGB_LCD` across three repos. Previously this could
only be changed over RAPI directly (Blocked by OpenEVSE_esp32_firmware) or from the controller's own front-panel
menu; there was no way to set it from the app.

## Background

The Adafruit RGBLCD character display used on classic OpenEVSE V6 boards (and
in JuiceBox v2 conversions, which reuse the JuiceBox's RGB-backlit LED Strip) can
run in one of two backlight modes:

- **Monochrome** — single-colour backlight.
- **RGB** — full colour, used for the state-colour backlight (green/yellow/
  teal/red/violet, see `docs/lcd.md`) and required for a JuiceBox v2
  replacement.

The controller firmware already exposed this as `$S0 0|1` (RAPI) /
`ECF_MONO_LCD` (EEPROM flags bit `0x0100`), gated behind the `LCD16X2` +
`RGBLCD` compile flags — only the `m328p_LCD_WIFI` build defines both;
`m328p_core` ("WiFi, no text LCD") defines neither, so `$S0` NAKs there. No
change was needed on the controller firmware (`open_evse`) side. `OpenEVSE_Lib`
could already *read* the flag but had no method to *set* it — that was the
actual gap this closes.

### OpenEVSE_Lib (branch `OpenEVSE9`, commit `8e93b3a`)

- `OPENEVSE_LCD_TYPE_MONO` / `OPENEVSE_LCD_TYPE_RGB` constants
  (`src/openevse.h`).
- `OpenEVSEClass::setLcdType(uint8_t type, callback)` (`src/openevse.cpp`) —
  sends `$S0 0|1`, following the same validate-then-`sendCmd` shape as
  `setServiceLevel()`/`feature()`.
- Bumped `library.json` to `0.0.24`.

### openevse_esp32_firmware (branch `RGB_LCD`, commit `b572f4c5`)

- `EvseMonitor::setLcdType()` (`src/evse_monitor.h`/`.cpp`) — calls
  `OpenEVSEClass::setLcdType()`, then re-reads `$GE` to refresh the cached
  flags word and fires `_settings_changed`, matching the pattern used by
  `enableDiodeCheck()` and friends.
- `EvseManager::setLcdType()` (`src/evse_man.h`) — thin pass-through, next to
  the existing `getLcdType()`.
- `app_config.cpp` — `lcd_type` (`"mono"` / `"rgb"`) added to `GET /config`
  and handled in the `POST /config` deserializer, unconditionally (like
  `diode_check`, `boot_lock`, etc. — there is no RAPI capability bit for
  "has a physical LCD", so unsupported builds just NAK the `$S0` silently).
- `platformio.ini` — `openevse/OpenEVSE` pin bumped `0.0.23` → `0.0.24` (both
  the main `[common]` lib_deps and the `native_simulator` env).

### gui-nightshift (branch `RGB_LCD`, commit `a2ef421`)

- Display settings page (`src/routes/settings/Display.svelte`) — new "2 Line
  LCD" field, a Monochrome/"Red - Green - Blue" segmented control (same
  pattern as the existing Theme/Clock controls), gated on `lcd_type` being
  present in the config store. Defaults to RGB when absent.
- Description note: "JuiceBox v2 Replacement select Red - Green - Blue."
- i18n keys added to `en.json` and all three translated catalogs
  (`source/es.json`, `source/fr.json`, `source/hu.json`) to satisfy the
  locale-parity test.
- Tests added to `Display.test.js` covering the default-RGB state and the
  write-on-change behaviour.

## Summary by CodeRabbit

- **New Features**
  - Added support for configuring the LCD display as monochrome or RGB.
  - LCD type settings are saved and restored when supported by the controller.
  - Configuration reports the selected LCD type when supported.
- **Bug Fixes**
  - Improved compatibility with controllers that do not support LCD type selection by avoiding unsupported configuration updates.
- **Chores**
  - Updated the OpenEVSE platform package dependency to version 0.0.24.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -38,7 +38,7 @@ lib_deps =
   https://github.com/jeremypoulter/ArduinoMongoose.git#ca0a817de014cf4ab1833b32383532366fca10d7  ; master: mbedTLS-3 port + unique SUBSCRIBE ids + TLS teardown UAF fix (97ff3b4)
   jeremypoulter/Micro Debug@0.0.5
   jeremypoulter/ConfigJson@0.0.6
-  openevse/OpenEVSE@0.0.23  ; PlatformIO registry package (was a git+tag pin) - D9 command support (protocol-gated) + relay-health $GL/$FH/$FK
+  openevse/OpenEVSE@0.0.24  ; PlatformIO registry package (was a git+tag pin) - D9 command support (protocol-gated) + relay-health $GL/$FH/$FK, 2-line LCD backlight type setter
   jeremypoulter/ESPAL@0.0.7
   jeremypoulter/StreamSpy@0.0.2
   jeremypoulter/MicroTasks@0.0.4
@@ -760,7 +760,7 @@ lib_deps =
   bblanchon/ArduinoJson@6.20.1
   jeremypoulter/Micro Debug@0.0.5
   jeremypoulter/ConfigJson@0.0.6
-  openevse/OpenEVSE@0.0.23
+  openevse/OpenEVSE@0.0.24
   jeremypoulter/ESPAL@0.0.7
   jeremypoulter/StreamSpy@0.0.2
   jeremypoulter/MicroTasks@0.0.4
--- a/src/app_config.cpp
+++ b/src/app_config.cpp
@@ -717,13 +717,17 @@ bool config_deserialize(DynamicJsonDocument &doc)
 
   if(doc.containsKey("lcd_type"))
   {
-    const char *val = doc["lcd_type"];
-    EvseMonitor::LcdType type = (val && strcmp(val, "mono") == 0) ?
-      EvseMonitor::LcdType::Mono : EvseMonitor::LcdType::RGB;
-    if(type != evse.getLcdType()) {
-      evse.setLcdType(type);
-      config_modified = true;
-      DBUGLN("lcd_type changed");
+    if(doc["lcd_type"].is<const char *>()) {
+      const char *val = doc["lcd_type"];
+      if(val && (strcmp(val, "mono") == 0 || strcmp(val, "rgb") == 0)) {
+        EvseMonitor::LcdType type = (strcmp(val, "mono") == 0) ?
+          EvseMonitor::LcdType::Mono : EvseMonitor::LcdType::RGB;
+        if(type != evse.getLcdType()) {
+          evse.setLcdType(type);
+          config_modified = true;
+          DBUGLN("lcd_type changed");
+        }
+      }
     }
   }
 
@@ -1035,6 +1039,5 @@ void config_reset()
   LittleFS.format();
   config_load_settings();
 }
-
 
 
--- a/src/app_config.cpp
+++ b/src/app_config.cpp
@@ -715,15 +715,29 @@ bool config_deserialize(DynamicJsonDocument &doc)
     }
   }
 
-  if(doc.containsKey("lcd_type"))
+  // Skipped entirely once a $S0 write has actually been rejected with $NK:
+  // that means this controller build doesn't have LCD16X2+RGBLCD compiled
+  // in, getLcdType() can never change no matter what's requested, and
+  // retrying on every POST would just re-trigger a config-change
+  // notification for a write that can't take. See the comment on
+  // EvseMonitor::_lcd_type_supported.
+  if(doc.containsKey("lcd_type") && evse.isLcdTypeSupported())
   {
     const char *val = doc["lcd_type"];
-    EvseMonitor::LcdType type = (val && strcmp(val, "mono") == 0) ?
-      EvseMonitor::LcdType::Mono : EvseMonitor::LcdType::RGB;
-    if(type != evse.getLcdType()) {
-      evse.setLcdType(type);
-      config_modified = true;
-      DBUGLN("lcd_type changed");
+    // ArduinoJson hands back nullptr for a non-string value, so this also
+    // covers {"lcd_type": true}/{"lcd_type": 0} etc. Anything other than
+    // exactly "mono" or "rgb" is ignored rather than silently treated as
+    // RGB - there's no existing precedent for a string-valued EVSE setting
+    // here to inherit a looser convention from.
+    bool isMono = val && 0 == strcmp(val, "mono");
+    bool isRgb  = val && 0 == strcmp(val, "rgb");
+    if(isMono || isRgb) {
+      EvseMonitor::LcdType type = isMono ? EvseMonitor::LcdType::Mono : EvseMonitor::LcdType::RGB;
+      if(type != evse.getLcdType()) {
+        evse.setLcdType(type);
+        config_modified = true;
+        DBUGLN("lcd_type changed");
+      }
     }
   }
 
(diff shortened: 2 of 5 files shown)


RAR:
Reviewed. Small and mostly clean; one real bug and a few things worth a second look.

## 1. `setLcdType()` drops `_settings_known = true`

`EvseMonitor::setLcdType()` is a verbatim copy of `enableFeature()`'s body with `_openevse.feature()` swapped for `_openevse.setLcdType()` — minus one line:

```cpp
if(RAPI_RESPONSE_OK == ret) {
  DBUGF("pilot = %ld, flags = %x", pilot, flags);
  _settings_flags = flags;
  _settings_known = true;     // <-- present in enableFeature(), missing here
}
```

That flag is load-bearing. `evseBoot()` resets it on controller change specifically so `getSettingsFlags()` cannot keep vouching for the previous controller, and the comment there notes consumers gate real decisions on it (Notifications will not prune persisted acks until it is true). On the path where `setLcdType()` is the first successful `$GE` since a controller change, `_settings_flags` ends up fresh while `isSettingsKnown()` still reports false.

The fix I would suggest is not to add the line back, but to stop copying the block: factor the refresh half of `enableFeature()` into a small helper and have both call it. The copy diverged on its first commit, which is the argument for not having one.

## 2. The GUI's capability gate cannot fire

`doc["lcd_type"]` sits inside `if(config_ver > 0)`, so it is emitted for every controller. The PR description says the GUI control is "gated on `lcd_type` being present in the config store" — that condition is always true.

So on `m328p_core` (the "WiFi, no text LCD" build, which defines neither `LCD16X2` nor `RGBLCD`) the app shows a live control whose `$S0` silently NAKs. And `getLcdType()` derives from `ECF_MONO_LCD`, a flags bit that means nothing on a build without the LCD compiled in, so such a device reports `"rgb"` — a backlight type for a display it does not have.

## 3. That combination does not converge

`if(type != evse.getLcdType())` never becomes false on such a controller, because the flags word never changes. Every `POST /config` carrying `lcd_type: "mono"` therefore re-sends `$S0`, sets `config_modified`, flushes the RAPI command queue and calls `increment_config()` — a config-change notification to every connected client, on every settings save, for a change that cannot take.

## 4. Anything that is not exactly `"mono"` becomes RGB

```cpp
const char *val = doc["lcd_type"];
EvseMonitor::LcdType type = (val && strcmp(val, "mono") == 0) ?
  EvseMonitor::LcdType::Mono : EvseMonitor::LcdType::RGB;
```

ArduinoJson returns `nullptr` from `as<const char *>()` for a non-string, so `{"lcd_type": true}` or `{"lcd_type": 0}` silently selects RGB rather than being rejected — as do `"Mono"`, `"MONO"`, `"monochrome"` and `""`. A typo flips a Mono unit to RGB instead of erroring.

This is the first string-valued EVSE setting in `config_deserialize()`, so there is no existing precedent to inherit here. Validating against both names and ignoring anything else would be closer to how the neighbouring booleans behave.

## Flash

`openevse_wifi_v1` built with the tracked `src/web_static` (no GUI submodule, the same fallback CI takes):

| | bytes |
|---|---|
| master | 1,791,161 |
| this PR | 1,792,181 |
| | **+1,020** |

## Merge note

This and #1233 both edit the `openevse/OpenEVSE` pin line in `platformio.ini`, so whichever lands second needs a rebase. Trivial to resolve — take `0.0.24` and merge both trailing comment fragments; I checked that `0.0.24` still carries `getCableTemperatures()` alongside the new `setLcdType()`.

RAR (dismissed):

chris1howell:
# RGB_LCD review response

Response to review of the 2-line LCD backlight type (`lcd_type`) feature on
branch `RGB_LCD`. Four findings were fixed; suggested remedy was worth doing and
has been applied.

| # | Finding | Disposition |
|---|---|---|
| 1 | `setLcdType()` drops `_settings_known = true` |Suggested refactor applied. |
| 2 | GUI capability gate cannot fire (`lcd_type` always emitted) | **Fixed** |
| 3 | `$S0` never converges → resend + notification on every POST | **Fixed** |
| 4 | Anything not exactly `"mono"` becomes RGB | **Fixed** |

Commits: `be22df74` (findings 2–4), `665cb63b` (merge of `origin/RGB_LCD`),
`7a3ada00` (finding 1's refactor).

---

## 1. `_settings_known` 


### What was done

The "stop copying the block" — is independently
valid: the 12-line refresh block genuinely was duplicated verbatim between the
two functions, which invites exactly the drift described. It is now `EvseMonitor::refreshSettingsFlags()`, called from both
sites (`7a3ada00`).

The extraction is mechanical and behaviour-preserving: same `$GE` call, same
*unconditional* `_settings_changed.Trigger()`, same callback semantics —
including that the callback still receives the **`$GE`'s** result rather than
the preceding write's. That last point is pre-existing behaviour shared by
every feature toggle and was deliberately not changed here.

Three other `$GE` readers were checked and deliberately left alone, because
their differences are load-bearing rather than drift:

| Caller | Why it differs |
|---|---|
| `evseBoot()` | Signals `_boot_ready` instead of triggering `_settings_changed` |
| `setServiceLevel()` | Chains a `getCurrentCapacity()` off the same response |
| `getSettingsFromEvse()` | Triggers **only on an actual change** — it runs once a minute, and firing every poll would push a needless event to every WebSocket and MQTT subscriber |

---

## 2 & 3. Capability gate and non-convergence — fixed

Both stem from one root cause, so they share a fix.

`doc["lcd_type"]` sat inside `if(config_ver > 0)`, so it was emitted — and on
`POST`, accepted — for every controller. On a build without `LCD16X2` +
`RGBLCD` (e.g. `m328p_core`, "WiFi, no text LCD") `$S0` is not compiled in, so
the controller's RAPI handler falls through to its `rc = -1` default and
answers `$NK`. `getLcdType()` therefore never changes, `if(type != evse.getLcdType())`
never becomes false, and every `POST /config` carrying `lcd_type` re-sent `$S0`,
set `config_modified`, and fired a config-change notification to every
connected client — for a write that cannot take.

RAPI has no capability query for this. `LCD16X2`/`RGBLCD` are compile-time
flags on the controller with no wire-visible signal either way; `$GV`'s own
documentation says *"protocol_version is deprecated … test commands for
compatibility, instead."* So support genuinely cannot be known without trying.

**Fix:** `EvseMonitor::_lcd_type_supported`, optimistic (`true`) at
construction, latching `false` the first time a `$S0` write is rejected with
`RAPI_RESPONSE_NK` specifically — *not* on `RAPI_RESPONSE_TIMEOUT`,
`QUEUE_FULL`, `NOT_CONNECTED` or `BUFFER_OVERFLOW`, none of which say anything
about whether the command exists. Exposed as `isLcdTypeSupported()` on both
`EvseMonitor` and `EvseManager`.

`app_config.cpp` now gates both directions on it:

```cpp
// GET — stop advertising a control that can never take effect here
if(evse.isLcdTypeSupported()) {
  doc["lcd_type"] = (EvseMonitor::LcdType::Mono == evse.getLcdType()) ? "mono" : "rgb";
}

// POST — stop retrying, and stop re-notifying, once we know it NAKs
if(doc.containsKey("lcd_type") && evse.isLcdTypeSupported())
```

The GUI needs no change: `Display.svelte` already gates on
`'lcd_type' in cfg`, so the control disappears on its own once the key stops
being emitted.

### Residual limitation

Before the first write attempt in a session there is no way to distinguish
"supported, not yet tried" from "unsupported". On unsupported hardware the
control therefore still appears once, and self-hides after that first `$NK`.
Closing that fully needs a controller-side capability signal, which is a
protocol change across two repos.

---

## 4. Input validation — fixed

```cpp
const char *val = doc["lcd_type"];
EvseMonitor::LcdType type = (val && strcmp(val, "mono") == 0) ?
  EvseMonitor::LcdType::Mono : EvseMonitor::LcdType::RGB;
```

Confirmed: ArduinoJson yields `nullptr` for a non-string, so `{"lcd_type": true}`
and `{"lcd_type": 0}` fell to the RGB branch, as did `"Mono"`, `"MONO"`,
`"monochrome"` and `""`. A typo flipped a Mono unit to RGB instead of erroring.

Now only exactly `"mono"` or `"rgb"` is acted on; anything else is ignored:

```cpp
bool isMono = val && 0 == strcmp(val, "mono");
bool isRgb  = val && 0 == strcmp(val, "rgb");
if(isMono || isRgb) { … }
```

**Note on overlap:** `origin/RGB_LCD` independently received
`2ae76efb fix(config): ignore invalid LCD type values` (coderabbitai) fixing
this same finding while the above was in progress. The merge (`665cb63b`)
conflicted on this block and was resolved in favour of the version here, which
is a superset — it carries the same validation plus the `isLcdTypeSupported()`
gating from findings 2–3, which the bot's commit does not have. The two are
otherwise equivalent in effect.

---

PR #1258, opened 2026-09-13
ArduinoJson v7: restore presence-check semantics, merge master, build fixes

RAR:
Stacked on #1044 (targets `copilot/sub-pr-978`); addresses the review at https://github.com/OpenEVSE/openevse_esp32_firmware/pull/1044#issuecomment-5654320356.

- **Merge current master** (three conflicts: `platformio.ini` OpenEVSE 0.0.24 + ConfigJson 0.0.7, `event_log.cpp` rfidTag, `app_config.cpp` lcd_type/cable_temp blocks).
- **Restore presence-check semantics.** 103 `is<T>()` gates that replaced a bare `containsKey()` are now `!doc["k"].isNull()`, so a numeric string, a float, or `1` for a bool still applies as it did under v6 (the only difference is an explicit JSON `null`, now ignored instead of converted to 0/false/""). 19 gates stay as `is<T>()`: the `energy_meter` sites that already type-checked under v6, the scheduler ids, and the structural `is<JsonObject>`/`is<JsonArray>` checks.
- **`/config` static document dropped.** `JsonDocument::clear()` frees every pool in 7.4.3, so the static retained nothing; the comment now says so and asks for the v6-era TFT heap measurement to be repeated.
- **Deprecation noise** 355 → 66 via `-D ARDUINOJSON_DEPRECATED\\(msg\\)=` (the remaining 66 are `JSON_*_SIZE` pragmas the macro cannot reach). Note the escaping: PlatformIO passes flags through `shlex` and then `sh -c`, so single quotes do not survive.
- **`patch_microocpp.py` fails the build** naming the file and first unmatched pattern when a replacement no longer applies, instead of proceeding to an unrelated compile error.
- Stray `_codeql_detected_source_root` symlink removed; unused `JsonSerialize<CAPACITY>` template parameter dropped.

ConfigJson 0.0.7 has the same `is<T>()` gate for every config option; fix proposed separately at jeremypoulter/ConfigJson (see linked PR). Until that ships, `POST /config` with a numeric string for a `uint32_t` option is still ignored under v7.

Verified: `openevse_wifi_v1` 1,802,025 B flash (91.7%, 8.5 KB under the PR head), `native_openevse`, `native_simulator`, 81/81 simulator tests, all 13 doctest suites.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring and detecting supported LCD types.
  * Added optional cable temperature monitoring and configuration, including sensor settings, status, limits, and controls.

* **Bug Fixes**
  * Improved configuration handling so present fields are processed reliably regardless of their stored JSON type.
  * Improved configuration saving and loading, including safer JSON parsing and preservation of notification acknowledgements across firmware versions.

* **Maintenance**
  * Updated serialization and library compatibility handling for improved reliability.


jeremypoulter:
Sorry already had resolved several of these issues locally, can you pull the changes and see what is left?

RAR:
Pulled `copilot/sub-pr-978` (be39151b + the master merge) in — 9ef164c1. Three real conflicts: `app_config.cpp` (comment only), `event_log.cpp` (took your v7 comments), `scheduler.cpp` (took your `obj["time"].is<const char*>()`).

Where your be39151b and this branch had both touched the same `is<T>()` sites, I dropped ours and kept yours (c872576a) — you'd left `is<const char*>()` where the value goes straight to `as<const char*>()`/`strcmp()`, which is the right call since a non-string would otherwise become a null pointer.

What's left in the PR now is just the parts you hadn't covered: `JsonSerialize` CAPACITY template removal, silencing `ARDUINOJSON_DEPRECATED` for MicroOcpp, `patch_microocpp.py` failing loudly on an unmatched replacement, `handleConfigGet` dropping the static document, and the stray `_codeql_detected_source_root`. Nine files.

native_test passes, `openevse_wifi_tft_v1` builds clean.

jeremypoulter (approved):

Issue #112, opened 2020-08-26
Add UI to enable/disable front button.


chris1howell:
There is currently no way to enable/disable the front button short of unplugging. Stations installed outdoors or in publicly accessible locations are venerable to tampering of settings or interrupting a charge.

RAPI command has been incorporated into openevse to enable/disable button 
$FF B 0
$FF B 1

https://github.com/lincomatic/open_evse/commit/c226ad80166ba79d47b27a96e911276b0a0cacb6

jeremypoulter:
With the v4 changes the WiFI module now always disables the button and handles the button presses it the WiFi module, so is there still a requirement to implement the disable button logic?

jeremypoulter:
@chris1howell Any thoughts?

chris1howell:
Some users want the button disabled regardless of where it is interpenetrated. If someone walks up to an outdoor station they can mess with your station. There should be a way to lockout public access to the button.

PR #1271, opened 2026-09-15
Switch mDNS advertising, resolution and discovery to ArduinoMongoose

jeremypoulter:
## Summary

Follow-up to #1155, **targeting its `copilot/update-mongoose-7-and-https-server-ui` branch**.

Switch hostname advertising/resolution and DNS-SD peer discovery to ArduinoMongoose's shared mDNS listener. This follows @RAR's diagnosis in https://github.com/OpenEVSE/openevse_esp32_firmware/pull/1155#issuecomment-5672663767 and removes the need to run the IDF and Mongoose resolvers together.

## Upstream dependency

Depends on **https://github.com/jeremypoulter/ArduinoMongoose/pull/112**, pinned to its current tip `49ec64dc710cf2a28c4dba31da3cadcfff08aac3`.

The required query-type, DNS-SD parsing, TXT encoding, listener-lifecycle, multicast and resolver-cache fixes are implemented and tested in ArduinoMongoose. The firmware uses the library APIs directly.

The pin has moved twice since this PR was opened, both re-pins of the same upstream PR #112 branch as it was reconciled forward:

- `3b377c1` → `0843769`: merged current ArduinoMongoose master (which also carries an unrelated Mongoose 7.22→7.23 bump) into PR #112's branch, and implemented its DNS-SD browse layer against the reconciled response shape (it had been left a stub mid-merge).
- `0843769` → `49ec64d`: **fixed a real regression** the `0843769` re-pin introduced. Three of this PR's own Integration Tests failed at `0843769` (`test_peer_discovery_mdns[4]`, `test_reciprocal_add_and_remove_do_not_loop`, `test_member_leaves_group_with_rekeyed_controller_host`) — mDNS peer discovery returned zero peers. Root cause: the browse rewrite (forced by Mongoose 7.23's new one-event-per-datagram model) dropped the periodic PTR re-query the previous implementation sent for the life of a browse; `Mdns.browse()` sent exactly one query and never resent it, so any peer that missed that single multicast — not yet listening, a dropped datagram, an RFC 6762 §6 throttle window — was permanently invisible for that discovery cycle, which is exactly what four native instances starting together would hit. `49ec64d` restores resending, bounded in both rate (`MG_MDNS_BROWSE_RETRY_MS`, default 1000ms) and count (`MG_MDNS_BROWSE_MAX_RETRIES`, default 4) so it cannot reproduce the *unbounded* per-poll re-query that caused the original RFC 6762 §6 problem this whole mDNS effort started from. Full account in that commit's message. **Confirmed fixed**: this PR's own Integration Tests job passed at `49ec64d` (78 tests, 77 passed, 1 skipped, 0 failed) — including all three tests that failed at `0843769`.

`DiscoveredService{instance, hostname, port, addresses, txt}` is unchanged across all three pins, so no firmware source edits were needed for any of them.

## Changes

- Start/restart `Mdns` after network bring-up/reconnect and advertised configuration changes; clear it when the network goes down.
- Preserve `_http._tcp` / `_openevse._tcp`, active HTTP/HTTPS port, and `type`, `version`, `id`, `ssl` TXT fields.
- Replace IDF asynchronous-query handles and blocking result polling with `Mdns.browse()`, a timed `services()` snapshot and `cancelBrowse()`.
- Use the SRV target as peer hostname, keep service-instance names separate, retain self filtering and same-subnet IPv4 preference.
- Disable ArduinoOTA's implicit IDF responder; advertise its `_arduino._tcp:3232` service and IDE TXT fields through Mongoose on OTA-enabled hardware.
- Remove native EpoxymDNS/Avahi linkage and corresponding CI/container prerequisites. The harness no longer requires host mDNS tools.
- Require all other test devices in the four-instance discovery test, excluding the local device and unrelated LAN devices.
- Capture native firmware output in per-start files: unread stdout/stderr pipes could fill and block the event loop during integration tests.

## Validation

- GUI build and tests: **144 files / 1,155 tests passed**.
- Divert simulation: **81 passed**.
- Firmware native unit tests: `native_test`'s doctest suites build and pass (PlatformIO's summary labels these skipped despite successful doctest execution; verified with verbose output, e.g. `test_http_etag`: 10 cases / 42 assertions, all passing).
- `openevse_wifi_v1`: **build passed**, using the repository's default `platform`.
- `openevse_wifi_v1_16mb`: **core-3/IDF5 build passed**.
- `native_openevse`: **build passed** against the current pin.
- Upstream ArduinoMongoose at `49ec64d`: `tests/unit` — `native` 69 cases / `native_test` 71 / `native_mdns_asan` 22, all passing; the browse retry fix is covered by two new tests and proven at production timing (no test speedup) under `native_test`. CodeRabbit review of the fix commit: no actionable findings.
- **This PR's own Integration Tests job, at the current pin (`49ec64d`)**: **78 tests, 77 passed, 1 skipped, 0 failed** — including the three discovery tests that failed at the previous pin. All ~19 ESP32 board builds, Unit Tests, Simulator Tests, divert simulator and Spectral also green at this commit.

ESP32 builds were compile-checked. On-device WiFi/Ethernet/AP and Arduino IDE OTA discovery still need hardware verification.

https://claude.ai/code/session_01HYWhw7dg4Arrpa5TCbGNwt


RAR (commented):
Read the diff, the full post-change `net_manager.cpp` / `loadsharing_discovery_task.cpp`, and ArduinoMongoose at the pinned `49ec64d` to check the contract the firmware now relies on.

The design is right — one listener for advertise/resolve/browse, IDF responder and Avahi both gone, `DiscoveredService` shape stable across all three pins. Four things I'd want addressed before merge.

## 1. mDNS is started before Mongoose is initialised on a first boot

`main.cpp:221` calls `net.begin()`, and `MicroTask.startTask()` runs `setup()` synchronously, so `NetManagerTask::setup()` → `manageState()` → `wifiStartInternal()` → `wifiStartAccessPoint()` → `WiFi.enableAP(true)` all complete before `Mongoose.begin()` at `main.cpp:226`.

With no SSID configured (first boot, factory reset), `isWifiModeAp()` is already true when `setup()` reaches `updateMdns()` at `net_manager.cpp:628`, so `Mdns.begin()` (`:647`) calls `mg_mdns_listen(Mongoose.getMgr(), …)` on a manager that has not been `mg_mgr_init()`ed. `mg_mgr_init` starts with `memset(mgr, 0, sizeof(*mgr))`, so five lines later the listener connection and its socket are orphaned — leaked, never polled, and `mgr->mdns` is cleared. `_mdnsConfig` has already been set to the AP signature, so `updateMdns()` never retries: `openevse-xxxx.local` and the service records are dead for the whole captive-portal session, and only come back when the signature changes (user configures WiFi and `haveNetworkConnection()` blanks `_mdnsConfig`).

The old code was immune because `MDNS.begin()` didn't touch Mongoose. Simplest fix: drop the `updateMdns()` call from `setup()` — the loop calls it at `:851` anyway, and the first loop pass runs after `main`'s `setup()` has finished.

## 2. The browse window outlives the library's record cache

`_poll_interval_ms` is 2000 and `_query_timeout_ms` is 5000, and `loop()` always returns the poll interval, so the window check at `loadsharing_discovery_task.cpp:196` first fires at **t0+6000** — and that is the only moment `Mdns.services()` is read. In the library each record carries `expires = last_response + MG_MDNS_CACHE_TTL_MS`, default **5000** (`MongooseMdns.cpp:350`, `mongoose.h:1320`), and `services()` drops anything with `expires <= now`.

So a peer that answers the initial PTR query and nothing after is expired before we ever look. It works today only because the browse resends at +1s…+4s and the last retry refreshes the record to t0+9000 — i.e. discovery correctness now rests on `MG_MDNS_BROWSE_MAX_RETRIES` staying ≥ 4 *and* every peer answering that final retry. A peer applying RFC 6762 §7.4 duplicate suppression, or any future retuning of those constants, silently loses peers with no failure signal.

Cheapest robust fix: read `services()` on every poll tick and merge by instance rather than snapshotting once at the end. That decouples the task from the library's cache TTL entirely.

## 3. No recovery if the listener fails to start or closes

`updateMdns()` returns void and runs only when `NetManagerTask` wakes; in the steady `Connected` state `manageState()` returns `MicroTask.Infinate`. If `Mdns.begin()` or any `addService` fails (`:661`, `:668` both call `end()` and leave `_mdnsConfig` empty), or the listener is later closed by Mongoose (`MG_EV_CLOSE` nulls `_mdns`), nothing schedules a retry until an unrelated network event happens to wake the task.

The same path swallows a bad hostname: `Mdns.begin()` rejects any name containing a `.` or longer than 63 chars, and the user-settable `hostname` config isn't validated, so `evse.home` turns off advertising, resolution *and* peer discovery with only a `DBUGLN`. Suggest `updateMdns()` returning a delay (short retry while `!Mdns.isActive()`) folded into `nextLoopDelay`.

## 4. `updateMdns()` silently cancels an in-flight browse

`Mdns.begin()` calls `end()`, which does `cancelBrowse()` and drops all browse records. Any advertised-config change (HTTPS toggle, port, hostname) or reconnect landing inside a discovery window leaves `LoadSharingDiscoveryTask` believing a browse is running; at t0+6000 it reads an empty snapshot and assigns it to `_cachedPeers` with `_cacheValid = true`, so the peer list goes empty in the UI until the next cycle. Worth either not replacing a non-empty cache with an empty result, or having `updateMdns()` poke the discovery task.

## Smaller points

- **`loadsharing_discovery_task.cpp:246`** — the same-subnet preference loop can only ever see one entry: `handleResponse()` does `rec->addresses = {resp.addr}`, and only when `resp.addr.addr.ip4` is non-zero. The retained comment about IPv6 AAAA records arriving alongside the A record no longer describes anything the library can produce, so the preference logic is effectively dead. Fine to keep as future-proofing, but the comment should say so.
- **Query volume.** With load sharing on, `_discovery_interval_ms` is 10 s and each browse now emits 5 PTR queries (1 + 4 retries) instead of 1 — 5× the multicast, each answered by every OpenEVSE on the LAN. The retries are also a fixed 1 s apart, where RFC 6762 §5.2 requires successive intervals to at least double. Given this whole effort started from a §6 rate-limit problem, worth a deliberate decision: back the retries off (1s/2s/4s) or lengthen the discovery interval.
- **Stale leftovers the sweep missed:**
  - `.github/workflows/build.yaml:163` — step still named "Install Avahi development libraries" (the one at `:52` was renamed).
  - `docs/ai/cloud-environment.md:114` — still describes the session-start hook as installing "the Avahi/OpenSSL dev headers".
  - `tests/integration/README.md:51` — still claims "Avahi is already installed as part of native firmware build prereqs".
  - `tests/integration/README.md:241` and `tests/integration/conftest.py:217` — still say `pio run -e native` where the prerequisites section was corrected to `native_openevse`.

## What checks out

The `_mdnsConfig` signature covers the fields that matter and `haveNetworkConnection()` correctly blanks it to force a multicast rejoin on same-IP reconnects; `_arduino._tcp` TXT matches ArduinoOTA's contract (`auth_upload=no` is right — no `setPassword()` anywhere in the tree); the OpenEVSE TXT set is nowhere near the library's 256-byte budget and `MAX_SERVICES` (8) has room for the three registered services; `lib_ignore = ${common.lib_ignore}, EpoxymDNS` doesn't collide with another key in that env and no source still includes `ESPmDNS.h` or calls `mdns_query_*`; and the integration-test tightening (`num_instances - 1`, matching on collected instance IDs, per-start log files instead of unread pipes) is a real improvement over the old tolerance-based assertion.

Issue #1275, opened 2026-09-17
Uplift LVGL 8.4 → 9.x for the TFT local UI


jeremypoulter:
LVGL 9 is not a drop-in replacement for our TFT local UI. Dependabot keeps proposing the bump as a one-line `platformio.ini` change (#1134 for 9.5.0, #1274 for 9.6.0), and both break the only two environments that link LVGL. This issue records why we should still do the uplift eventually, what it costs, and the findings needed to implement it, so the next attempt doesn't start from scratch.

Only `openevse_wifi_tft_v1` and `openevse_wifi_tft_v1_dev` (plus the `native_openevse_lvgl` host env) link LVGL, so nothing else in the repo is affected either way.

## Why do it

**The real reason is upstream maintenance, not features.**

| branch | last release |
|---|---|
| v8.4.0 | 2024-03-19 — end of the line. No 8.4.1, no 8.5. |
| v9.x | 9.6.0 on 2026-09-16; 9.5, 9.4, and 9.3.0 (2025-06) before it |

We are pinned to a branch that stopped receiving fixes two and a half years ago. Every rendering, memory and correctness fix since March 2024 is v9-only. For firmware that runs unattended in a garage, that is the argument.

Secondary benefits:

- Ends the recurring dependabot churn — every lvgl release re-proposes this PR.
- Ecosystem alignment: `espressif/esp_lvgl_port`, community ESP32 drivers and the SquareLine/EEZ designers all target v9. Relevant only if we ever want to lay out a screen in a tool rather than hand-writing C++.
- Forward headroom we would cash in later, not now:
  - `lv_observer` (subject/observer binding) is genuinely v9-only — confirmed absent from the v8.4.0 tree. The idiomatic v9 answer to the manual change-detection plumbing in `lcd_lvgl.cpp` / `charge_screen.cpp`.
  - The draw-unit architecture makes 2D acceleration pluggable. Worth nothing on an ESP32-WROOM with no accelerator; worth a lot if a display board ever moves to an S3/P4.
  - `LV_COLOR_FORMAT_RGB565_SWAPPED` is a first-class render format with dedicated blend paths, so the byte swap this ILI9488 needs becomes part of rendering instead of a config flag. Plausible small perf win — **unmeasured**.
  - LZ4/RLE image compression (v9-only) would shrink the ~19 KB of A8 mark assets. Trivial next to a 16 MB partition.

## Why not do it yet

[...]

PR #1134, opened 2026-06-21
Bump lvgl from 8.4.0 to 9.5.0

jeremypoulter:
Bump lvgl from 8.4.0 to 9.5.0

PR #1279, opened 2026-09-18
build: pin LVGL to the v8.4.0 release archive to stop the v9 dependabot churn

RAR:
The "stopping the churn" item from #1275.

`peterus/platformio_dependabot` has no ignore rule: it takes every registry package where `pio pkg outdated` reports a newer `latest`, rewrites the spec in `platformio.ini`, and opens a PR (the branch-exists check is the only dedupe, so each lvgl release gets a fresh one — #1134, #1274). The only thing it does not follow is a non-registry spec. So this pins LVGL by its release archive instead:

```
lvgl_lib = https://github.com/lvgl/lvgl/archive/refs/tags/v8.4.0.zip
```

used from the two `lib_deps` lists that carried `lvgl/lvgl@^8.3.9`. Same code — it resolves as `lvgl @ 8.4.0`, which `^8.3.9` already did. The comment beside it points at #1275 and says to bump the tag when the v9 uplift lands, so the pin does not outlive its reason.

Verified: `openevse_wifi_tft_v1` builds (2,592,471 B, 39.6%), dependency resolves as `lvgl @ 8.4.0`.

## Summary by CodeRabbit

* **Chores**
  * Updated the LVGL dependency to version 8.4.0.
  * Standardized LVGL dependency usage across supported TFT and native environments.


jeremypoulter (approved):
Loovs good to me

PR #1281, opened 2026-09-20
evse_man: register the evcc claim client id (262145)

RAR:
evcc's new OpenEVSE driver (evcc-io/evcc#33539, via [OpenEVSE/go-openevse](https://github.com/OpenEVSE/go-openevse)) controls the charger through `/claims` and claims as client **262145** = vendor `0x0004`, client `0x0001`. The go-openevse README describes that id as "registered as `EvseClient_evcc` in the firmware", but `evse_man.h` — which `api.yml` points at as the client-id registry — only lists vendors `0x0001`–`0x0003`.

This reserves vendor `0x0004` for evcc and names the client, so nothing else picks the same id. No behaviour change; the firmware already accepts any client id on `/claims`.

GUI half (names the claim in the rate pill / Manager tab instead of "null"): OpenEVSE/openevse-gui-nightshift#138.

## Summary by CodeRabbit

* **New Features**
  * Added support for identifying the evcc vendor and client in EVSE communications.
  * Introduced a unique evcc client identifier for interoperability with compatible systems.


jeremypoulter (approved):

jeremypoulter:
Looks good

PR #1284, opened 2026-09-21
feat(platformio): add LilyGO T-ETH-Lite (RTL8201) board environment

jdgarcia99:
## Summary
- Adds `env:lilygo-t-eth-lite-rtl8201` for the LilyGO T-ETH-Lite (ESP32-WROVER-E, RTL8201 PHY, 16MB flash, 8MB PSRAM), reusing the WT32 WS2812 feature set and adapting pins to this board's wiring
- PN532 RFID reader and MCP9808 temperature sensor I2C config is left commented out (not populated on this board by default), following the same pattern already used in `env:wt32-eth01-ws2812`

Related: OpenEVSE/openevse-gui-nightshift#139 adds a wizard UX improvement (skip the WiFi hand-off warning when already reachable over Ethernet) that benefits this board and other wired-ethernet boards already in the tree (wt32-eth01, olimex-gateway). Not a dependency of this PR.

## Test plan
- [ ] `pio run -e lilygo-t-eth-lite-rtl8201` builds successfully
- [ ] Flashed and verified Ethernet (RTL8201) link-up on the LilyGO T-ETH-Lite
- [ ] Verified WS2812 status LED and WiFi/RAPI button behavior on external wiring

## Summary by CodeRabbit

- **New Features**
  - Added support for the LilyGO T-ETH-Lite ESP32-WROVER-E board.
  - Enabled Ethernet connectivity, PSRAM, 16 MB flash, and custom partitioning.
  - Added board-specific NeoPixel lighting and button input support.
  - Added optional configuration for PN532 and MCP9808 peripherals.
  - Updated serial communication to use the board’s default UART0 pins.
  - Added support for the board’s integrated RTL8201 Ethernet controller.


jdgarcia99:
Thanks for flagging both — addressed in 63389654, though not exactly as suggested:

**RAPI wiring (line 507-509):** the diagnosis was right (the comment didn't match the code), but the fix isn't rewiring to RX=35/TX=33 or moving `NEO_PIXEL_PIN` — this board's RAPI link is physically wired to the chip's UART0 default pins (RX=44/TX=43), confirmed against real hardware. `RX2`/`TX2` were genuinely dead (only affect Serial2, unused here) and are now removed; the comment now describes the actual wiring instead.

**MCP9808 `SDA`/`SCL` (line 516-518):** the suggested rename to `I2C_SDA`/`I2C_SCL` would have broken the build. This project always calls `Wire.begin(I2C_SDA, I2C_SCL)` explicitly (`EvseMonitor::setup()`), so those are already correct and unrelated to the bare `SDA`/`SCL` symbols — those are consumed at compile time by arduino-esp32's own `Wire.cpp` (its no-args-`begin()` fallback, never exercised here since we always pass explicit pins). Confirmed by reproducing the build failure locally. Simplest fix: drop the `-D SDA=`/`-D SCL=` override entirely and let `board=esp32dev`'s own `pins_arduino.h` (which already declares them) satisfy the compiler — no need to define them ourselves, and doing so would actually collide with that declaration.

jdgarcia99:
Correction to my own comment above: the RAPI pins are TXD=GPIO1/RXD=GPIO3 (matching UART0's actual default on classic ESP32), not RX=44/TX=43 as I first wrote — I had the wiring wrong, since corrected in 0a65a4e4. The rest of the analysis stands.

PR #1285, opened 2026-09-21
fix(rfid): return an error when /rfid/add is called with RFID disabled

jdgarcia99:
## Summary
- `handleAddRFID` (`/rfid/add`) always responded `200 {"msg":"Waiting for badge"}`, even when `RfidTask::waitForTag()` silently no-ops because the `rfid_enabled` config bit is off (it's only set from the "RFID" feature toggle in Charge Manager, not from the RFID settings page itself)
- The web UI's Scan button therefore looked like it did nothing when RFID wasn't enabled yet, with no error surfaced anywhere
- Now the endpoint returns `400` with a clear message when RFID isn't enabled, so the frontend can show an actionable error — see OpenEVSE/openevse-gui-nightshift#140, which is the PR that actually surfaces this as a visible alert

## Test plan
- [x] `pio run -e lilygo-t-eth-lite-rtl8201` builds successfully
- [x] Reproduced on real hardware (LilyGO T-ETH-Lite + PN532): Scan silently did nothing before this fix
- [ ] Additional verification on other PN532-equipped boards welcome

## Summary by CodeRabbit

- **Bug Fixes**
  - Requests to add an RFID badge now return a clear error when RFID functionality is disabled, instead of waiting indefinitely for a badge.


jdgarcia99:
The `Build openevse_wifi_v1_16mb` job failure here is unrelated to this change — it's a transient `504` downloading `esp32-core-3.3.11-libs.tar.xz` from GitHub Releases during toolchain install (retried 5x, still failed). The same env builds successfully with the same base in #1284. Re-running the job should resolve it.

RAR (approved):
Mirrors the `config_rfid_enabled()` gate `RfidTask::waitForTag()` already applies, so the 400 only fires where the request was going to be silently dropped anyway. Pairs with OpenEVSE/openevse-gui-nightshift#140 for the visible error.

PR #1287, opened 2026-09-21
fix(build): work around core-3 tool-scons version-mismatch bug in scripts/pio

jeremypoulter:
## Summary
- `scripts/pio`'s `CORE3_ENVS` list was missing `openevse_wifi_tft_v1` and `openevse_wifi_tft_v1_dev` (only `openevse_wifi_v1_16mb` was listed), so those hybrid `arduino,espidf` envs were silently building against the wrong `PLATFORMIO_CORE_DIR`.
- Even after fixing that, `openevse_wifi_tft_v1_dev` still failed deterministically at the final link step with `ModuleNotFoundError: No module named 'SCons.Tool.FortranCommon'`. Root-caused it to a real bug in pioarduino/platform-espressif32 (55.03.311): `platform.json` pins `tool-scons` to its own `package-version`, but PlatformIO Core independently hardcodes a *different* `tool-scons` requirement to run its own build engine — both packages land in the same `packages/tool-scons` slot. When the platform notices a version mismatch, its own handler (`_handle_existing_tool` in `platform.py`) deletes `packages/tool-scons` to force a reinstall, but the reinstall check requires a `.piopm` marker that no longer exists post-delete, so it silently no-ops — leaving `tool-scons` gone for the rest of the build. `SCons.Tool.FortranCommon` is only imported lazily at the very last link step, so everything compiles fine for minutes before failing right at the end.
- Confirmed the exact deletion with kernel-level file tracing (`inotifywait`), and verified the fix end-to-end: `openevse_wifi_tft_v1_dev` now builds successfully (previously failed 100% of the time).

## Fix
`scripts/pio` now, for core-3 builds, primes `tool-scons` and syncs `platform.json`'s expected version to whatever actually landed on disk before every run — so the platform's exact-string version check never sees a mismatch in the first place. This is self-healing (reads the actual installed version each time rather than hardcoding one), so it survives future PlatformIO Core / platform releases changing their pinned versions.

## Test plan
- [x] `scripts/pio run -e openevse_wifi_tft_v1_dev` — `SUCCESS` (clean core-3 dir and warm cache both verified)
- [x] Re-ran a second time to confirm idempotency (no re-download thrash, still `SUCCESS`)
- [x] `scripts/pio run -e openevse_wifi_v1_16mb` unaffected (other core-3 env, still builds/fails on its own pre-existing, unrelated issue independent of this change)

## Summary by CodeRabbit

- **New Features**
  - Added support for the `openevse_wifi_tft_v1` and `openevse_wifi_tft_v1_dev` Core 3 environments.

- **Bug Fixes**
  - Improved Core 3 setup to install required build tooling automatically.
  - Improved version synchronization during setup when package metadata is available.
  - Preserved existing directory behavior for Core 2 and commands without a specified environment.


RAR (commented):
Root cause checks out against pioarduino's `platform.py`: `_handle_existing_tool()` removes the tool dir on a version mismatch and calls `install_tool()`, whose status check then finds neither `.piopm` nor `tools.json` and falls through to "already configured" — so tool-scons is simply gone. The mismatch itself is PlatformIO Core ≥ 6.2.0 (2026-09-05) moving `tool-scons` to `~4.41101.0` while 55.03.311 pins `4.40801.0`; 6.1.19 pins `~4.40801.0`, which is why CI already pins that version (`build.yaml:161`, 6797b994). The CI comment attributes it to SCons 4.11.1 being broken; this PR shows the deletion is the real cause, so the two are worth cross-referencing, and CI could drop its pin using the same sync. A line in the script comment saying it only bites on Core ≥ 6.2.0 — and never on pioarduino's own Core fork, which pins the same 4.8.1 tarball — would save the next person who can't reproduce it.

Ran `scripts/pio run -e openevse_wifi_tft_v1` from this branch on a host with the pioarduino Core fork and a warm `~/.platformio-core3`: SUCCESS, `pkg install` a no-op, sync left 4.40801.0 alone. The `CORE3_ENVS` omission was real — the TFT envs have been building in `~/.platformio` all along, which is also why the two platforms keep evicting each other's `framework-arduinoespressif32` there (`platform.py:616` removes every `framework-arduinoespressif32@*` copy).

One gap, and CodeRabbit has it right: on a cold `PIO_CORE3_DIR` the sync is guarded on `platforms/espressif32/platform.json`, which doesn't exist until `pio run` installs the platform — in the same process that then deletes tool-scons. So the first cold run fails and the second succeeds. When `platform.json` is absent, a `"$PIO_BIN" pkg install -e <env>` (or `platform install <url>`) before the sync closes it.

Minor: both workaround steps are `>/dev/null 2>&1 || true`, so if either ever fails the user gets the original FortranCommon error with no hint — one warning line on failure would help. And `CORE3_ENVS` could come from `pio project config --json-output` rather than a hand-kept list, since it has drifted once already.

PR #1290, opened 2026-09-21
time: default fresh installs to time.cloudflare.com instead of pool.ntp.org

### LINKED ISSUE #273: Default NTP server name
We shouldn't really be using  `pool.ntp.org` as a default NTP server (see <https://www.ntppool.org/en/vendors.html>), we should [apply](https://manage.ntppool.org/manage/vendor) for something like `openevse.pool.ntp.org` and set that as the default.

RAR:
Closes #273.

## Why

The issue's ask: we shouldn't ship bare `pool.ntp.org` as a device default — the pool's [vendor policy](https://www.ntppool.org/en/vendors.html) asks for a registered vendor zone. Rather than apply for and maintain `openevse.pool.ntp.org`, switch the default to **`time.cloudflare.com`**: global anycast, explicitly free for any client (embedded included), no registration, and it also offers NTS should we ever want authenticated time. At our load (one packet per charger every 8 hours) either provider is fine; this just removes the policy problem.

Considered and passed on: `time.nist.gov` (US-only servers, US-government single provider) and `time.google.com` (leap-second smearing, non-standard).

## What changes

- `SNTP_DEFAULT_HOST` in `src/emonesp.h` → `time.cloudflare.com`, with a comment recording the reason.
- `models/Config.yaml` example and a new `divert_sim/test_config.py` assertion on the default.
- **Existing installs are unaffected** — they keep whatever `sntp_hostname` they already have stored. Only fresh installs (and factory resets) pick this up. Networks handing out an NTP server via DHCP take precedence regardless (#1289).
- GUI placeholders follow in OpenEVSE/openevse-gui-nightshift#144.

`native_simulator` builds; `test_config.py` passes; `docs_coverage.py --strict` passes.

This is a project-wide default, so a nod from a maintainer on the provider choice is the main review point.

## Summary by CodeRabbit

- **Changes**
  - Fresh installations now use `time.cloudflare.com` as the default time synchronization server.
  - The example configuration has been updated to reflect the new default.
  - Existing installations retain their currently stored time synchronization server settings.
  - Configuration checks now verify the updated default hostname.


jeremypoulter:
The point of #273 was to register for the vendor pool, not a suggestion that we remove it. IIRC we chose NTP.org because it is independent so I am not sure this is the best move, so let's leave this open for a bit for comments.

chris1howell (approved):

